PR #44, opened 2025-05-13
feat(trees): tree-merging

kamilkisiela:


lib/query-planner/src/planner/tree/query_tree.rs
```diff
@@ -44,7 +44,7 @@ impl QueryTree {
         let mut accumulator = trees.remove(0);
 
         for item in trees {
-            accumulator.root = accumulator.root.merge_nodes(item.root);
+            accumulator.root = accumulator.root.merge_nodes(&item.root);
```
kamilkisiela, line 47: @dotansimha what do you think?

lib/query-planner/src/planner/tree/query_tree_node.rs
```diff
@@ -20,6 +20,35 @@ pub struct QueryTreeNode {
...
+
+fn merge_query_tree_node_list(target_list: &mut Vec<QueryTreeNode>, source_list: &[QueryTreeNode]) {
+    if source_list.is_empty() {
+        return; // nothing to merge from the source
+    }
+
+    for source_node in source_list.iter() {
+        let matching_target_node = target_list
+            .iter_mut()
+            .find(|target_node| **target_node == *source_node);
```
kamilkisiela: @dotansimha I have no idea if it's good or bad...

lib/query-planner/src/planner/tree/query_tree_node.rs
```diff
@@ -34,11 +63,16 @@ impl QueryTreeNode {
...
 
-    pub fn merge_nodes(mut self, mut other: QueryTreeNode) -> Self {
-        self.requirements.append(&mut other.requirements);
-        self.children.append(&mut other.children);
+    pub fn merge_nodes(&mut self, other: &Self) -> Self {
+        merge_query_tree_node_list(&mut self.children, &other.children);
+        merge_query_tree_node_list(&mut self.requirements, &other.requirements);
 
-        self
+        QueryTreeNode {
```
kamilkisiela, line 70: @dotansimha I failed to avoid cloning

lib/query-planner/src/tests/object_entities.rs
```diff
@@ -65,6 +87,30 @@ fn testing() -> Result<(), Box<dyn Error>> {
               amount of Float/COST
     ");
 
+    let gqt = QueryTree::merge_trees(qtps);
+
+    insta::assert_snapshot!(gqt.pretty_print(&graph)?, @r"
```
kamilkisiela, line 92: @dotansimha Here's the `merge_trees` in practice

lib/query-planner/src/planner/tree/query_tree.rs
dotansimha: originally I tried to take ownership so we can destruct it, but if you managed to do that by ref, it's even better. Cloning is ok in this case, we can iterate and improve this later!

lib/query-planner/src/planner/tree/query_tree_node.rs
dotansimha: can be `*target_node == source_node`

dotansimha: fixed

lib/query-planner/src/planner/tree/query_tree_node.rs
dotansimha: It's ok in this case, we'll try to consume `self` and `other` by using the actual value later, it can be tricky because we'll need to see if it even makes sense, as we might not always consume the whole `other`. I failed to address this previously, so what you did is exactly what I planned to do for the 1st iteration :)

lib/query-planner/src/tests/object_entities.rs
dotansimha: NICE

PR #47, opened 2025-05-15
FetchGraph

kamilkisiela:


lib/query-planner/src/graph/edge.rs
```diff
@@ -16,6 +16,9 @@ pub struct EntityMove {
 /// Represent a simple file move
 pub struct FieldMove {
     pub name: String,
+    pub type_name: String,
+    pub is_leaf: bool,
+    pub is_list: bool,
     pub join_field: Option<JoinFieldDirective>,
```
dotansimha, line 22: for later: feel like we can drop this from the actual Node at some point, as most of the data it holds is already part of the actual Node.

PR #58, opened 2025-05-19
Include input and output in QueryPlan nodes

kamilkisiela:


lib/query-planner/src/tests/overrides.rs
```diff
@@ -122,9 +122,9 @@ fn two_fields_simple_overrides() -> Result<(), Box<dyn Error>> {
     // `id` is missing now.
```
dotansimha, line 122: remove TODO

lib/query-planner/src/tests/requires_provides.rs
```diff
@@ -71,9 +71,9 @@ fn simple_requires_provides() -> Result<(), Box<dyn Error>> {
     // TODO: [3] is missing "upc" in input
```
dotansimha, line 71: remove TODO

lib/query-planner/src/utils/pretty_display.rs
```diff
@@ -0,0 +1,9 @@
+use std::fmt::{Formatter as FmtFormatter, Result as FmtResult};
+
+pub fn get_indent(depth: usize) -> String {
+    "  ".repeat(depth)
+}
+
+pub trait PrettyDisplay {
```
dotansimha, line 7: later we can allow passing custom args with generics, if you want :)

PR #61, opened 2025-05-19
fix(fetch_graph): print `output` in plan and add missing `__typename`

dotansimha:


lib/query-planner/src/ast/type_aware_selection.rs
```diff
@@ -140,6 +144,45 @@ fn merge_selection_set(target: &mut SelectionSet, source: &SelectionSet, as_firs
     }
 }
 
+fn find_selection_set_by_path<'a>(
```
dotansimha, line 147: nit: maybe we can implement this in a different, without having the same `fn` twice (mut/non-mut).

lib/query-planner/src/planner/fetch/fetch_graph.rs
```diff
@@ -862,6 +862,10 @@ fn process_plain_field_edge(
         false,
     );
 
+    if !field_is_leaf && !parent_fetch_step.input.has_typename_at_path(&fetch_path) {
```
dotansimha: @kamilkisiela is that the actually logic here?
I mean, leaf makes sense, but the check for existing one works well but only for the root of course.

lib/query-planner/src/planner/tree/query_tree.rs
```diff
@@ -19,13 +19,10 @@ impl QueryTree {
         QueryTree { root }
     }
 
-    #[instrument(skip(graph))]
+    #[instrument(skip(graph), fields(
```
dotansimha, line 22: just clean this up, it will make logging more quite.

lib/query-planner/src/planner/plan_nodes.rs
```diff
@@ -130,11 +129,7 @@ impl PrettyDisplay for FetchNode {
             requires.pretty_fmt(f, depth + 2)?;
             writeln!(f, "{indent}  }} =>")?;
         }
-        writeln!(f, "{indent}  {{")?;
-        for line in self.operation.lines() {
-            writeln!(f, "{indent}    {line}")?;
-        }
-        writeln!(f, "{indent}  }}")?;
+        writeln!(f, "{indent}  {}", self.operation)?;
```
dotansimha, line 132: @kamilkisiela it's less pretty, but you get the point :)

PR #67, opened 2025-05-21
Executor - `serve-command` - QP serialization

ardatan:
Ref GW-363
- Add missing node types
- Implement serialization using `kind`
- Align node properties with the original Apollo format
- Remove unnecessary properties like `is_leaf`, and types like `FlattenNodePathType`

### DIFF
--- a/lib/query-planner/src/ast/operation.rs
+++ b/lib/query-planner/src/ast/operation.rs
@@ -12,6 +12,8 @@ use super::{selection_item::SelectionItem, selection_set::SelectionSet};
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct OperationDefinition {
     pub name: Option<String>,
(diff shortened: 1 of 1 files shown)


lib/query-planner/src/ast/operation.rs
```diff
@@ -12,12 +12,13 @@ use super::{selection_item::SelectionItem, selection_set::SelectionSet};
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct OperationDefinition {
     pub name: Option<String>,
-    pub operation_type: RootOperationType,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub operation_kind: Option<OperationKind>,
```
dotansimha, line 18: Why `Option` here? I think `OperationKind` will always be set, there's no case where it does not configured.

PR #71, opened 2025-05-22
added correct structure for `input` / `output` selections

dotansimha:
- [x] add operation definition with variable definitions (for `representations`)
- [x] add handling for inline fragments
- [x] handle building of representations + entities query
- [x] handle input
- [x] handle output
- [x] keep pretty print

Closes https://github.com/graphql-hive/gateway-rs/issues/70 

image

### DIFF
--- a/lib/query-planner/src/tests/object_entities.rs
+++ b/lib/query-planner/src/tests/object_entities.rs
@@ -122,22 +122,43 @@ fn testing() -> Result<(), Box<dyn Error>> {
     QueryPlan {
       Sequence {
         Fetch(service: "store") {
-          {products{__typename id}}
+          {
+            products {
+              __typename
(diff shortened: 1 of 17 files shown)


lib/query-planner/src/tests/object_entities.rs
```diff
@@ -308,32 +341,59 @@ fn parent_entity_call_complex() -> Result<(), Box<dyn Error>> {
     QueryPlan {
       Sequence {
         Fetch(service: "d") {
-          {productFromD{__typename id name}}
```
kamilkisiela, line 311: I think something got lost here @dotansimha

dotansimha: fixed :)

PR #86, opened 2025-05-26
Executor - Use consumer schema for projection

ardatan:
Ref GW-371

### DIFF
--- a/bin/gateway/src/main.rs
+++ b/bin/gateway/src/main.rs
@@ -11,6 +11,7 @@ use query_plan_executor::ExecutionRequest;
 use query_plan_executor::SchemaMetadata;
 use query_planner::consumer_schema::ConsumerSchema;
 use query_planner::planner::Planner;
+use query_planner::state::supergraph_state::SupergraphState;
 use query_planner::utils::parsing::parse_operation;
 use query_planner::utils::parsing::parse_schema;
 use serde_json::json;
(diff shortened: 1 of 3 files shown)


bin/gateway/src/main.rs
```diff
@@ -98,141 +99,165 @@ fn collect_variables(
...
+fn get_subgraph_endpoint_map(
+    supergraph_ast: &graphql_parser::schema::Document<'static, String>,
+) -> HashMap<String, String> {
+    let mut subgraph_endpoint_map = HashMap::new();
+    for definition in &supergraph_ast.definitions {
+        if let graphql_parser::schema::Definition::TypeDefinition(TypeDefinition::Enum(enum_type)) =
+            definition
+        {
+            let name = &enum_type.name;
+            if name == "join__Graph" {
```
dotansimha: I think you can use `JoinGraph.NAME`? same for all other directives/enum and other parts that are coming from the Federation spec?

bin/gateway/src/main.rs
```diff
@@ -98,141 +99,165 @@ fn collect_variables(
...
-                                            subgraph_name = enum_value.to_string();
-                                        }
-                                        _ => {
-                                            panic!("Expected enum value for name");
-                                        }
+fn get_subgraph_endpoint_map(
+    supergraph_ast: &graphql_parser::schema::Document<'static, String>,
+) -> HashMap<String, String> {
+    let mut subgraph_endpoint_map = HashMap::new();
+    for definition in &supergraph_ast.definitions {
```
dotansimha: No need to parse it really, you can use the struct to mange directives and then do `.is(...)` and `.parse(...)`

ardatan: What do you mean? Could you give an example?

dotansimha: [https://github.com/graphql-hive/gateway-rs/tree/main/lib/query-planner/src/federation_spec](https://github.com/graphql-hive/gateway-rs/tree/main/lib/query-planner/src/federation_spec)

dotansimha: No need to parse it again 😄 we have utils for the Federation directives

PR #108, opened 2025-05-30
Validation & Caching Plans

ardatan:
- Caching the query plans
- Document Validation and caching the result
- Variable coercion validation
- Reuse the AST types from query planner instead of graphql-parser's types

Used https://github.com/moka-rs/moka/tree/main?tab=readme-ov-file#example-asynchronous-cache which is lock-free cache implementation

In order to have a proper validation, I needed to add introspection/base schema to the consumer schema. So both projection and validation will be able to know the default/base types/fields etc.

### DIFF
--- a/lib/query-planner/benches/qp_benches.rs
+++ b/lib/query-planner/benches/qp_benches.rs
@@ -53,4 +53,3 @@ fn all_benchmarks(c: &mut Criterion) {
 
 criterion_group!(benches, all_benchmarks);
 criterion_main!(benches);
-


.gitignore
```diff
@@ -6,3 +6,4 @@ node_modules/
 logs
 federation-audit-results.txt
 *.junit.xml
+supergraph.graphql
```
kamilkisiela: why?

bin/gateway/src/main.rs
```diff
@@ -84,168 +75,12 @@ struct ServeData {
...
-        let introspection_query = introspection_query_from_ast(&self.document);
-        let introspection_schema_root_json = json!(introspection_query.__schema);
-        SchemaMetadata {
-            possible_types: final_possible_types,
-            enum_values,
-            type_fields,
-            introspection_schema_root_json,
-        }
-    }
+    plan_cache: moka::future::Cache<u64, Arc<query_planner::planner::plan_nodes::QueryPlan>>,
```
kamilkisiela, line 81: We plan the `QueryPlan` which is okay, but in order to execute a plan we convert `DocumentNode` to `string`. This should be cached as well, as it's relatively expensive and is done for every subgraph call.

.gitignore
ardatan: Federation audit CLI puts supergraph file on the root for each test case, then run the gateway. So if you run it locally, you can push it to the repo by accident.

PR #133, opened 2025-06-07
feat(qp): support for multiple `@requires` with conflicting arguments

dotansimha:
- [x] deal with aliasing when fields conflict
- [x] add more tests
- [x] check that passthrough with parent also works in case of a conflict
- [x] prefer to alias `requires`-created fields over regular fields
- [x] figure out https://github.com/graphql-hive/gateway-rs/pull/133/files#r2142599148 
- [x] clean up the code 
- [x] patch relevant `response_path`
- [x] patch relevant `input` selection
- [x] fix variables support 
- [x] pass more audit tests


lib/query-plan-executor/src/lib.rs
```diff
@@ -215,11 +215,44 @@ impl ExecutableFetchNode for FetchNode {
             Value::Array(filtered_representations),
         );
 
+        let mut operation_def = self.operation.operation_def.clone();
+        if operation_def.variable_definitions.is_none() {
+            operation_def.variable_definitions = Some(Vec::new());
+        }
+
+        // TODO: Fix this in the query planner
```
dotansimha: @ardatan can you please elaborate what's needed here? I'm not sure I understood what's missing in the plan?

lib/query-planner/src/ast/operation.rs
```diff
@@ -111,7 +111,12 @@ pub struct VariableDefinition {
 
 impl Display for VariableDefinition {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "${}:{}", self.name, self.variable_type)
+        match &self.default_value {
```
dotansimha, line 114: @ardatan same

lib/query-planner/src/ast/type_aware_selection.rs
```diff
@@ -203,14 +203,16 @@ fn find_selection_set_by_path<'a>(
...
+            Segment::Field(field_name, args_hash) => {
                 let next_selection_set_option =
                     current_selection_set
                         .items
                         .iter()
                         .find_map(|item| match item {
                             SelectionItem::Field(field) => {
-                                if field.alias.as_ref().unwrap_or(&field.name).eq(field_name) {
+                                if field.selection_identifier() == field_name
+                                    && field.arguments_hash() == *args_hash
```
dotansimha, line 214: @kamilkisiela now that we have the args_hash there, we can compare it in a safer way.

lib/query-planner/src/ast/type_aware_selection.rs
```diff
@@ -270,14 +272,16 @@ fn find_selection_set_by_path_mut(
...
-            Segment::Field(field_name) => {
+            Segment::Field(field_name, args_hash) => {
                 let next_selection_set_option =
                     current_selection_set
                         .items
                         .iter_mut()
                         .find_map(|item| match item {
                             SelectionItem::Field(field) => {
-                                if field.alias.as_ref().unwrap_or(&field.name).eq(field_name) {
+                                if field.selection_identifier() == field_name
```
dotansimha, line 282: @kamilkisiela now that we have the args_hash there, we can compare it in a safer way.

lib/query-planner/src/ast/arguments.rs
```diff
@@ -11,6 +15,12 @@ pub struct ArgumentsMap {
     arguments_map: BTreeMap<String, Value>,
 }
 
+impl Hash for ArgumentsMap {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.arguments_map.hash(state);
```
kamilkisiela, line 20: it's BTreeMap, right? So different order of arguments gives different hash? Should we make it stable?

PR #135, opened 2025-06-08
feat(gw): switch to `tower` + `axum` instead of `actix-web` 

dotansimha:
- [x] axum instead of actix
- [x] split handlers to services
- [x] update deps 
- [x] make it work 
- [x] integrate with new normalization
- [x] make it pass graphlq-over-http audit
- [x] added basic CORS
- [ ] consider to break entry point to smaller pieces within the handler
- [x] tests 
- [x] test bench
- [ ] update bench subgraphs deps


supergraph.graphql
```diff
@@ -0,0 +1,79 @@
+schema
```
ardatan, line 1: I added this file to .gitignore before. Please don't remove it from `.gitignore` 😅

dotansimha: oopsi 🙂

PR #142, opened 2025-06-09
Rewrite Unification

ardatan:
No more Input and Output Rewrites but only Fetch Rewrites so both KeyRenamer and ValueSetter can be used on both input and output sides


lib/query-plan-executor/src/lib.rs
```diff
@@ -368,49 +372,29 @@
...
         &self,
         possible_types: &HashMap<String, Vec<String>>,
-        data: Value,
+        data: &mut Value,
         path: &[String],
-    ) -> Value {
+    ) {
         if path.is_empty() {
-            return self.set_value_to.clone();
+            return;
```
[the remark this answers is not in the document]
dotansimha: @ardatan confirm it's intended?

lib/query-plan-executor/src/lib.rs
```diff
@@ -420,27 +404,20 @@ impl ApplyInputRewrite for ValueSetter {
                         _ => type_condition, // Default to type_condition if not found
                     };
                     if entity_satisfies_type_condition(possible_types, type_name, type_condition) {
-                        let data = Value::Object(map);
-                        return self.apply_path(possible_types, data, remaining_path);
+                        self.apply_path(possible_types, data, remaining_path)
                     }
+                } else if remaining_path.is_empty() {
+                    *data = self.set_value_to.clone();
```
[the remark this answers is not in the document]
dotansimha: this one is intended, right?

.gitignore
```diff
@@ -7,3 +7,5 @@ logs
 federation-audit-results.txt
 *.junit.xml
 .DS_Store
+# This is created by the federation audits temporarily
+supergraph.graphql
```
kamilkisiela: let's name it `audit_supergraph.graphql` or put it in the audit directory and ignore just that.

dotansimha: done in https://github.com/graphql-hive/gateway-rs/pull/144

.gitignore
```diff
@@ -2,10 +2,4 @@
 .digraph
 *.pending-snap
 profile.json.gz
-node_modules/
```
ardatan, line 5: These are extra because we already have a gitignore under `audits`

PR #146, opened 2025-06-10
chore(qp): refactor `SupergraphState`, get rid of references, cleanup

dotansimha:
- [x] remove references in `SupergraphState` that makes it impossible to work with
- [x] eliminate unused lifetime declarations
- [x] cleanup old and unused code

### DIFF
--- a/lib/query-planner/src/graph/mod.rs
+++ b/lib/query-planner/src/graph/mod.rs
@@ -19,8 +19,7 @@ use crate::{
     utils::ast::strip_modifiers_from_type_string,
 };
 use error::GraphError;
-use graphql_parser::query::{Selection, SelectionSet, Type};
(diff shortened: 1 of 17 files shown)


lib/query-planner/src/graph/mod.rs
```diff
@@ -766,7 +741,7 @@ impl Graph {
                             field.name.to_string(),
                             parent_type_def.name().to_string(),
                             state.is_scalar_type(parent_type_def.name()),
-                            field_in_parent.source.field_type.is_list_like_type(),
+                            field_in_parent.field_type.is_list(),
```
kamilkisiela, line 744: is it true for `[String]!` (non-nullable list)`?

dotansimha: Yeah, it was broken while i worked on that, tests were failing (due to missing `@` in plans), and I fixed it 🫡

PR #147, opened 2025-06-10
Perform normalization based on supergraph state + normalization of `@provides` and `@requires`

kamilkisiela:
AST normalization is now based on the SupergraphState as we need subgraph information.
We need that information to normalize `@requires(fields:)` and `@provides(fields:)` which are bound to subgraph.

### DIFF
--- a/lib/query-planner/src/ast/arguments.rs
+++ b/lib/query-planner/src/ast/arguments.rs
@@ -54,11 +54,7 @@ impl From<&Vec<(String, ParserValue<'_, String>)>> for ArgumentsMap {
 
 impl From<Vec<(String, ParserValue<'_, String>)>> for ArgumentsMap {
     fn from(value: Vec<(String, ParserValue<'_, String>)>) -> Self {
-        let arguments_map = value
-            .iter()
-            .map(|(key, value)| (key.clone(), Value::from(value)))
-            .collect();
-        Self { arguments_map }
+        (&value).into()
     }
 }
 


lib/query-planner/src/ast/arguments.rs
```diff
@@ -52,6 +52,16 @@ impl From<&Vec<(String, ParserValue<'_, String>)>> for ArgumentsMap {
     }
 }
 
+impl From<Vec<(String, ParserValue<'_, String>)>> for ArgumentsMap {
```
dotansimha, line 55: This one can just be: 

```rs
impl From<Vec<(String, ParserValue<'_, String>)>> for ArgumentsMap {
  fn from(value: Vec<(String, ParserValue<'_, String>)>) -> Self {
      (&value).into()
  }
}
```

lib/query-planner/src/federation_spec/mod.rs
```diff
@@ -18,35 +24,89 @@ pub(crate) mod join_implements;
...
+        &mut parsed_doc,
+        None,
+        Some(RootTypes {
+            query: Some(type_name),
+            mutation: None,
+            subscription: None,
+        }),
+        Some(subgraph_name),
+    )
+    .unwrap_or_else(|err| panic!("Normalization error: {err}"));
```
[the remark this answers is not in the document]
dotansimha: @kamilkisiela confirm?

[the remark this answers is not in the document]
kamilkisiela: I just followed what we had, we were panicing here before

lib/query-planner/src/ast/arguments.rs
```diff
@@ -52,6 +52,16 @@ impl From<&Vec<(String, ParserValue<'_, String>)>> for ArgumentsMap {
...
+    fn from(value: Vec<(String, ParserValue<'_, String>)>) -> Self {
+        let arguments_map = value
+            .iter()
+            .map(|(key, value)| (key.clone(), Value::from(value)))
+            .collect();
+        Self { arguments_map }
+    }
+}
```
kamilkisiela:
```suggestion
impl From<Vec<(String, ParserValue<'_, String>)>> for ArgumentsMap {
    fn from(value: Vec<(String, ParserValue<'_, String>)>) -> Self {
        (&value).into()
    }
}
```

lib/query-planner/src/ast/arguments.rs
kamilkisiela: done

PR #152, opened 2025-06-11
Executor Improvements

ardatan:
**Avoid Cloning**
Now we avoid cloning as much as we can. We mutate the objects.

**Benchmarks for execution components**
- With HTTP
- Without HTTP but with local schemas
- With projection
- Without projection
- Projection

**Posting overall benchmark results**
- CI adds a comment showing the overall gateway performance

**Unit tests**
- I started with a very simple unit test that tests the executor with the local execution

### DIFF
--- a/lib/query-plan-executor/src/executors/http.rs
+++ b/lib/query-plan-executor/src/executors/http.rs
@@ -6,12 +6,18 @@ use tracing::instrument;
 use crate::{executors::common::SubgraphExecutor, ExecutionRequest, ExecutionResult};
 
 #[derive(Debug)]
-pub struct HTTPSubgraphExecutor<'a> {
-    pub subgraph_endpoint_map: &'a HashMap<String, String>,
-    pub http_client: &'a reqwest::Client,
+pub struct HTTPSubgraphExecutor {
+    pub subgraph_endpoint_map: HashMap<String, String>,
+    http_client: reqwest::Client,
 }
 
-impl HTTPSubgraphExecutor<'_> {
+impl HTTPSubgraphExecutor {
+    pub fn new(subgraph_endpoint_map: HashMap<String, String>) -> Self {
+        HTTPSubgraphExecutor {
+            subgraph_endpoint_map,
+            http_client: reqwest::Client::new(),
+        }
+    }
     async fn _execute(
         &self,
         subgraph_name: &str,
@@ -36,7 +42,7 @@ impl HTTPSubgraphExecutor<'_> {
 }
 
 #[async_trait]
-impl SubgraphExecutor for HTTPSubgraphExecutor<'_> {
+impl SubgraphExecutor for HTTPSubgraphExecutor {
     #[instrument(skip(self, execution_request))]
     async fn execute(
         &self,
--- a/lib/query-plan-executor/src/lib.rs
+++ b/lib/query-plan-executor/src/lib.rs
@@ -11,12 +11,12 @@ use query_planner::{
 };
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
-use std::{collections::HashMap, sync::Arc, vec};
+use std::{collections::HashMap, vec};
 use tracing::{debug, instrument, warn}; // For reading file in main
 
 use crate::{
     deep_merge::deep_merge_objects, executors::common::SubgraphExecutor,
-    executors::http::HTTPSubgraphExecutor, schema_metadata::SchemaMetadata,
+    schema_metadata::SchemaMetadata,
 };
 mod deep_merge;
 pub mod executors;
@@ -743,7 +743,7 @@ pub struct QueryPlanExecutionContext<'a> {
     // Using `Value` provides flexibility
     pub variable_values: &'a Option<HashMap<String, Value>>,
     pub schema_metadata: &'a SchemaMetadata,
-    pub executor: SubgraphExecutorType<'a>,
+    pub executor: &'a SubgraphExecutorType<'a>,
     pub errors: Vec<GraphQLError>,
     pub extensions: HashMap<String, Value>,
 }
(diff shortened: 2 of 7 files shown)


lib/query-plan-executor/src/lib.rs
```diff
@@ -228,12 +195,11 @@
             .await;
 
         // 5. Process the response
```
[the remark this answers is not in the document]
ardatan: Looks good?

lib/query-plan-executor/src/deep_merge.rs
```diff
@@ -0,0 +1,61 @@
...
+    match (target, source) {
+        // 1. Source is Null: Do nothing
+        (_, Value::Null) => {} // Keep target as is
+
+        // 2. Both are Objects: Merge recursively
+        (Value::Object(target_map), Value::Object(source_map)) => {
+            deep_merge_objects(target_map, source_map);
+        }
+
+        // 3. Both are Arrays of same length(?): Merge elements
```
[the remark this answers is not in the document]
ardatan: Not that important

lib/query-plan-executor/src/lib.rs
```diff
@@ -940,167 +924,213 @@
...
+                    new_obj.insert(response_key, Value::String(type_name.to_string()));
+                    continue;
+                }
+                let field_map = field_map.unwrap();
+                let field_type = field_map.get(&field.name);
+                if field.name == "__schema" && type_name == "Query" {
+                    obj.insert(
+                        response_key.to_string(),
+                        schema_metadata.introspection_schema_root_json.clone(),
+                    );
```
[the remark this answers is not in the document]
ardatan: But it projects `obj` not `new_obj`

bench/subgraphs/lib.rs
```diff
@@ -0,0 +1,4 @@
+pub mod accounts;
```
dotansimha, line 1: any reason to export this as a lib?

lib/query-plan-executor/src/executors/http.rs
```diff
@@ -0,0 +1,55 @@
+use std::collections::HashMap;
+
+use async_trait::async_trait;
+use tracing::instrument;
+
+use crate::{executors::common::SubgraphExecutor, ExecutionRequest, ExecutionResult};
+
+#[derive(Debug)]
+pub struct HTTPSubgraphExecutor<'a> {
+    pub subgraph_endpoint_map: &'a HashMap<String, String>,
```
dotansimha: I wonder if we should have a `subgraph_endpoint_map: HashMap<String, reqwest::Client>` ? and then re-use the client? can the URL already be baked into?

lib/query-plan-executor/src/executors/local.rs
```diff
@@ -0,0 +1,91 @@
...
+
+use async_graphql::{dynamic::Schema, PathSegment, Response, ServerError};
+use async_trait::async_trait;
+
+use crate::{
+    executors::common::SubgraphExecutor, ExecutionRequest, ExecutionResult, GraphQLError,
+    GraphQLErrorLocation,
+};
+
+pub struct LocalSubgraphExecutor<'a> {
```
dotansimha, line 12: Should this just be named `MockedSubgraphExecutor`? and limited to test-only using `#[cfg_attr(test)]`?

lib/query-plan-executor/src/lib.rs
```diff
@@ -467,39 +484,40 @@ impl ExecutablePlanNode for ParallelNode {
...
-            errors.extend(result_errors);
         }
-        // Return the merged representations
-        (data, representations, errors)
+
+        // Wait for all jobs to complete
+        let results = futures::future::join_all(jobs).await;
+
+        for node in other_nodes {
+            // Execute non-Fetch nodes sequentially (TODO!: consider parallel execution)
```
dotansimha: i wonder if this TODO should already be addressed in this PR?

ardatan: I removed cloning and did a refactor in the structure based on mutable references which ended up this. I'll cover it in an followup PR not to block this one.

lib/query-plan-executor/src/executors/http.rs
kamilkisiela: there can be ~180 clients or like in Netflix ~400

bench/subgraphs/lib.rs
ardatan: We use it for local execution in the tests.

lib/query-plan-executor/src/executors/local.rs
ardatan: It is not a mock really. It is an executor for local schemas.

lib/query-plan-executor/src/lib.rs
ardatan: Anyway I brought the parallelism back

lib/query-plan-executor/src/executors/http.rs
ardatan: They say this in their docs;

> The Client holds a connection pool internally, so it is advised that you create one and reuse it.

> You do not have to wrap the Client in an [Rc](https://doc.rust-lang.org/nightly/alloc/rc/struct.Rc.html) or [Arc](https://doc.rust-lang.org/nightly/alloc/sync/struct.Arc.html) to reuse it, because it already uses an [Arc](https://doc.rust-lang.org/nightly/alloc/sync/struct.Arc.html) internally.

bench/subgraphs/lib.rs
ardatan: Here;
https://github.com/graphql-hive/gateway-rs/pull/152/files#diff-8f5f51491c4b08147fbace137cc08c655cad4eb0b02e80689af66debeac9ad8cR70

PR #160, opened 2025-06-12
Fix operation kind for Mutation and Subscription

kamilkisiela:


lib/query-planner/src/tests/object_entities.rs
```diff
@@ -77,7 +77,7 @@ fn testing() -> Result<(), Box<dyn Error>> {
             "kind": "Fetch",
             "serviceName": "store",
             "operationKind": "query",
-            "operation": "{products{__typename id}}"
+            "operation": "query{products{__typename id}}"
```
ardatan, line 80: We can remove `query` for queries without variables, no? Maybe too much micro optimization. What do you think?

kamilkisiela: too much hassle (code wise) for 5 less characters imo

kamilkisiela: https://github.com/graphql-hive/gateway-rs/issues/163

ardatan: Was just nitpicking :D

PR #175, opened 2025-06-17
Refactor Executors

ardatan:
Now each subgraph has its own executor, and later on we can combine executors like WS + HTTP etc.

Removed extra `TestExecutor` implementation, so now LocalSubgraphExecutor is used for local schemas.

### DIFF
--- a/lib/query-plan-executor/src/executors/http.rs
+++ b/lib/query-plan-executor/src/executors/http.rs
@@ -21,7 +21,7 @@ impl HTTPSubgraphExecutor {
         &self,
         execution_request: ExecutionRequest,
     ) -> Result<ExecutionResult, reqwest::Error> {
-        println!("Executing HTTP request to subgraph at {}", self.endpoint);
+        trace!("Executing HTTP request to subgraph at {}", self.endpoint);
         self.http_client
             .post(&self.endpoint)
             .json(&execution_request)
@@ -34,7 +34,7 @@ impl HTTPSubgraphExecutor {
 
 #[async_trait]
 impl SubgraphExecutor for HTTPSubgraphExecutor {
-    #[instrument(level = "trace", skip(self), name = "http_subgraph_execute")]
+    #[instrument(level = "trace", skip(self), name = "http_subgraph_execute", fields(endpoint = %self.endpoint))]
     async fn execute(&self, execution_request: ExecutionRequest) -> ExecutionResult {
         self._execute(execution_request).await.unwrap_or_else(|e| {
             error!("Failed to execute request to subgraph: {}", e);
--- a/bin/gateway/src/pipeline/execution_service.rs
+++ b/bin/gateway/src/pipeline/execution_service.rs
@@ -18,7 +18,6 @@ use http::{Request, Response};
 use query_plan_executor::execute_query_plan;
 use serde_json::to_value;
 use tower::Service;
-use tracing::trace;
 
 #[derive(Clone, Debug, Default)]
 pub struct ExecutionService {
--- a/lib/query-plan-executor/src/executors/map.rs
+++ b/lib/query-plan-executor/src/executors/map.rs
@@ -21,7 +21,7 @@ impl SubgraphExecutorMap {
         }
     }
 
-    #[instrument(level = "trace", name = "subgraph_execute", skip_all)]
+    #[instrument(level = "trace", name = "subgraph_execute", skip_all, fields(subgraph_name = %subgraph_name, execution_request = ?execution_request))]
     pub async fn execute(
         &self,
         subgraph_name: &str,
--- a/lib/query-plan-executor/src/lib.rs
+++ b/lib/query-plan-executor/src/lib.rs
@@ -207,6 +207,7 @@ impl ExecutableFetchNode for FetchNode {
             extensions: None,
         };
         let mut fetch_result = execution_context
+            .subgraph_executor_map
             .execute(&self.service_name, execution_request)
             .await;
 
@@ -286,6 +287,7 @@ impl ExecutableFetchNode for FetchNode {
 
         // 3. Execute the fetch operation
         let fetch_result = execution_context
+            .subgraph_executor_map
             .execute(&self.service_name, execution_request)
             .await;
 
@@ -327,7 +329,7 @@ impl ExecutableFetchNode for FetchNode {
 
     #[instrument(
         level = "debug",
-        skip(variable_values),
+        skip(self, variable_values),
         name = "prepare_variables_for_fetch_node"
     )]
     fn prepare_variables_for_fetch_node(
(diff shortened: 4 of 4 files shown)


bin/gateway/src/main.rs
```diff
@@ -26,7 +26,8 @@ struct AppState {
     schema_metadata: SchemaMetadata,
     planner: Planner,
     validation_plan: graphql_tools::validation::validate::ValidationPlan,
-    executor: query_plan_executor::executors::http::HTTPSubgraphExecutor,
+    subgraph_endpoint_map: HashMap<String, String>,
+    subgraph_executor_map: query_plan_executor::SubgraphExecutorMap<'static>,
```
dotansimha: why does it needs to be `'static` ?

lib/query-plan-executor/benches/executor_benches.rs
```diff
@@ -77,15 +89,25 @@ fn query_plan_execution_without_projection_via_http(c: &mut Criterion) {
...
     let schema_metadata = planner.consumer_schema.schema_metadata();
-    let executor = HTTPSubgraphExecutor::new(subgraph_endpoint_map);
+    let mut subgraph_executor_map: SubgraphExecutorMap = HashMap::new();
+    let http_client = Arc::new(reqwest::Client::new());
+    for (name, endpoint) in subgraph_endpoint_map {
+        subgraph_executor_map.insert(
+            name.to_string(),
+            Arc::new(Box::new(HTTPSubgraphExecutor::new(
+                &endpoint,
+                http_client.clone(),
```
dotansimha: @ardatan maybe try to create a new `Arc::new(reqwest::Client::new())` here. 
I think if it's shared, it will share the whole pool. I think we are ok with creating a client for each subgraph and have it's own pool managed?

lib/query-plan-executor/src/executors/async_graphql.rs
```diff
@@ -54,8 +40,8 @@ impl From<ExecutionRequest> for async_graphql::Request {
     }
 }
 
-impl From<&ServerError> for GraphQLError {
-    fn from(error: &ServerError) -> Self {
+impl From<&async_graphql::ServerError> for GraphQLError {
```
dotansimha, line 43: Hmm, async-graphql is a separate experiment, why is it used here?

lib/query-plan-executor/benches/executor_benches.rs
ardatan: We already discussed it and we decided not to create a new http client for each subgraph. 

https://github.com/graphql-hive/gateway-rs/pull/152#discussion_r2145564430

lib/query-plan-executor/src/executors/async_graphql.rs
ardatan: The experiment iscovered in other draft PRs. This improves the existing LocalSubgraphExecutor.

bin/gateway/src/main.rs
ardatan: It is 'a now 👍

bin/gateway/src/shared_state.rs
```diff
@@ -1,42 +1,56 @@
...
     state::supergraph_state::SupergraphState,
 };
 
-pub struct GatewaySharedState {
+pub struct GatewaySharedState<'a> {
     pub schema_metadata: SchemaMetadata,
     pub planner: Planner,
     pub validation_plan: ValidationPlan,
-    pub executor: HTTPSubgraphExecutor,
+    pub subgraph_executor_map: SubgraphExecutorMap<'a>,
```
dotansimha: Why do we need a lifetime here?

dotansimha: I wonder if we really need a map of endpoints? 
I mean, the `SubgraphExecutorMap` can handle that maybe?

bin/gateway/src/shared_state.rs
```diff
@@ -1,42 +1,56 @@
...
             Planner::new_from_supergraph(&parsed_supergraph_sdl).expect("failed to create planner");
         let schema_metadata = planner.consumer_schema.schema_metadata();
+        let http_client = reqwest::Client::new();
+        let http_client_arc = Arc::new(http_client);
 
-        let executor = query_plan_executor::executors::http::HTTPSubgraphExecutor::new(
-            supergraph_state.subgraph_endpoint_map,
-        );
+        let subgraph_endpoint_map = supergraph_state.subgraph_endpoint_map.clone();
+        let mut subgraph_executor_map: SubgraphExecutorMap = HashMap::new();
```
dotansimha: Let's adopt better Rust approaches here. Try to use `iter` tools and build this without creating a mutable HashMap and iterate this. It's a good practice for iter tools

lib/query-plan-executor/src/executors/common.rs
```diff
@@ -4,9 +4,7 @@ use crate::{ExecutionRequest, ExecutionResult};
...
 pub trait SubgraphExecutor {
-    async fn execute(
-        &self,
-        subgraph_name: &str,
-        execution_request: ExecutionRequest,
-    ) -> ExecutionResult;
+    async fn execute(&self, execution_request: ExecutionRequest) -> ExecutionResult;
 }
+
+pub type SubgraphExecutorType<'a> = dyn SubgraphExecutor + Send + Sync + 'a;
```
dotansimha: This (+ the init logic) can maybe be in a dedicated `struct SubgraphExecutorMap` and not in a type alias. Maybe this new struct can handle things in a cleaner way and not inside the factory fn. 

What I mean is, if we have `sturct SubgraphExecutorMap`, it can handle creation, lifecycle, endpoints, and other parts, and encapsulate the logic inside.

lib/query-plan-executor/src/executors/http.rs
```diff
@@ -7,38 +7,29 @@ use crate::{executors::common::SubgraphExecutor, ExecutionRequest, ExecutionResu
...
 pub struct HTTPSubgraphExecutor {
-    pub subgraph_endpoint_map: HashMap<String, String>,
-    http_client: reqwest::Client,
+    pub endpoint: String,
+    pub http_client: Arc<reqwest::Client>,
 }
 
 impl HTTPSubgraphExecutor {
-    pub fn new(subgraph_endpoint_map: HashMap<String, String>) -> Self {
+    pub fn new(endpoint: &str, http_client: Arc<reqwest::Client>) -> Self {
```
dotansimha: I wonder if this `reqwest::Client` should just created inside here and not passed around? 
Any reason to have it shared and not allow each executor to have its own client?

ardatan: I already covered it here;
https://github.com/graphql-hive/gateway-rs/pull/175#discussion_r2154749361

PR #183, opened 2025-06-18
Experiment: sonic

kamilkisiela:


lib/query-plan-executor/src/deep_merge.rs
```diff
@@ -1,44 +1,43 @@
...
-        // 4. Fallback: Source is not Null, and cases 2 & 3 didn't match. Replace target with source.
-        (target_val, source) => {
-            // source is guaranteed not Null here due to arm 1
-            *target_val = source;
+    // 3. Both are Arrays of same length(?): Merge elements
+    if target.is_array() && source.is_array() {
+        let target_arr = target.as_array_mut().unwrap();
+        let source_arr = source.as_array().unwrap();
+        for (target_val, source_val) in target_arr.iter_mut().zip(source_arr) {
+            deep_merge(target_val, source_val.to_owned());
```
ardatan, line 24: `to_owned` generally clones the object which comes with a cost, if I don't miss anything.

PR #220, opened 2025-06-30
Fix requires-with-argument / handle variables correctly on serialization

ardatan:
Also merged this into this PR by accident  #221
Sorry :/

### DIFF
--- a/lib/query-plan-executor/src/projection.rs
+++ b/lib/query-plan-executor/src/projection.rs
@@ -151,6 +151,9 @@ fn project_selection_set(
             // If first is mutated, it means we added "{"
             if !first {
                 buffer.push('}');
+            } else {
+                // If first is still true, it means we didn't add anything, so we should just send an empty object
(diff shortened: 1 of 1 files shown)


lib/query-plan-executor/src/executors/http.rs
```diff
@@ -23,27 +25,39 @@ impl HTTPSubgraphExecutor {
...
+                if first_variable {
+                    body.push_str(FIRST_VARIABLE_STR);
+                    first_variable = false;
+                } else {
+                    body.push(',');
+                }
+                body.push('"');
+                body.push_str(variable_name);
+                body.push_str("\":");
+                let value_str = serde_json::to_string(variable_value).unwrap();
```
dotansimha: Can we avoid the `unwrap`? or is it fully safe here since we created the variable value? 
should it propagate?

ardatan: Done!

PR #224, opened 2025-06-30
Fix simple-interface-object audit / fix possible type check

ardatan:
Also includes #226 & #229 & #245

### DIFF
--- a/bin/gateway/src/main.rs
+++ b/bin/gateway/src/main.rs
@@ -98,9 +98,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .layer(GraphiQLResponderService::new_layer())
         .layer(GraphQLRequestParamsExtractor::new_layer())
         .layer(GraphQLParserService::new_layer())
+        .layer(GraphQLValidationService::new_layer())
         .layer(GraphQLOperationNormalizationService::new_layer())
         .layer(CoerceVariablesService::new_layer())
-        .layer(GraphQLValidationService::new_layer())
         .layer(QueryPlanService::new_layer())
         .layer(PropagateRequestIdLayer::new(REQUEST_ID_HEADER_NAME.clone()))
         .service(ExecutionService::new(expose_query_plan));
--- a/bin/gateway/src/pipeline/coerce_variables_service.rs
+++ b/bin/gateway/src/pipeline/coerce_variables_service.rs
@@ -2,10 +2,11 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use axum::body::Body;
-use http::Request;
+use http::{Method, Request};
 use query_plan_executor::variables::collect_variables;
+use query_planner::state::supergraph_state::OperationKind;
 use serde_json::Value;
-use tracing::{trace, warn};
+use tracing::{error, trace, warn};
 
 use crate::pipeline::error::{PipelineError, PipelineErrorVariant};
 use crate::pipeline::gateway_layer::{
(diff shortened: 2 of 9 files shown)


lib/query-plan-executor/src/projection.rs
```diff
@@ -251,37 +253,33 @@ fn project_selection_set_with_map(
...
+                    continue;
                 }
             }
             SelectionItem::InlineFragment(inline_fragment) => {
                 let type_condition = &inline_fragment.type_condition;
-                let satisfies_type_condition = &type_name == type_condition
-                    || possible_types_of_type.is_some_and(|s| s.contains(type_condition));
 
-                if satisfies_type_condition {
+                if schema_metadata
```
ardatan, line 268: This was wrong before. Because it should check if the inline fragment's type condition is satisfied with the object's typename not other way around.

lib/query-plan-executor/src/schema_metadata.rs
```diff
@@ -98,7 +117,9 @@ impl SchemaWithMetadata for ConsumerSchema {
         let introspection_schema_root_json = json!(introspection_query.__schema);
 
         SchemaMetadata {
-            possible_types: final_possible_types,
+            possible_types: PossibleTypes {
```
dotansimha, line 118: Can be `PossibleTypes::new(final_possible_types)` but it's just nitpicking :)

lib/query-plan-executor/src/schema_metadata.rs
```diff
@@ -9,12 +9,31 @@ use serde_json::{json, Value};
...
+
+impl PossibleTypes {
+    pub fn entity_satisfies_type_condition(&self, type_name: &str, type_condition: &str) -> bool {
+        if type_name == type_condition {
+            true
+        } else if let Some(possible_types_of_type) = self.map.get(type_condition) {
+            possible_types_of_type.contains(type_name)
+        } else {
+            false
+        }
```
dotansimha, line 31: nit: This could be a `match`, but it's just for readability. Nothing too crucial.

lib/query-plan-executor/src/schema_metadata.rs
```diff
@@ -9,12 +9,31 @@ use serde_json::{json, Value};
...
 #[derive(Debug)]
 pub struct SchemaMetadata {
-    pub possible_types: HashMap<String, HashSet<String>>,
+    pub possible_types: PossibleTypes,
     pub enum_values: HashMap<String, Vec<String>>,
     pub type_fields: HashMap<String, HashMap<String, String>>,
     pub introspection_schema_root_json: Value,
 }
 
+impl SchemaMetadata {}
```
dotansimha: empty? delete?

lib/query-plan-executor/src/schema_metadata.rs
dotansimha: Or actually, why don't move the logic of creating that `HashMap` into a `fn new` of the `struct`? 
I think it just makes sense to have it managed and created in a single place?

lib/query-plan-executor/src/schema_metadata.rs
ardatan: Can it really be because I am not sure if two conditions can be mixed?

lib/query-plan-executor/src/schema_metadata.rs
ardatan: Yeah, extra! Removed!

lib/query-plan-executor/src/schema_metadata.rs
ardatan: The schema visit logic is held by SchemaMetadata, so if we move it inside PossibleTypes, it will have another round of visit. So we visit the schema once here and put the map inside PossibleTypes. Not sure if `new` is needed because this is only used by this crate/module. Even `map` is private not public.

PR #226, opened 2025-07-01
Refactor parallelism in ParallelNode

ardatan:
Avoid `join_all` and use `FuturesUnordered` in order to process the done tasks without waiting all tasks to finish.
Execute and process fetch and flatten nodes in parallel
It doesn't give much boost but it does better parallelism(if my results are correct)

`simple-interface-object`
833.345514/s
927.135515/s
929.957448/s

This branch
935.630703/s
941.817198/s
946.55521/s

### DIFF
--- a/lib/query-plan-executor/src/lib.rs
+++ b/lib/query-plan-executor/src/lib.rs
@@ -1,5 +1,5 @@
 use async_trait::async_trait;
-use futures::{stream::FuturesUnordered, FutureExt, StreamExt};
+use futures::{future::BoxFuture, stream::FuturesUnordered, FutureExt, StreamExt};
 use query_planner::{
     ast::{operation::OperationDefinition, selection_item::SelectionItem},
     planner::plan_nodes::{
@@ -9,7 +9,7 @@ use query_planner::{
 };
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
-use std::{collections::BTreeSet, fmt::Write, future::Future, pin::Pin};
+use std::{collections::BTreeSet, fmt::Write};
 use std::{collections::HashMap, vec};
 use tracing::{instrument, trace, warn}; // For reading file in main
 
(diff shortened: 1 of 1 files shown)


lib/query-plan-executor/src/lib.rs
```diff
@@ -431,156 +437,143 @@ impl ExecutablePlanNode for ParallelNode {
...
-                    let normalized_path: Vec<&str> =
-                        flatten_node.path.iter().map(String::as_str).collect();
-                    let mut filtered_representations = String::with_capacity(1024);
-                    let fetch_node = match flatten_node.node.as_ref() {
-                        PlanNode::Fetch(fetch_node) => fetch_node,
-                        _ => {
-                            warn!(
+        {
+            // Here we call fetch nodes in parallel and non-fetch nodes sequentially.
+            let mut jobs: FuturesUnordered<Pin<Box<dyn Future<Output = ParallelJob> + Send>>> =
```
dotansimha: Simplify with `type` declaration? or try to see if it's can be inferred?

lib/query-plan-executor/src/lib.rs
```diff
@@ -431,156 +437,143 @@ impl ExecutablePlanNode for ParallelNode {
...
-            }
-            // Extend errors and extensions from the result
-            if let Some(errors) = result.errors {
-                all_errors.extend(errors);
-            }
-            if let Some(extensions) = result.extensions {
-                all_extensions.push(extensions);
+                    ParallelJob::Flatten((result, path)) => {
+                        // Process FlattenNode results
+                        // Process FlattenNode results
```
kamilkisiela:
```suggestion
                        // Process FlattenNode results
                        // Process FlattenNode results
```

ardatan: My old comments 😅 most of them are wrong! I'll clean them

lib/query-plan-executor/src/lib.rs
ardatan: Done!

lib/query-plan-executor/src/lib.rs
ardatan: Removed!

PR #229, opened 2025-07-01
Avoid extra allocation/clone in SubgraphExecutionRequest

ardatan:
Avoid cloning variables and operation string(aka `query`) for subgraph requests

### DIFF
--- a/lib/query-plan-executor/src/executors/http.rs
+++ b/lib/query-plan-executor/src/executors/http.rs
@@ -1,7 +1,10 @@
 use async_trait::async_trait;
 use tracing::{error, instrument, trace};
 
-use crate::{executors::common::SubgraphExecutor, ExecutionRequest, ExecutionResult};
+use crate::{
+    executors::common::SubgraphExecutor, json_writer::write_and_escape_string, ExecutionResult,
+    SubgraphExecutionRequest,
+};
 
 #[derive(Debug)]
 pub struct HTTPSubgraphExecutor {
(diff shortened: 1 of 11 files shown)


lib/query-plan-executor/src/executors/http.rs
```diff
@@ -19,14 +22,16 @@ impl HTTPSubgraphExecutor {
...
+        execution_request: SubgraphExecutionRequest<'a>,
     ) -> Result<ExecutionResult, reqwest::Error> {
         trace!("Executing HTTP request to subgraph at {}", self.endpoint);
 
-        let mut body =
-            "{\"query\":".to_string() + &serde_json::to_string(&execution_request.query).unwrap();
+        // We may want to remove it, but let's see.
+        let mut body = String::with_capacity(4096);
+        body.push_str("{\"query\":");
+        write_and_escape_string(&mut body, execution_request.query);
```
dotansimha, line 34: nice!

lib/query-plan-executor/src/executors/http.rs
```diff
@@ -19,14 +22,16 @@ impl HTTPSubgraphExecutor {
...
         &self,
-        execution_request: ExecutionRequest,
+        execution_request: SubgraphExecutionRequest<'a>,
     ) -> Result<ExecutionResult, reqwest::Error> {
         trace!("Executing HTTP request to subgraph at {}", self.endpoint);
 
-        let mut body =
-            "{\"query\":".to_string() + &serde_json::to_string(&execution_request.query).unwrap();
+        // We may want to remove it, but let's see.
+        let mut body = String::with_capacity(4096);
```
dotansimha, line 32: 4096 is generally enough? how did we end up with this value?

kamilkisiela: guess

kamilkisiela: it will never be correct

PR #232, opened 2025-07-02
Implement minification of subgraph calls

kamilkisiela:
- Use a generic `Text` type from `graphql_parser` in `ArgumentsMap`, `OperationDefinition`, `VariableDefinition`, and `SelectionItem`. This allows for more flexible parsing.
- Refactor `ASTHash` to use a generic hasher

Closes https://github.com/graphql-hive/gateway-rs/issues/151

### DIFF
--- a/lib/query-planner/src/ast/selection_set.rs
+++ b/lib/query-planner/src/ast/selection_set.rs
@@ -96,8 +96,7 @@ impl Serialize for SelectionSet {
     }
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize)]
-#[derive(Default)]
+#[derive(Clone, Debug, Deserialize, Serialize, Default)]
 pub struct FieldSelection {
     pub name: String,
     #[serde(skip_serializing_if = "SelectionSet::is_empty")]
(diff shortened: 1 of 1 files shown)


lib/query-planner/src/ast/minification/selection_id.rs
```diff
@@ -0,0 +1,14 @@
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
+
+use crate::ast::hash::ASTHash;
+use crate::ast::selection_set::SelectionSet;
+
+pub type SelectionId = u64;
+
+pub fn generate_selection_id(type_name: &str, selection_set: &SelectionSet) -> SelectionId {
```
dotansimha, line 9: nit: should this be something like a fn directly on the `SelectionItem` or its variants? (I'm asking because i can see it useful and it might be easier to accesS)

kamilkisiela: `SelectionSet` does not have info about `type_name`

PR #255, opened 2025-07-10
Progressive Override

kamilkisiela:
- [x] QP
- [x] Executor

**Caching**
If we compute the percentage value on every request, or add new/modify non-used flags and include that in the cache key (of normalization cache) then the cache hit/miss ratio will be super low.

That's why I decided to extract all labels and percentage values used in the supergraph, to then compare that with random percentage value and active flags generated for the request.
This way we will have a stable cache key and keep cache hit/miss ratio as high as we can.

### DIFF
--- a/lib/query-planner/src/federation_spec/join_field.rs
+++ b/lib/query-planner/src/federation_spec/join_field.rs
@@ -1,6 +1,9 @@
 use graphql_parser::schema::{Directive, Value};
 
-use crate::{graph::edge::OverrideLabel, state::supergraph_state::TypeNode};
+use crate::{
+    graph::{edge::OverrideLabel, PERCENTAGE_SCALE_FACTOR},
+    state::supergraph_state::TypeNode,
+};
 
 use super::directives::FederationDirective;
 
@@ -66,7 +69,9 @@ impl JoinFieldDirective {
                     // (11.12 becomes 1112000000.0)
                     // Cast to u64 for storage
                     // (11120.0 becomes 1112000000)
-                    return OverrideLabel::Percentage((value * 100_000_000.0) as u64);
+                    return OverrideLabel::Percentage(
+                        (value * (PERCENTAGE_SCALE_FACTOR as f64)) as u64,
+                    );
                 }
                 Err(error) => {
                     panic!(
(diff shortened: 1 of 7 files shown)


lib/query-planner/src/federation_spec/join_field.rs
```diff
@@ -27,13 +28,55 @@ impl Default for JoinFieldDirective {
...
+            };
+
+            if !is_precision_valid {
+                panic!("Invalid precision for percentage override. Must be no more than 8 fraction digits.");
+            }
+
+            match value_str.parse::<f64>() {
+                Ok(value) => {
+                    if !(0.0..=100.0).contains(&value) {
+                        panic!("Invalid percentage value. Must be between 0 and 100.");
```
dotansimha, line 64: panic feels too aggressive? 👀

kamilkisiela: composition won't allow for such values

kamilkisiela: and I did not want to turn it into a Result. We can add a todo if you feel like it

bin/gateway/src/pipeline/progressive_override_service.rs
```diff
@@ -0,0 +1,63 @@
...
+use axum::body::Body;
+use http::Request;
+use query_planner::graph::{PlannerOverrideContext, PERCENTAGE_SCALE_FACTOR};
+use rand::Rng;
+
+use super::error::PipelineError;
+use super::gateway_layer::{GatewayPipelineLayer, GatewayPipelineStepDecision, ProcessorLayer};
+
+/// Contains the request-specific context for progressive overrides.
+/// This will be stored in the request extensions.
```
kamilkisiela:
```suggestion
/// This is stored in the request extensions
```

lib/query-planner/src/federation_spec/join_field.rs
kamilkisiela, line 64:
```suggestion
                        // TODO: convert it into Err
                        panic!("Invalid percentage value. Must be between 0 and 100.");
```

lib/query-planner/src/ast/operation.rs
```diff
@@ -139,21 +139,19 @@ impl Display for OperationDefinition {
...
             if !variable_definitions.is_empty() {
-                write!(f, "(")?;
-                let len = variable_definitions.len();
-                for (i, variable_definition) in variable_definitions.iter().enumerate() {
-                    let is_last = i == len - 1;
-                    write!(f, "{}", variable_definition)?;
-
-                    if !is_last {
-                        write!(f, ", ")?;
+                f.write_str("(")?;
```
dotansimha, line 142: ❤️  much more readable with `peek`

PR #259, opened 2025-07-10
feat(multi-type-step): make `FetchStep` compatible with multiple definitions

dotansimha:
- [x] figure out how to construct multi-type selections for `output`
- [x] adjust `Display` for `output`
- [x] confirm `output` builds the same way (no panics)
- [ ] Remove existing `output` + Adjust optimizations (https://github.com/graphql-hive/gateway-rs/pull/270)

`...`

- [ ] figure out how to construct multi-type selections for `input`


lib/query-planner/src/planner/fetch/fetch_graph.rs
```diff
@@ -110,10 +111,6 @@ impl FetchGraph {
         Ok(self.graph.index_twice_mut(index1, index2))
     }
 
-    #[instrument(level = "trace",skip_all, fields(
```
kamilkisiela, line 113: oh I liked it

dotansimha: i can bring it back 😄

PR #260, opened 2025-07-10
Cover last 2 failing test cases in the audit

kamilkisiela:
By definition `@shareable` means: QP can pick any field to resolve data, it shouldn't matter which one is used.
Performing type expansion and fetching data from two subgraphs breaks that rule.

### DIFF
--- a/lib/query-planner/src/tests/interface.rs
+++ b/lib/query-planner/src/tests/interface.rs
@@ -457,33 +457,6 @@ fn node_query_with_multiple_type_fragments() -> Result<(), Box<dyn Error>> {
 
     assert!(query_plan.is_err());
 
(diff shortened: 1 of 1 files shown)


lib/query-planner/src/graph/tests.rs
```diff
@@ -366,7 +366,7 @@ mod graph_tests {
         let (viewed_incoming, viewed_outgoing) = find_node(&graph, &node1.display_name());
         viewed_outgoing.assert_field_edge("id", "String/foo");
         assert_eq!(viewed_incoming.edges.len(), 1);
-        assert_eq!(viewed_outgoing.edges.len(), 1);
+        assert_eq!(viewed_outgoing.edges.len(), 2);
```
dotansimha, line 369: these tests can be killed in favor of some kind of a snapshot

PR #261, opened 2025-07-11
Refactor GW layers and avoid cloning

ardatan:


bin/gateway/src/pipeline/coerce_variables_service.rs
```diff
@@ -36,8 +36,8 @@ impl GatewayPipelineLayer for CoerceVariablesService {
     #[tracing::instrument(level = "trace", name = "CoerceVariablesService", skip_all)]
     async fn process(
         &self,
-        mut req: Request<Body>,
-    ) -> Result<(Request<Body>, GatewayPipelineStepDecision), PipelineError> {
+        req: &mut Request<Body>,
```
dotansimha, line 39: This actually worked, also with lifetimes? I remember when trying to deal with it, the best recommendation I saw that actually worked, it to take ownership and consume the request (with `parts`) and then re-construct it.

ardatan: I deliberately avoid taking ownership and reconstructing(which is cloning technically) to reduce the overhead and it works.

PR #279, opened 2025-07-16
Fix union overfetching

kamilkisiela:
by adding type conditions to the FetchNodes

summary of changes -> https://github.com/graphql-hive/gateway-rs/pull/279#issuecomment-3083438920

Closes #277

### DIFF
--- a/lib/query-planner/src/tests/arguments.rs
+++ b/lib/query-planner/src/tests/arguments.rs
@@ -108,7 +108,7 @@ fn fed_audit_requires_with_argument_conflict() -> Result<(), Box<dyn Error>> {
                   {
                     "Field": "products"
                   },
-                  "List"
+                  "@"
                 ],
                 "node": {
                   "kind": "Fetch",
@@ -158,7 +158,7 @@ fn fed_audit_requires_with_argument_conflict() -> Result<(), Box<dyn Error>> {
                   {
                     "Field": "products"
                   },
-                  "List"
+                  "@"
                 ],
                 "node": {
                   "kind": "Fetch",
@@ -314,7 +314,7 @@ fn requires_arguments_deeply_nested_requires() -> Result<(), Box<dyn Error>> {
               {
                 "Field": "feed"
               },
-              "List"
+              "@"
             ],
             "node": {
               "kind": "Fetch",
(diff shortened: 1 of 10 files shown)


lib/query-plan-executor/src/lib.rs
```diff
@@ -1033,26 +1033,77 @@ fn entity_satisfies_type_condition(
...
+                Value::Array(arr) => {
+                    // Filter an array based on matching `__typename`
+                    arr.iter_mut()
+                        .flat_map(|item| {
+                            if contains_typename(item, type_name, false) {
+                                traverse_and_collect(item, remaining_path)
+                            } else {
+                                vec![]
+                            }
+                        })
```
kamilkisiela:
```suggestion
                        .filter(|item| contains_typename(item, type_name, false))
                        .flat_map(|item| traverse_and_collect(item, remaining_path))
```

lib/query-planner/src/tests/arguments.rs
```diff
@@ -336,10 +342,14 @@ fn requires_arguments_deeply_nested_requires() -> Result<(), Box<dyn Error>> {
           {
             "kind": "Flatten",
             "path": [
-              "feed",
-              "@",
-              "_internal_qp_alias_0",
-              "@"
+              {
```
dotansimha, line 345: these changes are intended right? 
I mean, what's the benefit of `"List"` over `"@"`?

kamilkisiela: none, it's just the `Serialized` version the enum

lib/query-planner/src/tests/requires_provides.rs
```diff
@@ -126,10 +128,16 @@ fn simple_requires_provides() -> Result<(), Box<dyn Error>> {
...
+                "Field": "me"
+              },
+              {
+                "Field": "reviews"
+              },
               "@",
-              "product"
+              {
+                "Field": "product"
+              }
```
ardatan, line 140: Can't we also get rid of InlineFragment in the require selection set then move that type condition to here?
So instead of having `... on Product` in requires selections, we can have `{ Cast: Product }` in the flatten node's path directly, no?

PR #299, opened 2025-07-28
feat(executor, gw): error propagation

ardatan:
Reverts graphql-hive/gateway-rs#298

### DIFF
--- a/lib/query-plan-executor/src/lib.rs
+++ b/lib/query-plan-executor/src/lib.rs
@@ -10,7 +10,7 @@ use query_planner::{
 };
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
-use std::collections::BTreeSet;
+use std::collections::VecDeque;
 use std::{collections::HashMap, vec};
 use tracing::{instrument, trace, warn}; // For reading file in main
 
@@ -21,6 +21,7 @@ use crate::{
     schema_metadata::{PossibleTypes, SchemaMetadata},
 };
 pub mod deep_merge;
+mod error_normalization;
 pub mod executors;
 pub mod introspection;
 mod json_writer;
@@ -99,9 +100,10 @@ fn process_errors_and_extensions(
         execution_context.extensions.extend(extensions);
     }
 }
+
 struct ExecuteForRepresentationsResult {
     entities: Option<Vec<Value>>,
-    indexes: BTreeSet<usize>,
+    indexes_in_paths: Vec<VecDeque<usize>>,
     errors: Option<Vec<GraphQLError>>,
     extensions: Option<HashMap<String, Value>>,
 }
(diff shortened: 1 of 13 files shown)


lib/query-plan-executor/src/lib.rs
```diff
@@ -509,15 +510,21 @@ impl ExecutablePlanNode for ParallelNode {
                             }
                         };
                         let requires_nodes = fetch_node.requires.as_ref().unwrap();
-                        let mut index = 0;
-                        let mut indexes = BTreeSet::new();
-                        let normalized_path = flatten_node.path.as_slice();
                         filtered_representations.push(b'[');
+                        let mut number_of_indexes = 0;
+                        let normalized_path = flatten_node.path.as_slice();
```
dotansimha, line 515: `let number_of_indexes = flatten_node.path.as_slice().iter().filter(|v| matches!(v, FlattenNodePathSegment::List)).count()`

lib/query-plan-executor/src/lib.rs
```diff
@@ -528,37 +535,33 @@ impl ExecutablePlanNode for ParallelNode {
...
-                                    };
+                                        execution_context.project_requires(
+                                            &requires_nodes.items,
+                                            entity,
+                                            &mut filtered_representations,
+                                            indexes_in_paths.is_empty(),
+                                            None,
+                                        )
+                                    }
+                                    .unwrap_or(false);
```
dotansimha: This `unwrap_or` hides an error inside? 
I mean, if this fail do we need some printing to the log? what's the scenario where an error is returned from here, and why the error is not relevant and dropped in favor of a default value?

lib/query-plan-executor/src/lib.rs
```diff
@@ -589,29 +592,57 @@ impl ExecutablePlanNode for ParallelNode {
...
-                                &mut |target| {
-                                    if result.indexes.contains(&index_of_traverse) {
-                                        let entity =
-                                            entities.get_mut(index_of_entities).unwrap().take();
-                                        // Merge the entity into the target
-                                        deep_merge::deep_merge(target, entity);
-                                        index_of_entities += 1;
+                    ParallelJob::Flatten(mut result, path) => {
+                        if let Some(entities) = result.entities {
+                            'entity_loop: for (entity, indexes_in_path) in
```
dotansimha, line 611: confirming this part for now, but with the new refactor i wonder if this section of code can be more readable?

lib/query-plan-executor/src/lib.rs
```diff
@@ -666,11 +697,19 @@ impl ExecutablePlanNode for FlattenNode {
         let requires_nodes = fetch_node.requires.as_ref().unwrap();
         filtered_representations.push(b'[');
         let mut first = true;
+        let normalized_path = self.path.as_slice();
```
dotansimha, line 714: Same suggestion as before, it can be simplified.

lib/query-plan-executor/src/lib.rs
ardatan: Hmm not sure if this is optimal enough. With this usage, it will create a new set of path segments then count it, no? But I avoid to create a new set/vec and count them directly with a single iteration.

lib/query-plan-executor/src/lib.rs
ardatan: I added a `warn!` to print the error message when it happens. It shouldn't happen since we write to a string buffer but just in case it fails.

lib/query-plan-executor/src/lib.rs
ardatan: I left comments why they are different.

lib/query-plan-executor/src/lib.rs
ardatan: I left comments for the one above.

PR #330, opened 2025-08-12
New query plan execution (for axum)

kamilkisiela:


lib/executor/src/execution/plan.rs
```diff
@@ -0,0 +1,561 @@
...
+                return None;
+            }
+        };
+
+        ctx.handle_errors(response.errors);
+
+        // The `response.data` has a lifetime tied to `deserializer`, which is tied to `bytes`.
+        // Since `bytes` has been transmuted to `'a`, `response.data` can also be considered
+        // to have lifetime `'a`.
+        let data: Value<'exec> = unsafe { std::mem::transmute(response.data) };
```
dotansimha: I think this `transmute` + `unsafe` can be removed?

lib/executor/src/execution/plan.rs
```diff
@@ -0,0 +1,561 @@
...
+                    }
+
+                    deep_merge(&mut ctx.final_response, data);
+                }
+            }
+            ExecutionJob::FlattenFetch(job) => {
+                if let Some((mut data, output_rewrites)) =
+                    self.process_subgraph_response(ctx, job.response, job.fetch_node_id)
+                {
+                    if let Some(entities) = data.take_entities() {
```
dotansimha: this can be `if let Some(mut entities) = data.take_entities() {` and you can remove the `transmute` and `unsafe`

PR #337, opened 2025-08-14
feat(gw): http connection pooling with semaphore + deduplication for subgraph reqs

kamilkisiela:


lib/executor/src/executors/config.rs
```diff
@@ -0,0 +1,27 @@
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct HttpExecutorConfig {
+    #[serde(default = "default_max_connections_per_host")]
+    pub max_connections_per_host: usize,
```
dotansimha, line 6: If I understand correctly, the `max_connections_per_host` configuration is used for two things:
1 - the `pool_max_idle_per_host` 
2 - the Semaphore itself 

Is it intended? Or should we consider to split it to two different configurations? (`max_connections_per_host` for the http-client and `max_concurrent_subgraph_requests` for the semaphore counter)

kamilkisiela: It’s done per host, see semaphores_by_origin

PR #338, opened 2025-08-14
Implement request deduplication for subgraphs

kamilkisiela:


lib/executor/src/executors/dedupe.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub status: StatusCode,
+    pub headers: HeaderMap,
+    pub body: Bytes,
+}
+
+#[derive(Debug, Clone, Eq)]
+pub struct RequestFingerprint {
+    method: Method,
+    url: Uri,
+    /// BTreeMap to ensure case-insensitivity and consistent order for hashing
```
dotansimha, line 18: I don't think case-insensitivity is an actual thing? (BTreeMap and HashMap are both case sensitive as far as I know?)

PR #339, opened 2025-08-14
feat(gw): error propagation

ardatan:
Rewrite of https://github.com/graphql-hive/gateway-rs/pull/299 within the new structure

- Normalize error paths with `_entities` into the real paths using flatten paths
- Bring back some of old test components so I could write the unit tests based on local schemas without actual HTTP requests
- Add `serviceName` and `code` to the subgraph errors while passing it to the final gw response
- Remove `traverse_and_callback_mut` as it is no longer needed because we collect indexes in paths and use it with flatten path to get the mutable reference in the final response without traversing again

### DIFF
--- a/lib/executor/src/execution/plan.rs
+++ b/lib/executor/src/execution/plan.rs
@@ -378,6 +378,7 @@ impl<'exec> QueryPlanExecutor<'exec> {
                                 }
                             }
                         }
+                        let normalized_path = job.flatten_node_path.as_slice();
                         'entity_loop: for (entity, hash) in entities
                             .into_iter()
                             .zip(job.representation_hashes.iter_mut())
@@ -387,7 +388,7 @@ impl<'exec> QueryPlanExecutor<'exec> {
                             {
                                 for indexes_in_path in target_paths {
                                     let mut target: &mut Value<'exec> = &mut ctx.final_response;
-                                    for path_segment in job.flatten_node_path.as_slice().iter() {
+                                    for path_segment in normalized_path {
                                         match path_segment {
                                             FlattenNodePathSegment::List => {
                                                 let index = indexes_in_path.pop_front().unwrap();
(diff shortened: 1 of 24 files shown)


lib/executor/src/execution/plan.rs
```diff
@@ -377,30 +378,109 @@ impl<'exec> Executor<'exec> {
...
-                                            unsafe { std::mem::transmute(entity.clone()) };
-                                        deep_merge(target, new_val);
+                        'entity_loop: for (entity, hash) in entities
+                            .into_iter()
+                            .zip(job.representation_hashes.iter_mut())
+                        {
+                            if let Some(target_paths) =
+                                job.filtered_representations_hashes.get_mut(hash)
+                            {
+                                for indexes_in_path in target_paths {
```
kamilkisiela, line 389: In `traverse_and_callback_mut` we give it a path and once it reaches the final segment, it executes the callback.
The traverse of `final_response` starts at the root, then it moves to the first path segment until it hits the final segment.

In your code, we always start from root, for every element in the `target_paths` vector.
At least from what I see, but maybe I'm missing something.

What was wrong with `traverse_and_callback_mut`?

lib/executor/src/execution/plan.rs
```diff
@@ -76,47 +79,25 @@ pub async fn execute_query_plan<'exec>(
...
 }
 
 struct FlattenFetchJob {
+    subgraph_name: String,
     flatten_node_path: FlattenNodePath,
     response: Bytes,
     fetch_node_id: i64,
     representation_hashes: Vec<u64>,
-    representation_hash_to_index: HashMap<u64, usize>,
+    filtered_representations_hashes: HashMap<u64, Vec<VecDeque<usize>>>,
```
kamilkisiela, line 100: Can you add a comment explaining the HashMap value? Or create a dedicated struct if it will make it more obvious?

lib/executor/src/execution/plan.rs
```diff
@@ -377,30 +378,109 @@ impl<'exec> Executor<'exec> {
...
-                            &mut ctx.final_response,
-                            normalized_path,
-                            self.schema_metadata,
-                            &mut |target| {
-                                let hash = job.representation_hashes[index];
-                                if let Some(entity_index) =
-                                    job.representation_hash_to_index.get(&hash)
-                                {
-                                    if let Some(entity) = entities.get(*entity_index) {
-                                        // SAFETY: `new_val` is a clone of an entity that lives for `'a`.
```
kamilkisiela, line 393: This comment is missing, it would be nice to include it, so we know why we use transmute.

lib/executor/src/response/error_normalization.rs
```diff
@@ -0,0 +1,186 @@
...
+    subgraph_name: &str,
+    normalized_path: &[FlattenNodePathSegment],
+    representation_hashes: &[u64],
+    hashes_to_indexes: &HashMap<u64, Vec<VecDeque<usize>>>,
+) -> Vec<GraphQLError> {
+    let mut new_errors: Vec<GraphQLError> = Vec::new();
+    if let Some(path_in_error) = &error.path {
+        if let Some(GraphQLErrorPathSegment::String(first_path)) = path_in_error.first() {
+            if first_path == "_entities" {
+                if let Some(GraphQLErrorPathSegment::Index(entity_index)) = path_in_error.get(1) {
```
kamilkisiela, line 25: Can we somehow reduce the depth level here?

lib/executor/src/execution/plan.rs
ardatan: I actually didn't touch the logic. But instead I hold the array indices (the actual values of `@` in the flatten path) so I can use them for fixing the error paths in the final response. Since we already know the path, traverse_and_callback_mut is no longer needed because we already use this information to access the entity in the fina lresponse. 
Nothing is wrong with it but `mutable` version of `traverse_and_callback` is no longer needed. Immutable version is still used to collect the entities and also indices in the paths.

lib/executor/src/execution/plan.rs
ardatan: They are the indexes of the representations. Indexes are the actual values of `@` in the flatten path.

lib/executor/src/response/error_normalization.rs
ardatan: I'll try 👍

lib/executor/src/execution/plan.rs
kamilkisiela: What about the loop within a loop. We start from the `final_response` for every `for indexes_in_path in target_paths`

PR #400, opened 2025-09-03
feat(router): support `hive` as source for supergraph, `https` connector

dotansimha:


bin/router/src/shared_state.rs
```diff
@@ -2,72 +2,29 @@ use std::sync::Arc;
...
-        )?;
-
-        Ok(Arc::new(Self {
-            schema_metadata,
-            planner,
+    pub fn new(router_config: HiveRouterConfig) -> Self {
+        Self {
             validation_plan: graphql_tools::validation::rules::default_rules_validation_plan(),
-            subgraph_executor_map,
             plan_cache: moka::future::Cache::new(1000),
```
ardatan: All those caches are supergraph specific so I think `RouterSharedState` is already the state which lasts with the supergraph.

dotansimha: Yeah, I plan to have a `shared_state.next(new_state)` that will take care of the caches as well 🫡

PR #413, opened 2025-09-10
fix(executor): if `data` is empty, add `null` to the serialized response instead of leaving it empty

ardatan:
In case of a `data` property with a non-object value, it was skipping the serialization of `data` which ends up with an invalid JSON result.
```diff
{
-   "data": ,
+  "data": null,
"errors": []
}
```

### DIFF
--- a/lib/executor/src/projection/response.rs
+++ b/lib/executor/src/projection/response.rs
@@ -53,7 +53,6 @@ pub fn project_by_operation(
             buffer.put(EMPTY_OBJECT);
         }
     } else {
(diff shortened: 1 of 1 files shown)


lib/executor/src/projection/response.rs
```diff
@@ -52,6 +52,9 @@ pub fn project_by_operation(
             // If no selections were made, we should return an empty object
             buffer.put(EMPTY_OBJECT);
         }
+    } else {
+        warn!("The root data is not an object. Returning null for data field.");
```
ardatan:
```suggestion
```

PR #414, opened 2025-09-10
feat(router): add a flag to disable the caching for query plans

ardatan:
The following disables the caching of the query plans, the caching is enabled by default.

```yaml
query_planner:
   cache:
      enabled: true | false
```

### DIFF
--- a/bin/router/src/pipeline/query_plan.rs
+++ b/bin/router/src/pipeline/query_plan.rs
@@ -9,6 +9,7 @@ use hive_router_query_planner::planner::plan_nodes::QueryPlan;
 use hive_router_query_planner::planner::PlannerError;
 use hive_router_query_planner::utils::cancellation::CancellationToken;
 use ntex::web::HttpRequest;
+use tracing::debug;
 use xxhash_rust::xxh3::Xxh3;
 
 #[inline]
@@ -74,6 +75,7 @@ fn get_plan(
     let is_pure_introspection = filtered_operation_for_plan.selection_set.is_empty()
         && normalized_operation.operation_for_introspection.is_some();
     if is_pure_introspection {
+        debug!("No need for a plan, as the incoming query only involves introspection fields");
         return Ok(QueryPlan {
             kind: "QueryPlan".to_string(),
             node: None,
--- a/bin/router/src/pipeline/query_plan.rs
+++ b/bin/router/src/pipeline/query_plan.rs
@@ -6,6 +6,7 @@ use crate::pipeline::normalize::GraphQLNormalizationPayload;
 use crate::pipeline::progressive_override::{RequestOverrideContext, StableOverrideContext};
 use crate::shared_state::RouterSharedState;
 use hive_router_query_planner::planner::plan_nodes::QueryPlan;
+use hive_router_query_planner::planner::PlannerError;
 use hive_router_query_planner::utils::cancellation::CancellationToken;
 use ntex::web::HttpRequest;
 use xxhash_rust::xxh3::Xxh3;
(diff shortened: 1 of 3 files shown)


lib/router-config/src/query_planner.rs
```diff
@@ -7,4 +7,6 @@ pub struct QueryPlannerConfig {
     /// When set to `true` and an incoming request has a `hive-expose-query-plan: true` header, the query plan will be exposed in the response, as part of `extensions`.
     #[serde(default)]
     pub allow_expose: bool,
+    #[serde(default)]
+    pub no_plan_cache: bool,
```
dotansimha: I think `no_plan_cache: true` feels like you flip the condition too many times? 

why not just `cache: false` (and then the default is `true`)?

dotansimha (changes requested):

lib/router-config/src/query_planner.rs
kamilkisiela: Let's not do `cache: boolean`, but be explicit.
What if we add `redis` cache of plans? What if we have an in-memory cache and redis at the same time and we want to disable one of them?

Imo, we should either has an option to disable caching entirely with a config option like `{ cache: { enabled: boolean } }` or `{ cache: { in_memory: { enabled: boolean } } }`.

kamilkisiela: This way we can have options for the cache that applies to in-memory, redis, disk-based whatever and also space for configs specific to each option.

bin/router/src/pipeline/query_plan.rs
```diff
@@ -21,9 +21,29 @@ pub async fn plan_operation_with_cache(
         StableOverrideContext::new(&app_state.planner.supergraph, request_override_context);
 
     let filtered_operation_for_plan = &normalized_operation.operation_for_plan;
+
+    if app_state.router_config.query_planner.no_plan_cache {
+        if filtered_operation_for_plan.selection_set.is_empty()
```
kamilkisiela: Let's share the logic of producing the plan whether or not the cache is enabled.

lib/router-config/src/query_planner.rs
ardatan: So are we going to have seperate cache storage options for each feature instead of having one single cache storage options for all features?
I think one cache storage can be shared by multiple features. So caching can be a boolean flag for query planning.

```yml
cache:
  redis: ...
query_plan:
   cache: false
respobse_cache:
  ...
```

vs
```yml
query_plan:
  cache:
    redis: ...
respobse_cache:
  cache:
    redis: ...
```

kamilkisiela:
```yaml
query_plan:
  cache:
    redis: ...
respobse_cache:
  cache:
    redis: ...
```

Does not mean you need to pass connection params twice, it could be done on higher level:

```
redis: ...
query_plan:
  cache:
    redis: 
      size: 1200
      key_prefix: ...
respobse_cache:
  cache:
    redis: 
      size: 300
      key_prefix: ...
```

You see my point?

ardatan: I've never seen this type of nested storage-specific settings before, but I don't have a strong opinion against. So I'll move it under `cache` for now, then we can tackle the storage-specific options later on.

bin/router/src/pipeline/query_plan.rs
ardatan: Do you think it looks better now?

lib/router-config/src/query_planner.rs
ardatan: It is `cache.enabled` now. Do you think it looks good now?

PR #422, opened 2025-09-15
refactor(router, executor): refactor request fingerprinting

kamilkisiela:
- Refactors the request fingerprinting logic to use a function rather than a struct with `Hash` and `PartialEq` traits. This simplifies the code and improves performance. It also adds header normalization in config to handle case-insensitivity.
- Passes upstream headers to subgraph executors
- Log to stdout in dev cli

### DIFF
--- a/lib/executor/src/executors/dedupe.rs
+++ b/lib/executor/src/executors/dedupe.rs
@@ -1,6 +1,7 @@
 use ahash::AHasher;
 use bytes::Bytes;
 use http::{HeaderMap, Method, StatusCode, Uri};
+use ntex_http::HeaderMap as NtexHeaderMap;
 use std::collections::BTreeMap;
 use std::hash::{BuildHasher, BuildHasherDefault, Hash, Hasher};
 
@@ -15,6 +16,7 @@ pub fn request_fingerprint(
     method: &Method,
     url: &Uri,
     req_headers: &HeaderMap,
+    upstream_headers: &NtexHeaderMap,
     body_bytes: &[u8],
     fingerprint_headers: &[String],
 ) -> u64 {
@@ -25,7 +27,13 @@ pub fn request_fingerprint(
     let mut headers = BTreeMap::new();
     if fingerprint_headers.is_empty() {
         // fingerprint all headers
+
         for (key, value) in req_headers.iter() {
+            if let Ok(value_str) = value.to_str() {
+                headers.insert(key.as_str(), value_str);
+            }
+        }
+        for (key, value) in upstream_headers.iter() {
             if let Ok(value_str) = value.to_str() {
                 headers.insert(key.as_str(), value_str);
             }
@@ -33,6 +41,10 @@ pub fn request_fingerprint(
     } else {
         for header_name in fingerprint_headers.iter() {
             if let Some(value) = req_headers.get(header_name) {
+                if let Ok(value_str) = value.to_str() {
+                    headers.insert(header_name, value_str);
+                }
+            } else if let Some(value) = upstream_headers.get(header_name) {
                 if let Ok(value_str) = value.to_str() {
                     headers.insert(header_name, value_str);
                 }
--- a/lib/executor/src/executors/dedupe.rs
+++ b/lib/executor/src/executors/dedupe.rs
@@ -21,7 +21,7 @@ pub fn request_fingerprint(
     let build_hasher = ABuildHasher::default();
     let mut hasher = build_hasher.build_hasher();
 
-    /// BTreeMap to ensure case-insensitivity and consistent order for hashing
+    // BTreeMap to ensure case-insensitivity and consistent order for hashing
     let mut headers = BTreeMap::new();
     if fingerprint_headers.is_empty() {
         // fingerprint all headers
--- a/lib/executor/src/executors/dedupe.rs
+++ b/lib/executor/src/executors/dedupe.rs
@@ -2,7 +2,7 @@ use ahash::AHasher;
 use bytes::Bytes;
 use http::{HeaderMap, Method, StatusCode, Uri};
 use std::collections::BTreeMap;
-use std::hash::{BuildHasherDefault, Hash, Hasher};
+use std::hash::{BuildHasher, BuildHasherDefault, Hash, Hasher};
 
 #[derive(Debug, Clone)]
 pub struct SharedResponse {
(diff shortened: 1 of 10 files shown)


lib/executor/src/executors/dedupe.rs
```diff
@@ -11,66 +12,52 @@ pub struct SharedResponse {
...
-            // fingerprint all headers
-            for (key, value) in req_headers.iter() {
+        for (key, value) in req_headers.iter() {
+            if let Ok(value_str) = value.to_str() {
+                headers.insert(key.as_str(), value_str);
+            }
+        }
+        for (key, value) in upstream_headers.iter() {
+            if let Ok(value_str) = value.to_str() {
+                headers.insert(key.as_str(), value_str);
```
ardatan, line 38: Why do we respect the gateway request headers in the fingerprint? 
If a subgraph request doesn't use anything from the gateway request, then they can be deduplicated safely. So I think we should only respect subgraph request's headers. A subgraph request that doesn't use anything from GW request can be shared with other GW request that doesn't have a subgraph request uses the GW request headers no?

GW Request from A client -> Subgraph request w/ nothing propagated from A but the body etc are the same with the one below
GW Request from B client -> Subgraph request w/ nothing propagated from B but the body etc are the same with one above

And those subgraph requests can be deduplicated. But if we add GW request headers to the fingerprint then it won't be possible to dedupe these two.

lib/executor/src/executors/dedupe.rs
```diff
@@ -11,66 +12,52 @@ pub struct SharedResponse {
...
             }
         }
-
-        Self {
-            method: method.clone(),
-            url: url.clone(),
-            headers,
-            body: body_bytes.to_vec(),
-        }
     }
```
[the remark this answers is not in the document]
kamilkisiela: @gemini-code-assist `deserialize_and_normalize_dedupe_fingerprint_headers` and `normalize_dedupe_fingerprint_headers` are responsible for reading the header names and normalizing them to lower case

lib/executor/src/executors/dedupe.rs
kamilkisiela: What if the gateway receives a request with `Authorization: Bearer xyz`, but in the request to the subgraph, you chose to send this information in some other form, like through the extensions or something else?

ardatan: If it is in the extensions then it changes the request body which is also included in the fingerprint, no?
So the fingerprint should only respect what is sent to the subgraph. If it has anything specific for that GW request, then it will already change the subgraph request headers, body or even the endpoint, so not sure if having GW request headers or anything GW request specific in the fingerprint is needed.

kamilkisiela: sure, but what when you want to deduplicate based on a custom header that is not sent to the subgraph? Like a feature flag or whatever. Subgraph calls will be the same, but for some reason you want to execute those anyway.

If I follow your example, then why we even need to selectively pick headers, and not just hash the whole thing?

Deduplicating purely on a subgraph call and rely only on that, means users have no control over what is deduplicated and what is not based on the upstream call.

ardatan: Then we can have a configuration option to modify the way of creating the fingerprint instead of adding all of the headers from the GW request to the fingerprint. By default, it can use subgraph request details then we can make the fingerprint elements configurable by the user?

kamilkisiela: An option to specifically pass the gw headers? Come on :) Passing them by default has 0 cost.
How can you make the elements configurable? For extensions for example. We would introduce some weird syntax for no reason.

ardatan: If the fingerprint is the hash of = request body + headers, then why we need a weird syntax for extensions? Extensions are part of the body. If the subgraph request has anything from GW request headers, then it will change the body or the headers, so the fingerprint will be different. If the user needs to send that query without deduplication then it can be disabled by them or fingerprint components can be configured but not sure if that's needed for 90% of cases. The subgraph request with a "query" operation will be safely deduplicated if the request body(query, variables, extensions which is the JSON body sent) + headers, then we are good in most cases.
I'm not talking about the cost of adding GW request headers to the fingerprint but the efficiency of the deduplication.

kamilkisiela: explain to me the efficiency of the fingerprint

ardatan: The right word is ` the efficiency of the deduplication` actually.
Let's say I have two requests sent to the GW with different auth headers, and they both have a subgraph request which are identical per `body` (`query`, `variables` and `extensions`) and `headers`(content-type etc but no auth header.
Then adding GW request headers to the fingerprint will not deduplicate these identical inflight subgraph requests because you have GW request headers in the fingerprint. But if we only respect what's sent(subgraph request), then these will be deduped which is more efficient or I am fully missing the point.

lib/router-config/src/traffic_shaping.rs
```diff
@@ -51,3 +54,17 @@ fn default_dedupe_enabled() -> bool {
 fn default_dedupe_fingerprint_headers() -> Vec<String> {
     vec!["authorization".to_string()]
```
dotansimha, line 55: Should this also include `Cookie`? It's fairly common (and more secure) to use it instead of `Authorization` header

lib/executor/src/executors/dedupe.rs
```diff
@@ -11,66 +12,52 @@ pub struct SharedResponse {
...
+            }
+        }
+        for (key, value) in upstream_headers.iter() {
+            if let Ok(value_str) = value.to_str() {
+                headers.insert(key.as_str(), value_str);
+            }
+        }
+    } else {
+        for header_name in fingerprint_headers.iter() {
+            if let Some(value) = req_headers.get(header_name) {
```
dotansimha, line 43: If I understood this one correctly, then it means that it will filter the downstream (to-subgraph) as well, and we talked about now allowing the user to change these? 👀

lib/executor/src/executors/dedupe.rs
```diff
@@ -11,66 +12,52 @@ pub struct SharedResponse {
...
+    upstream_headers: &NtexHeaderMap,
+    body_bytes: &[u8],
+    fingerprint_headers: &[String],
+) -> u64 {
+    let build_hasher = ABuildHasher::default();
+    let mut hasher = build_hasher.build_hasher();
+
+    // BTreeMap to ensure case-insensitivity and consistent order for hashing
+    let mut headers = BTreeMap::new();
+    if fingerprint_headers.is_empty() {
```
dotansimha, line 28: I'm pondering on what should be the default case when the list is set to empty value. should it be  all headers? or no headers at all?

lib/router-config/src/traffic_shaping.rs
```diff
@@ -21,7 +21,10 @@ pub struct TrafficShapingExecutorConfig {
     /// A list of headers that should be used to fingerprint requests for deduplication.
     ///
     /// If not provided, the default is to use the "authorization" header only.
-    #[serde(default = "default_dedupe_fingerprint_headers")]
+    #[serde(
+        default = "default_dedupe_fingerprint_headers",
+        deserialize_with = "deserialize_and_normalize_dedupe_fingerprint_headers"
+    )]
     pub dedupe_fingerprint_headers: Vec<String>,
```
dotansimha, line 28: nit: this could already be `HeaderName` instead of `String`, and then you can deserialize it (and validate that the string is an actual valid header name) while parsing

PR #427, opened 2025-09-16
Add cancellation and deduplication to the query planner

kamilkisiela:
Few things to unpack here:
- deduplication
- cancellation
- timeout

**Deduplication**
When Router receives one or more queries that are identical and have no plan yet, it shares the planning phase for them and produces a single plan that is distributed to all of them. We save CPU time.

**Cancellation**
Sometimes the query planning may take a long time. That's why I added an option to cancel it at any time.
We pass a cancellation token (like AbortSignal in JS) all over the places and check from time to time if bail or continue planning.

**Timeout**
Cancellation feature + deadline. I added a config option to the planner, to timeout (and cancel) a long running planning phase. It's 10s by default, but it can be adjusted by the user.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -93,7 +93,7 @@ Query planning configuration.
 |Name|Type|Description|Required|
 |----|----|-----------|--------|
 |**allow\_expose**|`boolean`|A flag to allow exposing the query plan in the response.<br/>When set to `true` and an incoming request has a `hive-expose-query-plan: true` header, the query plan will be exposed in the response, as part of `extensions`.<br/>Default: `false`<br/>||
-|**timeout**|`string`|The maximum time in milliseconds for the query planner to create an execution plan.<br/>This acts as a safeguard against overly complex or malicious queries that could degrade server performance.<br/>When the timeout is reached, the planning process is cancelled.<br/><br/>Default: 10s.<br/>Default: `"10s"`<br/>||
+|**timeout**|`string`|The maximum time for the query planner to create an execution plan.<br/>This acts as a safeguard against overly complex or malicious queries that could degrade server performance.<br/>When the timeout is reached, the planning process is cancelled.<br/><br/>Default: 10s.<br/>Default: `"10s"`<br/>||
 
 **Example**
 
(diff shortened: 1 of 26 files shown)


lib/query-planner/src/planner/mod.rs
```diff
@@ -66,6 +71,15 @@ impl From<QueryPlanError> for PlannerError {
     }
 }
 
+impl From<CancellationError> for PlannerError {
```
dotansimha, line 74: you can avoid this by using `#[from]` of `thiserror`

lib/router-config/src/query_planner.rs
```diff
@@ -1,10 +1,34 @@
...
-    #[serde(default)]
+    #[serde(default = "default_query_planning_allow_expose")]
     pub allow_expose: bool,
+    /// The maximum time in milliseconds for the query planner to create an execution plan.
+    /// This acts as a safeguard against overly complex or malicious queries that could degrade server performance.
+    /// When the timeout is reached, the planning process is cancelled.
+    ///
+    /// Default: 10000 (10 seconds).
+    #[serde(default = "default_query_planning_timeout_ms")]
+    pub timeout_ms: u64,
```
dotansimha: Instead of `u64`, you can use `Duration` directly here. It will allow users to easily use `1000ms` or `10s`

dotansimha: and then the field name can be `timeout`

kamilkisiela: Screenshot 2025-09-18 at 14 33 18

isn't that weird though?

kamilkisiela:
```
$ npm run generate
```

docs/README.md
```diff
@@ -92,11 +93,37 @@ Query planning configuration.
...
 ```yaml
 allow_expose: false
+timeout: 10s
+
+```
+
+<a name="query_plannertimeout"></a>
+### query\_planner\.timeout: object
+
+The maximum time in milliseconds for the query planner to create an execution plan.
```
dotansimha: `The max time duration ...` - no longer milliseconds

PR #455, opened 2025-09-27
feat(router): jwt auth

dotansimha:
- [x] figure out background tasks 
- [x] figure out per-request context 
- [x] jwks fetching 
- [x] prefetching jwks 
- [x] decode jwt
- [x] jwt validation
- [x] error handling
- [x] context injection  
- [ ] e2e testing

### DIFF
--- a/bin/router/src/background_tasks/mod.rs
+++ b/bin/router/src/background_tasks/mod.rs
@@ -1,16 +1,15 @@
 use async_trait::async_trait;
+use futures::future::join_all;
+use std::sync::Arc;
+use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
+use tracing::{debug, info};
 
 #[async_trait]
 pub trait BackgroundTask: Send + Sync {
     fn id(&self) -> &str;
     async fn run(&self, token: CancellationToken);
 }
-
-use futures::future::join_all;
-use std::sync::Arc;
-use tokio::task::JoinHandle;
-use tracing::info;
 
 pub struct BackgroundTasksManager {
     cancellation_token: CancellationToken,
@@ -29,7 +28,7 @@ impl BackgroundTasksManager {
     where
         T: BackgroundTask + 'static,
     {
-        info!("Registering background task: {}", task.id());
+        info!("registering background task: {}", task.id());
         let child_token = self.cancellation_token.clone();
         let task_arc = Arc::new(task);
 
@@ -44,7 +43,7 @@ impl BackgroundTasksManager {
         info!("shutdown triggered, stopping all background tasks...");
         self.cancellation_token.cancel();
 
-        info!("waiting for background tasks to finish...");
+        debug!("waiting for background tasks to finish...");
         join_all(self.task_handles).await;
 
         println!("all background tasks have been shut down gracefully.");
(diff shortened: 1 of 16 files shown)


bin/router/src/background_tasks/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+        self.task_handles.push(handle);
+    }
+
+    pub async fn shutdown(self) {
+        info!("shutdown triggered, stopping all background tasks...");
+        self.cancellation_token.cancel();
+
+        debug!("waiting for background tasks to finish...");
+        join_all(self.task_handles).await;
```
ardatan, line 52: `join_all` waits for all futures to be resolved but I think `FuturesUnordered` can be more efficient. What do you think?

```rust
let mut jobs = FuturesUnordered::new();
while let Some(result) = jobs.next().await {
   // Called whenever a job is finished instead of waiting all of them
}

File: lib/router-config/src/jwt_auth.rs [hunk 2]
```diff
@@ -0,0 +1,123 @@
...
+    /// If not specified, the default list of all supported algorithms in [`jsonwebtoken` crate](https://crates.io/crates/jsonwebtoken) are used.
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        default = "default_allowed_algorithms"
+    )]
+    #[schemars(with = "Option<Vec<String>>")]
+    pub allowed_algorithms: Option<Vec<Algorithm>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    /// Forward the JWT token to the upstream service in the specified header.
+    pub forward_token_to_upstream_header: Option<String>,
```
ardatan: I think this flag has not been implemented yet. Maybe you hold it for Header Manipulation PR

dotansimha: yeah not implemented, i didn't want to add it before we do the Header manipulation, because probably i can leverage the same solution for forwarding

bin/router/src/background_tasks/mod.rs
dotansimha: Yeah it's more efficient, but for the case of shutdown signal, I really don't care if it will be ordered or not (or the impact on perf, since we already shutting the gw down...)

ardatan (approved):

bin/router/src/jwt/mod.rs
```diff
@@ -268,12 +267,34 @@ impl JwtAuthRuntime {
...
+                            .as_ref()
+                            .clone(),
                         token,
-                        payload: token_data,
-                    });
+                    ));
                 }
+
+                if self.config.forward_claims_to_upstream_header.enabled {
+                    jwt_ctx.payload = Some((
```
ardatan: Does it really make sense to forward the payload with headers? Couldn't it be better to forward it through extensions as in Hive Gateway?
https://the-guild.dev/graphql/hive/docs/gateway/authorization-authentication#in-upstream-graphql-subgraphs

dotansimha: I wasn't sure what would be the best here. Other HTTP servers like Envoy pass it as headers (`forward_payload_header`). But there it's also base64.  https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/jwt_authn/v3/config.proto 
I'm not sure if it should be in `extensions` as JSON object, or as headers (raw? base64?)

ardatan: `extensions` is GraphQL-native way to do this independent from the transport and payload format. Headers can be bloated with the payload if it is big.

dotansimha: done!

PR #464, opened 2025-10-01
fix: ensure variables passed to subgraph requests

rawkode:
- Forward collected operation variables to each subgraph fetch.
- Previously execute_fetch_node always built HttpExecutionRequest with variables: None, so any resolver that expected arguments saw them missing, threw Variable \…` not provided, and the federated response ended up as invalid JSON ({"data":,"errors":[…]}`).
- Build the borrowed HashMap<&str, &sonic_rs::Value> from the query plan’s variable_values before invoking the executor so subgraphs receive the inputs they need.


lib/executor/src/execution/plan.rs
```diff
@@ -522,6 +522,12 @@ impl<'exec> Executor<'exec> {
         node: &FetchNode,
         representations: Option<Vec<u8>>,
     ) -> Result<ExecutionJob, PlanExecutionError> {
+        let variable_refs = self.variable_values.as_ref().map(|vars| {
```
ardatan: Isn't it better to pass the used variables only?
```rust
let variables = if let Some(variable_values) = self.variable_values {
            node.variable_usages.as_ref().map(|variable_usages| {
                variable_usages
                    .iter()
                    .filter_map(|var_name| {
                        variable_values
                            .get(var_name.as_str())
                            .map(|v| (var_name.as_str(), v))
                    })
                    .collect()
            })
        } else {
            None
        };
```

rawkode: Yeah, silly oversight. Let me clean this up. Thanks for the suggestion

lib/executor/src/execution/plan.rs
```diff
@@ -555,3 +558,88 @@ fn condition_node_by_variables<'a>(
...
+
+        let selected = select_fetch_variables(&variable_values, Some(&usages)).unwrap();
+
+        assert_eq!(selected.len(), 1);
+        assert!(selected.contains_key("present"));
+        assert!(!selected.contains_key("missing"));
+    }
+
+    #[test]
+    fn select_fetch_variables_returns_all_when_usages_absent() {
```
ardatan: Why do we take all variables when `variable_usages` is missing?

rawkode: Sorry, I asked codex to write the tests for me and I didn't review it closely enough. I'm fixing this now.

ardatan:
Thanks for the PR! 🙏 
The changes look good to me! After others approve, we can merge and release!

dotansimha (approved):

rawkode:
Thanks, y'all 👍🏻

PR #471, opened 2025-10-06
Override Subgraph URLs

ardatan:


ardatan:
@gemini-code-assist review and summary

docs/README.md
```diff
@@ -39,6 +40,21 @@ log:
...
   format: json
   level: info
+override_subgraph_urls:
+  subgraphs:
+    accounts:
+      url: https://accounts.example.com/graphql
+    products:
+      expression: |2-
+
+                if .request.headers."x-region" == "us-east" {
```
starJammer, line 68: Is this a case sensitive matching that is occurring? Or is it case-insensitive?
Are there any other request attributes that can be used here? Looks like the entire request object is available for use.

starJammer: Additionally, does the statement have to return a value here? 
Or will the default automatically be reused if no value is returned in the expression?

For example, suppose my default url is `http://hello.us.com`.

Is this a valid configuration for the `products` subgraph?:

```
override_subgraph_urls:
  subgraphs:
    products:
      expression: |2-
                if .request.headers."x-region" == "eu-east" {
                    "https://hello.eu.eu/graphql"
                }
```

I want to point out TWO things here:

1. Notice that the default is `http` (plain http) and the EU version of the URL is `https` (secure). Is this sort of change to the scheme supported? This was not originally supported in the Apollo router and but I submitted a PR for it to be supported here https://github.com/apollographql/router/pull/6906 

Since Apollo didn't support it by default, I don't want to assume that it is supported here.


2. Notice that I didn't include an `else` statement in the expression in the configuration above. I would expect that the default URL would be used in this case. Is this supported, or does the expression have to return a value, even the default value, or will the default value be used if no value is returned?

starJammer: My Rust isn't good at all so I can't quite look at the code myself to answer the above questions.

ardatan: > Notice that I didn't include an else statement in the expression in the configuration above. I would expect that the default URL would be used in this case. Is this supported, or does the expression have to return a value, even the default value, or will the default value be used if no value is returned?

Yes the URL in the supergraph will be used if the expression doesn't return anything.

> Notice that the default is http (plain http) and the EU version of the URL is https (secure). Is this sort of change to the scheme supported? This was not originally supported in the Apollo router and but I submitted a PR for it to be supported here 

Scheme doesn't matter in our implementation currently. It is possible to switch in between http and https.

ardatan: > Is this a case sensitive matching that is occurring? Or is it case-insensitive?

Header names are always lower-cased actually. So they are case-insensitive.

> Are there any other request attributes that can be used here? Looks like the entire request object is available for use.

It has details like URL(search params etc), headers, method and the original router operation. After the implementation of JWT, it will have the jwt payload there too. Once we have a proper documentation then we'll explain the available variables in the VRL Context!

https://github.com/graphql-hive/router/blob/3b7a28ce5342666b990bbc4db0879cecb8a53dd2/lib/executor/src/execution/plan.rs#L54

starJammer: Sounds great! This is perfect. Because my organization chose to NOT use Apollo enterprise, we're only using basic query features of the router so switching to this new router now should be super easy. 

The only thing we'll want to wait for is the ability to tie it to the Hive Registry for the supergraph schema. I saw there is an issue open for that and that it's already on the timeline.

kamilkisiela (changes requested):
I see we have a new subgraph executor, but I don't think we should.
The `executor_map` is a hashmap where keys are endpoints, and I think that's why you did it this way, but it could be changed into subgraph's name instead and then the endpoint would be dynamic.

This way we just add `resolve_endpoint` to existing executor and move on.

About `expression` returning nothing and falling back to original value. I don't like this, it could be more explicit and done with returning `.original_url` value or something like that.

Issue #321, opened 2025-08-07
RFC: CSRF Prevention


dotansimha:
Both Hive Gateway and Apollo Router follow the similar approaches on [CSRF](https://owasp.org/www-community/attacks/csrf)

The only difference is the default. By default, Hive Gateway doesn't have it, and Apollo Router has it in production mode.

The configuration is very simple. It is a list of the headers;

```
csrf:
   required-headers:
     - X-Apollo-Operation-Name
```

In case of enabling this feature, the router will check if one of the following conditions met;

- `Content-Type` other than `text/plain`, `application/x-www-form-urlencoded` or `multipart/form-data`
- One of the required headers exist

References
- https://the-guild.dev/graphql/hive/docs/gateway/other-features/security/csrf-prevention
- https://www.apollographql.com/docs/graphos/routing/security/csrf

PR #472, opened 2025-10-07
CSRF Prevention

ardatan:
Implements https://github.com/graphql-hive/router/issues/321

### DIFF
--- a/lib/router-config/src/csrf.rs
+++ b/lib/router-config/src/csrf.rs
@@ -1,0 +1,32 @@
+/// Configuration for CSRF prevention.
+/// 
+/// Cross-site request forgery (CSRF) is an attack that forces an end user to execute unwanted actions on a web application in which they're currently authenticated.
+/// By enabling CSRF prevention, the router will check for the presence of specific headers in incoming requests to the `/graphql` endpoint.
+/// If the required headers are not present, the router will reject the request with a `403 Forbidden` response.
+/// This helps to ensure that requests are coming from trusted sources and not from malicious third-party sites.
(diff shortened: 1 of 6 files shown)


lib/router-config/src/csrf.rs
```diff
@@ -0,0 +1,33 @@
...
+/// ## Case sensitivity
+/// Header names are case-insensitive, so `X-CSRF-Token` and `x-csrf-token` are treated the same.
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+#[derive(Debug, Default, Deserialize, Serialize, JsonSchema, Clone)]
+#[schemars(example = csrf_prevention_example_1())]
+pub struct CSRFPreventionConfig {
+    #[serde(default)]
+    /// A list of required header names for CSRF protection.
+    pub required_headers: Vec<String>,
```
kamilkisiela, line 27: Are you sure:
```
csrf:
  required_headers:
    - X-CsRf-ToKeN
```

will work the same way as `x-csrf-token`?

Maybe it's worth adding a normalization step to the config? Like when the config is loaded and serde deserializes it, the list of strings (headers) is normalized (lowercased).

ardatan: You're right. Let me do it.

ardatan: I added tests and it seems to work correctly.
https://github.com/graphql-hive/router/blob/csrf-prevention/bin/router/src/pipeline/csrf_prevention.rs#L125

PR #488, opened 2025-10-16
feat(router): Subgraph endpoint overrides

### LINKED ISSUE #461: Subgraph URL Rewriting Support - Multi-region graphql routing
I have a use case and DON'T know if the new Hive Router supports this yet. 
I couldn't immediately find anything in the docs.

The use case is "simple". 
I have one router and 2 instances of the same service. One instance is in the US and the other is in the EU.

I want to have the router dynamically route to either the US instance or the EU instance. In my case it would be off of either a REGION header OR a JWT claim in the JWT token.

I achieved this with the Apollo router using  Rhai plugin to overwrite the subgraph request URL based on the REGION header. 

Is this possible with the new Hive Router?

I know I could instantiate a new router in each region as an alternative solution, but currently that solution path is not possible for me.

Any comments are appreciated. Thank you.

kamilkisiela:
This adds a way to override subgraph URLs at runtime - either statically or dynamically per request using expressions.

```yaml
override_subgraph_urls:
  accounts:
    url:
      expression: |
        if .request.headers."x-region" == "eu" {
          "https://eu.example.com/accounts"
        } else {
          .original_url # represents the subgraph url defined in supergraph sdl
        }
  products:
    url: "https://example.com/products"
```

---

I also slightly changed the e2e testing setup as I was facing problems with race conditions.

---

Fixes #461
Closes #471
Ref ROUTER-149


bin/router/src/shared_state.rs
```diff
@@ -31,7 +31,7 @@ pub struct RouterSharedState {
     pub validate_cache: Cache<u64, Arc<Vec<ValidationError>>>,
     pub parse_cache: Cache<u64, Arc<graphql_parser::query::Document<'static, String>>>,
     pub normalize_cache: Cache<u64, Arc<GraphQLNormalizationPayload>>,
-    pub router_config: HiveRouterConfig,
+    pub router_config: Arc<HiveRouterConfig>,
```
dotansimha, line 34: I assume this `Arc` is needed because you want to point to the endpoint in the map, instead of cloning the URL?

lib/router-config/src/override_subgraph_urls.rs
```diff
@@ -0,0 +1,85 @@
...
+    ///     } else if .request.headers."x-region" == "eu-west" {
+    ///       "https://products-eu-west.example.com/graphql"
+    ///     } else {
+    ///       .original_url
+    ///     }
+    pub url: UrlOrExpression,
+}
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[serde(untagged)]
+pub enum UrlOrExpression {
```
dotansimha, line 51: I really like this way of configuring the different options between `String` and `{ ... }`. It makes it much simpler and easy to read and understand. Also feels more future-proof and we can add more stuff after (like subscriptions that we discussed).

lib/router-config/src/override_subgraph_urls.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// A static URL string.
+    Url(String),
+    /// A dynamic value computed by a VRL expression.
+    Expression { expression: String },
+}
+
+fn override_subgraph_urls_example_1() -> OverrideSubgraphUrlsConfig {
+    let expression = r#"
```
dotansimha, line 59: For some reason this got a `2-` prefix on the output (see README.md) - on purposes?

lib/router-config/src/override_subgraph_urls.rs
```diff
@@ -0,0 +1,85 @@
...
+        if .request.headers."x-region" == "us-east" {
+            "https://products-us-east.example.com/graphql"
+        } else if .request.headers."x-region" == "eu-west" {
+            "https://products-eu-west.example.com/graphql"
+        } else {
+          .original_url
```
dotansimha, line 65: I like this syntax and the way it handles the default/current state! Very explicit and easy to understand.

bin/router/src/shared_state.rs
kamilkisiela: HiveRouterConfig is used in two places, so we either clone or wrap it with Arc and clone

lib/router-config/src/override_subgraph_urls.rs
kamilkisiela: I don't define `2-` anywhere, so no idea why it's there

PR #494, opened 2025-10-20
fix(executor): handle subgraph errors with extensions correctly

ardatan:
Ref ROUTER-157

When subgraphs returns an error with extensions but without `data`, the deserialization was breaking due to the issue -> https://github.com/cloudwego/sonic-rs/issues/114

A manual deserialization is needed in this case

### DIFF
--- a/lib/executor/src/response/subgraph_response.rs
+++ b/lib/executor/src/response/subgraph_response.rs
@@ -83,8 +83,6 @@ impl<'de> de::Deserialize<'de> for SubgraphResponse<'de> {
 
 #[cfg(test)]
 mod tests {
-    use sonic_rs::JsonValueTrait;
-
     // When subgraph returns an error with custom extensions but without `data` field
     #[test]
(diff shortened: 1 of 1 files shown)


kamilkisiela (changes requested):
Overall looks good. Can you also remove `set()` and `get()` functions from the `GraphQLErrorExtensions` struct? They are unused.

ardatan:
@kamilkisiela The idea is to avoid nested `extensions` call
So instead of `error.extensions.extensions.get`, we can do `error.extensions.get`.
https://github.com/graphql-hive/router/pull/494/files#diff-1574fb8bdd74bf365b839d32cd853092e7ce8921740e806a30e8279b7be61691R110

ardatan:
@kamilkisiela I removed `get/set`. My idea is to make sure `extensions` deserialized correctly. 
> I feel like in tests we could just do snapshots and laverage serialization and deserialization.

`SubgraphResponse` is not serializable.

PR #499, opened 2025-10-21
feat(router): Hive Console Usage Reporting

ardatan:
Hive Console Client integration
Ref ROUTER-102
~~Blocked by https://github.com/graphql-hive/console/pull/7143~~

Documentation -> https://github.com/graphql-hive/console/pull/7171

TODOs:
- ~~Release `hive-console-sdk` and add it to Cargo.toml here~~
- ~~Update documentation and env overrides after https://github.com/graphql-hive/router/pull/519 lands.~~

### DIFF
--- a/bin/router/src/pipeline/usage_reporting.rs
+++ b/bin/router/src/pipeline/usage_reporting.rs
@@ -19,13 +19,26 @@ use crate::{
     consts::ROUTER_VERSION,
 };
 
+#[derive(Debug, thiserror::Error)]
+pub enum UsageReportingError {
+    #[error("Usage Reporting - Access token is missing. Please provide it via 'HIVE_ACCESS_TOKEN' environment variable or under 'usage_reporting.access_token' in the configuration.")]
+    MissingAccessToken,
+    #[error("Usage Reporting - Failed to initialize usage agent: {0}")]
+    AgentCreationError(#[from] AgentError),
+}
+
 pub fn init_hive_user_agent(
     bg_tasks_manager: &mut BackgroundTasksManager,
     usage_config: &UsageReportingConfig,
-) -> Result<Arc<UsageAgent>, AgentError> {
+) -> Result<Arc<UsageAgent>, UsageReportingError> {
     let user_agent = format!("hive-router/{}", ROUTER_VERSION);
+    let access_token = usage_config
+        .access_token
+        .as_deref()
+        .ok_or(UsageReportingError::MissingAccessToken)?;
+
     let hive_user_agent = UsageAgent::try_new(
-        &usage_config.access_token,
+        access_token,
         usage_config.endpoint.clone(),
         usage_config.target_id.clone(),
         usage_config.buffer_size,
--- a/lib/router-config/src/usage_reporting.rs
+++ b/lib/router-config/src/usage_reporting.rs
@@ -10,7 +10,7 @@ pub struct UsageReportingConfig {
     pub enabled: bool,
 
     /// Your [Registry Access Token](https://the-guild.dev/graphql/hive/docs/management/targets#registry-access-tokens) with write permission.
-    pub access_token: String,
+    pub access_token: Option<String>,
 
     /// A target ID, this can either be a slug following the format “$organizationSlug/$projectSlug/$targetSlug” (e.g “the-guild/graphql-hive/staging”) or an UUID (e.g. “a0f4c605-6541-4350-8cfe-b31f21a4bf80”). To be used when the token is configured with an organization access token.
     #[serde(deserialize_with = "deserialize_target_id")]
@@ -83,7 +83,7 @@ impl Default for UsageReportingConfig {
     fn default() -> Self {
         Self {
             enabled: default_enabled(),
-            access_token: String::new(),
+            access_token: None,
             target_id: None,
             endpoint: default_endpoint(),
             sample_rate: default_sample_rate(),
--- a/docs/README.md
+++ b/docs/README.md
@@ -17,7 +17,7 @@
 |[**query\_planner**](#query_planner)|`object`|Query planning configuration.<br/>Default: `{"allow_expose":false,"timeout":"10s"}`<br/>||
 |[**supergraph**](#supergraph)|`object`|Configuration for the Federation supergraph source. By default, the router will use a local file-based supergraph source (`./supergraph.graphql`).<br/>||
 |[**traffic\_shaping**](#traffic_shaping)|`object`|Configuration for the traffic-shaping of the executor. Use these configurations to control how requests are being executed to subgraphs.<br/>Default: `{"all":{"dedupe_enabled":true,"pool_idle_timeout":"50s","request_timeout":"30s"},"max_connections_per_host":100}`<br/>||
-|[**usage\_reporting**](#usage_reporting)|`object`|Configuration for usage reporting to GraphQL Hive.<br/>Default: `{"accept_invalid_certs":false,"access_token":"","buffer_size":1000,"client_name_header":"graphql-client-name","client_version_header":"graphql-client-version","connect_timeout":"5s","enabled":false,"endpoint":"https://app.graphql-hive.com/usage","exclude":[],"flush_interval":"5s","request_timeout":"15s","sample_rate":"100%","target_id":null}`<br/>|yes|
+|[**usage\_reporting**](#usage_reporting)|`object`|Configuration for usage reporting to GraphQL Hive.<br/>Default: `{"accept_invalid_certs":false,"access_token":null,"buffer_size":1000,"client_name_header":"graphql-client-name","client_version_header":"graphql-client-version","connect_timeout":"5s","enabled":false,"endpoint":"https://app.graphql-hive.com/usage","exclude":[],"flush_interval":"5s","request_timeout":"15s","sample_rate":"100%","target_id":null}`<br/>||
 
 **Additional Properties:** not allowed  
 **Example**
@@ -121,7 +121,7 @@ traffic_shaping:
   max_connections_per_host: 100
 usage_reporting:
   accept_invalid_certs: false
-  access_token: ''
+  access_token: null
   buffer_size: 1000
   client_name_header: graphql-client-name
   client_version_header: graphql-client-version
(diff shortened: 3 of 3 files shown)


bin/router/src/lib.rs
```diff
@@ -86,7 +86,30 @@ pub async fn configure_app_from_config(
...
+    let usage_agent = if let Some(usage_config) = &router_config.usage {
+        let agent = hive_console_client::agent::UsageAgent::new(
+            parsed_schema.clone(),
+            usage_config.token.clone(),
+            usage_config.endpoint.clone(),
+            usage_config.buffer_size,
+            usage_config.connect_timeout,
+            usage_config.request_timeout,
+            usage_config.accept_invalid_certs,
+        );
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist it doesn't return `Result` so your code is wrong.

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -0,0 +1,99 @@
...
+    if client_request_details
+        .operation
+        .name
+        .is_some_and(|op_name| usage_config.exclude.contains(&op_name.to_string()))
+    {
+        return;
+    }
+    let client_name = get_header_value(req, &usage_config.client_name_header);
+    let client_version = get_header_value(req, &usage_config.client_version_header);
+    let timestamp = SystemTime::now()
```
dotansimha, line 78: can be `as_millis` instead of sec*1000

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -0,0 +1,99 @@
...
+            usage_config.connect_timeout,
+            usage_config.request_timeout,
+            usage_config.accept_invalid_certs,
+            flush_interval,
+            "hive-router".to_string(),
+        )
+    })
+}
+
+pub fn send_usage_report(
```
dotansimha: I feel like the name here is not good, as the function really collects the operation. 
Maybe `collect_usage_report`?

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -0,0 +1,99 @@
...
+            .map(|op_name| op_name.to_owned()),
+        persisted_document_hash: None,
+    };
+    usage_agent
+        .add_report(execution_report)
+        .unwrap_or_else(|err| tracing::error!("Failed to send usage report: {}", err));
+}
+
+fn get_header_value(req: &HttpRequest, header_name: &str) -> Option<String> {
+    req.headers()
```
dotansimha: The value of the header doesn't have to be a `String` here. As long as you don't need, it can remain `&str` here and return as such. 
Even if eventually it wll be cloned internally by the usage-agent, i don't think it should happen here.

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -0,0 +1,99 @@
...
+use hive_console_sdk::agent::{ExecutionReport, UsageAgent};
+use hive_router_config::{usage_reporting::UsageReportingConfig, HiveRouterConfig};
+use hive_router_plan_executor::execution::plan::{ClientRequestDetails, PlanExecutionOutput};
+use ntex::web::HttpRequest;
+use rand::Rng;
+use tokio_util::sync::CancellationToken;
+
+use crate::background_tasks::BackgroundTask;
+
+pub fn from_config(router_config: &HiveRouterConfig) -> Option<UsageAgent> {
```
dotansimha: I think this should get only the config part that's relevant to it, and then turn `UsageAgent`. The condition/decision on making the agent should happen in the caller function.

dotansimha: Also, naming! probably `create_hive_usage_agrent` is a better name here.

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -0,0 +1,99 @@
...
+        hive_console_sdk::agent::UsageAgent::new(
+            usage_config.token.clone(),
+            usage_config.endpoint.clone(),
+            usage_config.target_id.clone(),
+            usage_config.buffer_size,
+            usage_config.connect_timeout,
+            usage_config.request_timeout,
+            usage_config.accept_invalid_certs,
+            flush_interval,
+            "hive-router".to_string(),
```
dotansimha: I assume this is the user-agent to use?
You can use `ROUTER_VERSION` const and append it here, so we'll have something like `hive-router@VERSION` or `hive-router/VERSION`.

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -0,0 +1,99 @@
...
+        )
+    })
+}
+
+pub fn send_usage_report(
+    schema: Arc<Document<'static, String>>,
+    start: Instant,
```
dotansimha: I feel like passing the Instant here over and over can be replaced with measuring the total time and then just pass it here as `Duration`?

bin/router/src/pipeline/usage_reporting.rs
dotansimha: Also, this function is called from the hotpath, just for the sake of micro-perf, let's `#[inline]` it.

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -0,0 +1,99 @@
...
+        operation_body: client_request_details.operation.query.to_owned(),
+        operation_name: client_request_details
+            .operation
+            .name
+            .map(|op_name| op_name.to_owned()),
+        persisted_document_hash: None,
+    };
+    usage_agent
+        .add_report(execution_report)
+        .unwrap_or_else(|err| tracing::error!("Failed to send usage report: {}", err));
```
dotansimha: nit: I feel like `unwrap_or_else` here could be more readable with a `if let` or `match` on the `Result`.

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -0,0 +1,99 @@
...
+    req.headers()
+        .get(header_name)
+        .and_then(|v| v.to_str().ok())
+        .map(|s| s.to_string())
+}
+
+#[async_trait]
+impl BackgroundTask for UsageAgent {
+    fn id(&self) -> &str {
+        "usage_report_flush_interval"
```
dotansimha: `hive_console_usage_report_task`

bin/router/src/lib.rs
```diff
@@ -110,12 +110,21 @@ pub async fn configure_app_from_config(
         true => Some(JwtAuthRuntime::init(bg_tasks_manager, &router_config.jwt).await?),
         false => None,
     };
+    let usage_agent = pipeline::usage_reporting::from_config(&router_config).map(Arc::new);
```
dotansimha: replace with the new `fn` name i suggested above. 

nit: use just the fn name, not full import here.

bin/router/src/schema_state.rs
```diff
@@ -39,6 +40,7 @@ pub struct SupergraphData {
     pub metadata: SchemaMetadata,
     pub planner: Planner,
     pub subgraph_executor_map: SubgraphExecutorMap,
+    pub schema: Arc<Document<'static, String>>,
```
dotansimha: rename to `supergraph_schema` as it's not clear if that's a supergraphg or a public-api schema

bin/router/src/shared_state.rs
```diff
@@ -18,12 +19,14 @@ pub struct RouterSharedState {
     pub override_labels_evaluator: OverrideLabelsEvaluator,
     pub cors_runtime: Option<Cors>,
     pub jwt_auth_runtime: Option<JwtAuthRuntime>,
+    pub usage_agent: Option<Arc<UsageAgent>>,
```
dotansimha: I think we should pick one path:
1- call it `usage_agent` and then we need to abstract it a bit (like we did with supergraph loading)
2-call it `hive_usage_agent` here.

I tend to go with 2 for now

bin/router/src/shared_state.rs
```diff
@@ -48,4 +52,6 @@ pub enum SharedStateError {
     CORSConfig(#[from] Box<CORSConfigError>),
     #[error("invalid override labels config: {0}")]
     OverrideLabelsCompile(#[from] Box<OverrideLabelsCompileError>),
+    #[error("error creating usage agent: {0}")]
```
dotansimha: `error creating hive usage agent:`

lib/router-config/src/usage_reporting.rs
```diff
@@ -0,0 +1,86 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct UsageReportingConfig {
+    /// Your [Registry Access Token](https://the-guild.dev/graphql/hive/docs/management/targets#registry-access-tokens) with write permission.
+    pub token: String,
```
dotansimha: I think this should be `access_token` as this is explicitly how we call it everywhere in Console.

lib/router-config/src/lib.rs
```diff
@@ -92,6 +93,10 @@ pub struct HiveRouterConfig {
     /// Configuration for overriding labels.
     #[serde(default, skip_serializing_if = "HashMap::is_empty")]
     pub override_labels: OverrideLabelsConfig,
+
+    /// Configuration for usage reporting to GraphQL Hive.
+    #[serde(default)]
+    pub usage_reporting: Option<usage_reporting::UsageReportingConfig>,
```
dotansimha: Since we have a `enabled` field in `UsageReportingConfig`, this one should not be wrapped with `Option`. 
It should be:

```
#[serde(default = "usage_reporting::UsageReportingConfig::default")]
    pub usage_reporting: usage_reporting::UsageReportingConfig,
```

And the `impl Default for UsageReportingConfig` should be implemented to configure it with `enabled: false`

lib/router-config/src/usage_reporting.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub target_id: Option<String>,
+    /// For self-hosting, you can override `/usage` endpoint (defaults to `https://app.graphql-hive.com/usage`).
+    #[serde(default = "default_endpoint")]
+    pub endpoint: String,
+    /// Sample rate to determine sampling.
+    /// 0.0 = 0% chance of being sent
+    /// 1.0 = 100% chance of being sent
+    /// Default: 1.0
+    #[serde(default = "default_sample_rate")]
+    pub sample_rate: f64,
```
dotansimha: should this be a more user-friendly value? 
I mean, if for durations we are using `humantime` and allow things like `10s`, then why not allow user to write here `10%` instead of `0.1`?

lib/router-config/src/usage_reporting.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// A timeout for only the connect phase of a request to Hive Console
+    /// Unit: seconds
+    /// Default: 5 (s)
+    #[serde(default = "default_connect_timeout")]
+    pub connect_timeout: u64,
+    /// A timeout for the entire request to Hive Console
+    /// Unit: seconds
+    /// Default: 15 (s)
+    #[serde(default = "default_request_timeout")]
+    pub request_timeout: u64,
```
dotansimha: This should be humantime, see other plugins for example.

lib/router-config/src/usage_reporting.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub buffer_size: usize,
+    /// Accepts invalid SSL certificates
+    /// Default: false
+    #[serde(default = "default_accept_invalid_certs")]
+    pub accept_invalid_certs: bool,
+    /// A timeout for only the connect phase of a request to Hive Console
+    /// Unit: seconds
+    /// Default: 5 (s)
+    #[serde(default = "default_connect_timeout")]
+    pub connect_timeout: u64,
```
dotansimha: This should be humantime, see other plugins for example.

lib/router-config/src/usage_reporting.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// A timeout for the entire request to Hive Console
+    /// Unit: seconds
+    /// Default: 15 (s)
+    #[serde(default = "default_request_timeout")]
+    pub request_timeout: u64,
+    /// Frequency of flushing the buffer to the server
+    /// Default: 5 seconds
+    #[serde(default = "default_flush_interval")]
+    pub flush_interval: u64,
```
dotansimha: This should be humantime, see other plugins for example.

lib/router-config/src/usage_reporting.rs
```diff
@@ -0,0 +1,86 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct UsageReportingConfig {
+    /// Your [Registry Access Token](https://the-guild.dev/graphql/hive/docs/management/targets#registry-access-tokens) with write permission.
+    pub token: String,
+    /// A target ID, this can either be a slug following the format “$organizationSlug/$projectSlug/$targetSlug” (e.g “the-guild/graphql-hive/staging”) or an UUID (e.g. “a0f4c605-6541-4350-8cfe-b31f21a4bf80”). To be used when the token is configured with an organization access token.
+    pub target_id: Option<String>,
```
dotansimha, line 17: nit: this one can be validated during de-serialization as either `{string}/{string}/{string}` or `uuid`.

Cargo.toml
```diff
@@ -26,8 +26,10 @@ lto = true
 codegen-units = 1
 
 [workspace.dependencies]
-graphql-tools = "0.4.0"
-graphql-parser = "0.4.1"
+graphql-parser = { version = "0.5.0", package = "graphql-parser-hive-fork" }
```
dotansimha: why use the custom one? we dropped it on purpose.

lib/router-config/src/lib.rs
ardatan: Actually we don't have an `enabled` flag in `UsageReportingConfig`.

Cargo.toml
ardatan: SDK uses this one.

dotansimha: So we need to fix that in SDK. I don't think the SDK has any reason to still use it now. 
Router shouldn't use this custom one

lib/router-config/src/lib.rs
dotansimha: Then I think we should. We need to be explicit on these, otherwise it might end up with lack of ability to enable/disale via things like env vars.

ardatan: If we introduce an `enabled` flag, user will need to define `enabled: true` even if they provide the env variables. Hive Gateway doesn't have an `enabled` flag for example. But let me try to make it enabled in case of env vars.

Cargo.toml
ardatan: https://github.com/graphql-hive/console/pull/7196

lib/router-config/src/lib.rs
ardatan: Done!

lib/router-config/src/usage_reporting.rs
ardatan: This is needed
https://github.com/graphql-hive/console/pull/7196

lib/router-config/src/usage_reporting.rs
ardatan: This is needed
https://github.com/graphql-hive/console/pull/7196

lib/router-config/src/usage_reporting.rs
ardatan: Done 👍

lib/router-config/src/usage_reporting.rs
ardatan: Done 👍

bin/router/src/shared_state.rs
ardatan: Done 👍

bin/router/src/shared_state.rs
ardatan: Done 👍

bin/router/src/schema_state.rs
ardatan: Done 👍

bin/router/src/lib.rs
ardatan: Done 👍

bin/router/src/pipeline/usage_reporting.rs
ardatan: Done 👍

bin/router/src/pipeline/usage_reporting.rs
ardatan: Done 👍

bin/router/src/pipeline/usage_reporting.rs
ardatan: Done 👍

bin/router/src/pipeline/usage_reporting.rs
ardatan: Done 👍

bin/router/src/pipeline/usage_reporting.rs
ardatan: Done 👍

bin/router/src/pipeline/usage_reporting.rs
ardatan: Done 👍

bin/router/src/pipeline/usage_reporting.rs
ardatan: Done 👍

bin/router/src/pipeline/usage_reporting.rs
ardatan: Done 👍

lib/router-config/src/usage_reporting.rs
ardatan: Done 👍

lib/router-config/src/usage_reporting.rs
ardatan: Done

Cargo.toml
dotansimha: can be reverted now

bin/router/Cargo.toml
```diff
@@ -19,6 +19,7 @@ path = "src/main.rs"
 hive-router-query-planner = { path = "../../lib/query-planner", version = "2.0.2" }
 hive-router-plan-executor = { path = "../../lib/executor", version = "6.0.0" }
 hive-router-config = { path = "../../lib/router-config", version = "0.0.10" }
+hive-console-sdk = "0.0.0"
```
dotansimha: `0.1.0`

Cargo.toml
ardatan: Done!

ardatan: Yes! Done!

lib/router-config/src/usage_reporting.rs
ardatan: Done!

lib/router-config/src/usage_reporting.rs
ardatan: Done!

bin/router/Cargo.toml
ardatan: Done!

PR #518, opened 2025-10-23
feat(router): added support for label overrides

kamilkisiela:
ROUTER-163

This pull request introduces a new top-level configuration option, `override_labels`, allowing for fine-grained control over progressive override labels (`@override(label: "<string>")`). This feature enables teams to dynamically activate or deactivate field overrides based on static booleans or runtime expressions evaluated against incoming request data.

**How to Configure Override Labels**

The new `override_labels` key can be added to your `router.config.yaml`. It accepts a map where each key is the name of an override label and the value defines the condition under which it should be considered "active".

**Example 1: Static Override**

For simple on/off toggling, you can use a boolean value.

```yaml
override_labels:
  # This label will be active for every single request, enabling the associated override.
  use_new_billing_service: true
  # This label will never be active.
  use_legacy_inventory: false
```

**Example 2: Dynamic Override with an Expression**
For more complex scenarios, you can provide an object with an `expression` key. The expression is evaluated at runtime for each request using the VRL. If the expression evaluates to `true`, the label becomes active for that request.

```yaml
override_labels:
  # Activate a label only for internal users or beta testers based on a header.
  activate-beta-feature:
    expression: "request.headers['x-user-group'] == 'beta'"
```

**Error Handling**

- **Startup**: If an expression in the configuration is invalid, the router will fail to start and log a descriptive error, preventing deployment with a broken configuration.
- **Runtime**: If an expression fails to execute during a request (e.g., it produces a non-boolean result), the router will return a `500 Internal Server Error` with the error code `LABEL_EVALUATION_FAILED`.

### DIFF
--- a/bin/router/src/pipeline/mod.rs
+++ b/bin/router/src/pipeline/mod.rs
@@ -1,7 +1,11 @@
 use std::{borrow::Cow, sync::Arc};
 
-use hive_router_plan_executor::execution::plan::PlanExecutionOutput;
-use hive_router_query_planner::utils::cancellation::CancellationToken;
+use hive_router_plan_executor::execution::plan::{
+    ClientRequestDetails, OperationDetails, PlanExecutionOutput,
+};
+use hive_router_query_planner::{
+    state::supergraph_state::OperationKind, utils::cancellation::CancellationToken,
+};
 use http::{header::CONTENT_TYPE, HeaderValue, Method};
 use ntex::{
     util::Bytes,
@@ -12,7 +16,7 @@ use crate::{
     pipeline::{
         coerce_variables::coerce_request_variables,
         csrf_prevention::perform_csrf_prevention,
-        error::PipelineError,
+        error::{PipelineError, PipelineErrorFromAcceptHeader, PipelineErrorVariant},
         execution::execute_plan,
         execution_request::get_execution_request,
         header::{
@@ -111,7 +115,6 @@ pub async fn execute_pipeline(
     validate_operation_with_cache(req, supergraph, schema_state, shared_state, &parser_payload)
         .await?;
 
-    let progressive_override_ctx = request_override_context()?;
     let normalize_payload = normalize_request_with_cache(
         req,
         supergraph,
@@ -120,12 +123,34 @@ pub async fn execute_pipeline(
         &parser_payload,
     )
     .await?;
-    let query = Cow::Owned(execution_request.query.clone());
+    let query: Cow<'_, str> = Cow::Owned(execution_request.query.clone());
     let variable_payload =
         coerce_request_variables(req, supergraph, execution_request, &normalize_payload)?;
 
     let query_plan_cancellation_token =
         CancellationToken::with_timeout(shared_state.router_config.query_planner.timeout);
+
+    let progressive_override_ctx =
+        request_override_context(&shared_state.override_labels_evaluator, || {
+            ClientRequestDetails {
+                method: req.method().clone(),
+                url: req.uri().clone(),
+                headers: req.headers(),
+                operation: OperationDetails {
+                    name: normalize_payload.operation_for_plan.name.clone(),
+                    kind: match normalize_payload.operation_for_plan.operation_kind {
+                        Some(OperationKind::Query) => "query",
+                        Some(OperationKind::Mutation) => "mutation",
+                        Some(OperationKind::Subscription) => "subscription",
+                        None => "query",
+                    },
+                    query: query.clone(),
+                },
+            }
+        })
+        .map_err(|error| {
+            req.new_pipeline_error(PipelineErrorVariant::LabelEvaluationError(error))
+        })?;
 
     let query_plan_payload = plan_operation_with_cache(
         req,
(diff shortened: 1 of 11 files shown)


bin/router/src/pipeline/progressive_override.rs
```diff
@@ -77,3 +114,105 @@ impl StableOverrideContext {
...
+                    return Err(LabelEvaluationError::ExpressionResolutionFailure {
+                        label: label.clone(),
+                        source: err,
+                    });
+                }
+            }
+        }
+
+        Ok(active_flags)
+    }
```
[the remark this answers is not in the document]
kamilkisiela: @gemini-code-assist  `static_enabled_labels` represent a list of labels that was set to `true` in the configuration file (static values), so it's one time operations and cloning is cheap. The labels that come from expressions are never colliding with those static, as you can configure a label to be either static or dynamic (expression).

bin/router/src/shared_state.rs
```diff
@@ -8,12 +8,14 @@ use std::sync::Arc;
...
 use crate::jwt::JwtAuthRuntime;
 use crate::pipeline::cors::{CORSConfigError, Cors};
+use crate::pipeline::progressive_override::{OverrideLabelsCompileError, OverrideLabelsEvaluator};
 
 pub struct RouterSharedState {
     pub validation_plan: ValidationPlan,
     pub parse_cache: Cache<u64, Arc<graphql_parser::query::Document<'static, String>>>,
     pub router_config: Arc<HiveRouterConfig>,
     pub headers_plan: HeaderRulesPlan,
+    pub override_labels_evaluator: Arc<OverrideLabelsEvaluator>,
```
dotansimha: As far as I can see from the code, this field doesn't need to be inside an `Arc` as the evaluator is not really shared by itself. The overall router state is shared (and Arc-ed) so you don't really need to `Arc` this one as well.

bin/router/src/shared_state.rs
```diff
@@ -37,7 +43,9 @@ impl RouterSharedState {
 #[derive(thiserror::Error, Debug)]
 pub enum SharedStateError {
     #[error("invalid headers config: {0}")]
-    HeaderRuleCompileError(#[from] Box<HeaderRuleCompileError>),
+    HeaderRuleCompile(#[from] Box<HeaderRuleCompileError>),
     #[error("invalid regex in CORS config: {0}")]
-    CORSConfigError(#[from] Box<CORSConfigError>),
+    CORSConfig(#[from] Box<CORSConfigError>),
+    #[error("invalid override labels config: {0}")]
+    OverrideLabelsCompile(#[from] Box<OverrideLabelsCompileError>),
```
dotansimha, line 50: nit: I wasn't sure if a `Box` is really needed here to begin with (not specifically in your changes, but in general).

lib/router-config/src/lib.rs
```diff
@@ -69,6 +70,10 @@ pub struct HiveRouterConfig {
     /// Configuration for overriding subgraph URLs.
     #[serde(default)]
     pub override_subgraph_urls: override_subgraph_urls::OverrideSubgraphUrlsConfig,
+
+    /// Configuration for overriding labels.
+    #[serde(default)]
```
dotansimha: nit: `skip_serializing_if = "HashMap::is_empty"` to avoid this from being serialized as `{}` in the api ref.

bin/router/src/pipeline/progressive_override.rs
```diff
@@ -77,3 +114,105 @@ impl StableOverrideContext {
...
+        let mut target = VrlTargetValue {
+            value: VrlValue::Object(BTreeMap::from([(
+                "request".into(),
+                (&client_request).into(),
+            )])),
+            metadata: VrlValue::Object(BTreeMap::new()),
+            secrets: VrlSecrets::default(),
+        };
+
+        let mut state = VrlState::default();
```
dotansimha, line 187: side-note/nit: we might want to have some `ExpressionRuntime` wrapper in the future to reduce the VRL setup overhead, seems like it's already repeated in multiple places.

bin/router/src/pipeline/mod.rs
```diff
@@ -116,13 +119,35 @@ pub async fn execute_pipeline(
...
+                        Some(OperationKind::Mutation) => "mutation",
+                        Some(OperationKind::Subscription) => "subscription",
+                        None => "query",
+                    },
+                    query: query.clone(),
+                },
+            }
+        })
+        .map_err(|error| {
+            req.new_pipeline_error(PipelineErrorVariant::LabelEvaluationError(error))
```
dotansimha, line 152: I wonder if a failure in evaluating the label needs to be marked as a overall request failure? 

I think in such case, the override just needs to be skipped and logged?

kamilkisiela: the goal of an override is to route to X or Y, when it fails to be resolved, you route to a place you may not want to rout. People will notice warnings after like a week or something.

bin/router/src/shared_state.rs
kamilkisiela: Without `Box` you will hit a point where there's so much to put on the stack, that clippy will start to shout "move it to the heap, move it to the heap".

Issue #508, opened 2025-10-22
Expose specific configurations as env vars


dotansimha:
* Logging
* Supergraph 

Also, remove the default env-var source of `config`

PR #519, opened 2025-10-24
feat(config): configuration override using env vars, enable/disable graphiql via config

dotansimha:
Closes https://github.com/graphql-hive/router/issues/508 

## TODO
- [x] setup nice way to define env vars and their overrides
- [x] Define logger overrides
- [x] Define supergraph overrides (file/hive)
- [x] Define HTTP overrides
- [x] Remove existing `config` extractor
- [x] Fix printing of error message in cases where server fails to start and logger is not configured yet
- [x] graphiql disable/enable
- [x] hive supergraph poll interval
- [x] Fix CI
- [x] Update docs and README 
- [x] Docs https://github.com/graphql-hive/console/pull/7172

### DIFF
--- a/lib/router-config/src/env_overrides.rs
+++ b/lib/router-config/src/env_overrides.rs
@@ -14,6 +14,10 @@ pub struct EnvVarOverrides {
     #[envconfig(from = "LOG_FILTER")]
     pub log_filter: Option<String>,
 
+    // GraphiQL overrides
+    #[envconfig(from = "GRAPHIQL_ENABLED")]
+    pub graphiql_enabled: Option<bool>,
+
     // HTTP overrides
     #[envconfig(from = "PORT")]
     pub http_port: Option<u64>,
@@ -27,6 +31,8 @@ pub struct EnvVarOverrides {
     pub hive_console_cdn_endpoint: Option<String>,
     #[envconfig(from = "HIVE_CDN_KEY")]
     pub hive_console_cdn_key: Option<String>,
+    #[envconfig(from = "HIVE_CDN_POLL_INTERVAL")]
+    pub hive_console_cdn_poll_interval: Option<String>,
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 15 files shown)


lib/router-config/src/env_overrides.rs
```diff
@@ -0,0 +1,86 @@
+use config::{builder::BuilderState, ConfigBuilder, ConfigError};
+use envconfig::Envconfig;
+use tracing::debug;
+
+use crate::log::{LogFormat, LogLevel};
+
+#[derive(Envconfig)]
+pub struct EnvVarOverrides {
```
dotansimha, line 8: @kamilkisiela @ardatan we defined all overrides here.

lib/router-config/src/env_overrides.rs
```diff
@@ -0,0 +1,86 @@
...
+    #[error("Missing required environment variable: {0}")]
+    MissingRequiredEnvVar(&'static str),
+}
+
+impl EnvVarOverrides {
+    pub fn apply_overrides<T: BuilderState>(
+        mut self,
+        mut config: ConfigBuilder<T>,
+    ) -> Result<ConfigBuilder<T>, EnvVarOverridesError> {
+        if let Some(log_level) = self.log_level.take() {
```
dotansimha, line 53: Then apply all overrides here.

lib/router-config/src/env_overrides.rs
```diff
@@ -0,0 +1,86 @@
...
+    #[envconfig(from = "HOST")]
+    pub http_host: Option<String>,
+
+    // Supergraph overrides
+    #[envconfig(from = "SUPERGRAPH_FILE_PATH")]
+    pub supergraph_file_path: Option<String>,
+    #[envconfig(from = "HIVE_CDN_ENDPOINT")]
+    pub hive_console_cdn_endpoint: Option<String>,
+    #[envconfig(from = "HIVE_CDN_KEY")]
+    pub hive_console_cdn_key: Option<String>,
```
kamilkisiela, line 33: @n1ru4l @jdolle @enisdenjo 
it's a chance to have better (or the same) env vars for Hive CDN related stuff. WDYT?

dotansimha: We'll go with `HIVE_CDN_ENDPOINT` and `HIVE_CDN_KEY`. Thanks!

kamilkisiela (approved):

PR #526, opened 2025-10-26
fix(query-planner, router): fix introspection for federation v1 supergraph

dotansimha:
The consumer-facing schema was missing filtering for `enum core__Purpose`, `directive @core`, and `directive @join_owner`. This caused introspection errors due to failure to find other associated types.


ardatan (approved):
Do you think we should add some tests to make sure the supergraph variant that has `core__x` definitions?

PR #531, opened 2025-10-28
Enhance CI benchmarks with regression detection

kamilkisiela:
This commit introduces a new script to detect performance regressions. It compares the 'http_reqs' rate metric from the PR summary against the main branch summary and determines if there is a regression of more than 2%.

Also, the commit adds .gitignore entries, a script to detect regressions, and changes to the CI workflow to execute the benchmarks and compare the results.

### DIFF
--- a/bench/ci-detect-regression.sh
+++ b/bench/ci-detect-regression.sh
@@ -4,7 +4,7 @@
 #
 # It's a script to detect performance regression between two k6 benchmark summary files.
 # It compares the 'http_reqs' rate metric from the PR summary against the main branch
-# summary and determines if there is a regression of more than 5%.
+# summary and determines if there is a regression of more than 2%.
 # If a regression is detected, the script exits with a non-zero status code.
 
 # Ensure jq and bc are installed


bench/ci-detect-regression.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+# This script is meant to be run in CI (./github/workflows/ci.yaml#router-benchmark).
+#
+# It's a script to detect performance regression between two k6 benchmark summary files.
+# It compares the 'http_reqs' rate metric from the PR summary against the main branch
+# summary and determines if there is a regression of more than 5%.
```
kamilkisiela:
```suggestion
# summary and determines if there is a regression of more than 2%.
```

bench/ci-detect-regression.sh
```diff
@@ -0,0 +1,67 @@
...
+diff=$(echo "scale=4; (($PR_RATE - $MAIN_RATE) / $MAIN_RATE) * 100" | bc)
+
+# Print the results
+echo "Main branch http_reqs rate: $MAIN_RATE"
+echo "PR branch http_reqs rate:   $PR_RATE"
+printf "Difference: %.2f%%\n" "$diff"
+
+# Check if the difference is a regression of more than 5%
+# bc returns 1 for true, 0 for false. We compare with a negative number.
+is_regression=$(echo "$diff < -5" | bc)
```
kamilkisiela:
```suggestion
is_regression=$(echo "$diff < -2" | bc)
```

bench/ci-detect-regression.sh
```diff
@@ -0,0 +1,67 @@
...
+
+if [ "$is_regression" -eq 1 ]; then
+    echo "Performance regression detected! The PR is more than 5% slower than main."
```
kamilkisiela:
```suggestion
    echo "Performance regression detected! The PR is more than 2% slower than main."
```

.github/workflows/ci.yaml
```diff
@@ -128,11 +128,46 @@ jobs:
...
         run: k6 run bench/k6.js
+      # run only on PR
+      - name: Run k6 benchmark for Pull Request
+        if: github.event_name == 'pull_request'
+        run: k6 run -e SUMMARY_PATH=./bench/results/pr bench/k6.js
+      - name: Checkout main branch in a separate directory
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+        with:
+          ref: main
```
dotansimha, line 149: we do not chain PR a lot, but this should really be `${{ github.base_ref }}`

PR #534, opened 2025-10-28
refactor(jwt): improve the implementation of `jwt` plugin and expose it to expressions

dotansimha:
Closes https://github.com/graphql-hive/router/issues/511 

## Overview 

This PR refactors the internals of `jwt` auth flow:
- [x] Fixed a bug where auth failed when additional claims were passed, due to internal deserialization failure. The struct passed to `jsonwebtoken::decode` is now based on `serde_json` and not `sonic_rs` (not supported). 
- [x] `ClientRequestDetails` now contains information about the authentication state, including the raw `claims`, extracted `scopes` (from `token_payload.scope` or `.scopes`). Both string-based (space delimited) and array-based values are valid for scopes. 
- [x] JWT Claims forwarding via GraphQL `extensions` is now using `ClientRequestDetails` and the forwarding-plan is built based on the `ClientRequestDetails` data. 
- [x] Expressions that uses `ClientRequestDetails` now have full access to `.request.jwt` with the status and the information neede for checking/accessing auth details.  
- [x] Slight refactor to where `ClientRequestDetails` is defined. Also the conversion of `ClientRequestDetails` into `vrl::Value` is more generic now and moved next to `ClientRequestDetails` (it's being used in multiple places and multiple types of extensions) 
- [x] Update `jwt` configuration structure, and added `jwt.enabled` flag (default = `false`), making it easier for the router to read though these configurations.  

Expressions can now use `request.jwt` and access:

- `request.jwt` will always be an object
- `request.jwt.authenticated` with value of `true` or `false`
- `request.jwt.prefix` can either be a `string` or `null` (if prefix is not used) 
- `request.jwt.token` can be `string` (when `authenticated=true`) or `null` (when `authenticated=false`)
- `request.jwt.claims` will always be an array (either empty or with values), containing the full JWT token claims payload. 
- `request.jwt.scopes` will always be an array (either empty or with values), containing the scopes extracted from the claims 

The whole `jwt` and the fields will always be there, in order to make interaction easier with VRL (due to possible failable expression when an array is `null` value) 

## TODO 

- [x] Add `jwt` details to `ClientRequestDetails` 
- [x] Improve JWT config by adding `enabled` field 
- [x] Refactor how JWT forwarding (via `extensions`) is implemented 
- [x] Pass `jwt` claims to expressions 
- [x] Add e2e tests 

## Examples 

```yaml
# Passes the user-id held in `.sub` claims of the token to the subgraph, or EMPTY
headers:
  all:
    request:
      - insert:
          name: X-User-ID
          expression: |
            if .request.jwt.authenticated == true {
              .request.jwt.claims.sub
            } else {
              "EMPTY"
            }
```

```yaml
# Passes a custom header based on the status of the authentication and the status of the JWT scopes
headers:
 subgraphs:
    accounts:
      request:
        - insert:
            name: X-Can-Read
            expression: |
              log(.request.jwt.scopes)

              if .request.jwt.authenticated == true && includes!(.request.jwt.scopes, "read:accounts") {
                "Yes"
              } else {
                "No"
              }

```


bin/router/src/jwt/context.rs
```diff
@@ -61,6 +73,8 @@ pub struct JwtClaims {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub jti: Option<String>,
 
+    // we are using serde to deserialize the additional claims
+    // because the jsonwebtoken crate is using `serde_json` internally, and the `sonic_rs::Value` is not recognized as valid type
     #[serde(flatten)]
-    pub additional_claims: HashMap<String, Value>,
+    pub additional_claims: BTreeMap<String, serde_json::Value>,
```
kamilkisiela: it's BTreeMap instead of HashMap because of caching / hashing / determinism?

dotansimha: Not really, changing to `HashMap`.

kamilkisiela: It's fine, just was curious if there's a reason

dotansimha: It's a leftover from an attempt I did to make it with (with `sonic_rs::Value`), but turns out it doesn't help and I didn't revert it :)

PR #537, opened 2025-10-30
Implement JWT claims caching to avoid re-parsing on every

kamilkisiela:
Adds short-lived (5s) cache of decoded and verified JWT payloads.

With this optimization we go from 21415 rps to 37638 rps.

The 5s TTL is imo a good balance between performance and any cache invalidations that may occur.
Yeah I thought about doing it configurable, but does it really have to be? People may abuse it and set it to a high number that could interfere with expiration date of the token itself. We can always add it, when requested and needed.

- Added `JwtClaimsCache` as a shared state.
- Updated `validate_request` to use JWT caching.
- Updated error types to be `Clone` to allow caching to work.

| Before | After |
|:------:|:----:|
| Screenshot 2025-10-30 at 12 45 40 | Screenshot 2025-10-30 at 12 45 47 |

### DIFF
--- a/bin/router/src/shared_state.rs
+++ b/bin/router/src/shared_state.rs
@@ -81,8 +81,9 @@ impl RouterSharedState {
             parse_cache: moka::future::Cache::new(1000),
             cors_runtime: Cors::from_config(&router_config.cors).map_err(Box::new)?,
             jwt_claims_cache: Cache::builder()
-                // Consistent with parse_cache and prevents unbounded memory usage.
-                .max_capacity(1000)
+                // High capacity due to potentially high token diversity.
+                // Capping prevents unbounded memory usage.
+                .max_capacity(10_000)
                 .expire_after(JwtClaimsExpiry)
                 .build(),
             router_config: router_config.clone(),


bin/router/src/jwt/mod.rs
```diff
@@ -265,26 +266,47 @@ impl JwtAuthRuntime {
...
+                return Ok(());
+            }
+        };
+
+        let validation_result = cache
+            .try_get_with(token.clone(), async {
+                let valid_jwks = self.jwks.all();
+                self.authenticate(&valid_jwks, request)
+                    .map(|(payload, _, _)| Arc::new(payload))
+            })
```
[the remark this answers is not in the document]
kamilkisiela: It's fine

bin/router/src/shared_state.rs
```diff
@@ -4,19 +4,68 @@ use hive_router_plan_executor::headers::{
...
 pub struct RouterSharedState {
     pub validation_plan: ValidationPlan,
     pub parse_cache: Cache<u64, Arc<graphql_parser::query::Document<'static, String>>>,
     pub router_config: Arc<HiveRouterConfig>,
     pub headers_plan: HeaderRulesPlan,
     pub override_labels_evaluator: OverrideLabelsEvaluator,
     pub cors_runtime: Option<Cors>,
+    /// Cache for validated JWT claims to avoid re-parsing on every request.
+    /// The cache key is the raw JWT token string.
+    /// Stores the parsed claims payload for 5s.
```
kamilkisiela:
```suggestion
    /// Cache for validated JWT claims to avoid re-parsing on every request.
    /// The cache key is the raw JWT token string.
    /// Stores the parsed claims payload for 5s,
    /// but no longer than `exp` date.
```

bin/router/src/shared_state.rs
```diff
@@ -30,6 +80,11 @@ impl RouterSharedState {
             headers_plan: compile_headers_plan(&router_config.headers).map_err(Box::new)?,
             parse_cache: moka::future::Cache::new(1000),
             cors_runtime: Cors::from_config(&router_config.cors).map_err(Box::new)?,
+            jwt_claims_cache: Cache::builder()
+                // Consistent with parse_cache and prevents unbounded memory usage.
+                .max_capacity(1000)
```
kamilkisiela:
```suggestion
                // High capacity due to potentially high token diversity.
                // Capping prevents unbounded memory usage.
                .max_capacity(10_000)
```

Issue #351, opened 2025-08-20
Federation directive-based authorization/authentication


ardatan:
## Directives
Since Federation v2.5, three directives are introduced `@authenticated`, `@requiresScopes` and `@policy`

- Granular regular protection with `directive @authenticated on  FIELD_DEFINITION  | OBJECT  | INTERFACE  | SCALAR  | ENUM`;
When you add this directive in your subgraph, it is directly added in the supergraph during composition. So it marks the component resolvable only if defined auth conditions met. 

- RBAC with `directive @requiresScopes(scopes: [[federation__Scope!]!]!) on FIELD_DEFINITION | OBJECT  | INTERFACE  | SCALAR  | ENUM`
This directive becomes resolvable by the user only when the set of scopes are available in the auth payload.

`AND / OR`;
```graphql
extend schema @link(url: "https://specs.apollo.dev/federation/v2.5", import: ["@requiresScopes"])
 
type Query {
  # This field requires the user to have `read:user` OR `read:admin` scopes
  me: User! @requiresScopes(scopes: [["read:user"], ["read:admin"]])
  # This field requires the user to have `read:user` AND `read:admin` scopes
  protectedField: String @requiresScopes(scopes: [["read:admin", "read:user"]])
  publicField: String
}
```

- Dynamic RBAC with `directive @policy(policies: [[federation__Policy!]!]!) on  | FIELD_DEFINITION  | OBJECT  | INTERFACE  | SCALAR  | ENUM`
Same with `@requiresScopes` but it requires an external call/logic to fetch policies which are treated in a similar way with `scopes` above.
So everytime we need to compare policies, the gateway needs to call something like below;
```ts
  fetchPolicies: async user => {
    const res = await fetch('https://policy-service.com', {
      headers: {
        Authorization: `Bearer ${user.token}`
      }
    })
    // Expected to return an array of strings
    return res.json()
  }
```

## Example Scenarios
The scenarios below show how the gateway behaves with the directives
### Scenario 1: Accessing `@authenticated` field when unauthorized

We have the supergraph schema below;
```graphql
  extend schema
    @link(
      url: "https://specs.apollo.dev/federation/v2.5"
      import: ["@authenticated"]
    )
  type Query {
    me: User @authenticated
    post(id: ID!): Post
  }

  type User {
    id: ID!
    username: String
    email: String
    posts: [Post!]!
  }

  type Post {
    id: ID!
    author: User!
    title: String!
    content: String!
    views: Int @authenticated
[...]

PR #538, opened 2025-10-30
feat(router): support authenticated and requiresScopes directives

kamilkisiela:
The implementation is based on the analysis from issue #351.

The authorization logic can be configured to operate in two modes:

- **`filter` (default mode):** This mode silently removes any fields from the incoming GraphQL operation that the user is not authorized to access. For each field removed, a corresponding `AuthorizationError` is added to the `errors` array in the GraphQL response, while the rest of the accessible data is returned as requested.
- **`reject` mode:** In this mode, the router will reject any GraphQL operation that attempts to access one or more unauthorized fields. The entire request is denied, and a descriptive error is returned.

Access control is defined within the supergraph schema using the following directives:

- `@authenticated`: Restricts access to a field to only authenticated users.
- `@requiresScopes(scopes: [[String]])`: Provides more granular control by requiring the user to possess specific scopes. The directive supports `AND` logic for scopes within a nested list (e.g., `["read:users", "write:users"]`) and `OR` logic for scopes across nested lists (e.g., `[["read:users"], ["admin:users"]]`).

## Key Implementation Details

- **Three-Phase Process:** The authorization logic is designed for high performance and is executed in three phases:
    1.  **Build Authorization Metadata:** when router loads the supergraph, it parses also the supergraph schema to create an in-memory representation of all `@authenticated` and `@requiresScopes` directives. This metadata is built once and reused for all incoming requests.
    2.  **Collect Unauthorized Paths:** For each incoming operation, the authorization logic traverses the query. It checks each field against the pre-built metadata and the current `UserAuthContext`. An `UnauthorizedPathTree` data structure is used to efficiently collect all field paths that the user is not permitted to access.
    3.  **Transform or Reject the Operation:** Based on the findings from the previous phase and the configured mode, a final decision is made. In `filter` mode, if any unauthorized paths were found, the original operation is rebuilt, stripping out only the unauthorized fields to create a new, sanitized operation. In `reject` mode, the presence of any unauthorized path leads to an rejection of the entire request.
- **Performance-First Design:** Authorization is a critical path in the request pipeline, this feature was developed with efficiency as a core principle. A new suite of benchmarks has been added to measure its performance impact. The results are good, demonstrating that the entire authorization process adds a negligible overhead of just **1-4 microseconds** for medium-sized queries, ensuring that security does not come at the cost of performance.
- **Pre-Query Planning Execution:** Authorization is performed on the operation's Abstract Syntax Tree (AST) before the query planning and execution stages. This ensures that unauthorized field requests never reach the subgraphs, preventing potential data leaks and unnecessary processing.
- **Decoupled from Authentication:** The core logic operates on a `UserAuthContext` struct, which contains the user's authentication status and a set of scopes. This decouples the authorization mechanism from the specific authentication method (e.g., JWT validation).
- **Comprehensive GraphQL Support:** The implementation correctly handles various GraphQL features to ensure authorization is applied accurately, including:
  - Conditional directives (`@include` and `@skip`), where authorization is bypassed for fields that are not included in the final operation.
  - Complex inheritance scenarios with interfaces and object types, correctly applying authorization rules defined on both.
- **Test Coverage:** Suite of end-to-end tests for both `filter` and `reject` modes, covering various scenarios.

### DIFF
--- a/bin/router/src/pipeline/authorization/mod.rs
+++ b/bin/router/src/pipeline/authorization/mod.rs
@@ -23,7 +23,7 @@ use crate::pipeline::authorization::tree::UnauthorizedPathTrie;
 use crate::pipeline::coerce_variables::CoerceVariablesPayload;
 use crate::pipeline::normalize::GraphQLNormalizationPayload;
 
-use hive_router_config::authentication::UnauthorizedMode;
+use hive_router_config::authorization::UnauthorizedMode;
 use hive_router_config::HiveRouterConfig;
 use hive_router_plan_executor::execution::client_request_details::JwtRequestDetails;
 use hive_router_plan_executor::introspection::schema::SchemaMetadata;
@@ -88,7 +88,7 @@ pub fn enforce_operation_authorization(
     variable_payload: &CoerceVariablesPayload,
     jwt_request_details: &JwtRequestDetails<'_>,
 ) -> AuthorizationDecision {
-    if !router_config.authentication.directives.enabled {
+    if !router_config.authorization.directives.enabled {
         return AuthorizationDecision::NoChange;
     }
 
@@ -97,7 +97,7 @@ pub fn enforce_operation_authorization(
     }
 
     let reject_mode =
-        router_config.authentication.directives.unauthorized.mode == UnauthorizedMode::Reject;
+        router_config.authorization.directives.unauthorized.mode == UnauthorizedMode::Reject;
 
     apply_authorization_to_operation(
         normalized_payload,
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,7 +4,7 @@
 
 |Name|Type|Description|Required|
 |----|----|-----------|--------|
-|[**authentication**](#authentication)|`object`|Default: `{"directives":{"enabled":true,"unauthorized":{"mode":"filter"}}}`<br/>|yes|
+|[**authorization**](#authorization)|`object`|Default: `{"directives":{"enabled":true,"unauthorized":{"mode":"filter"}}}`<br/>|yes|
 |[**cors**](#cors)|`object`|Configuration for CORS (Cross-Origin Resource Sharing).<br/>Default: `{"allow_any_origin":false,"allow_credentials":false,"enabled":false,"policies":[]}`<br/>|yes|
 |[**csrf**](#csrf)|`object`|Configuration for CSRF prevention.<br/>Default: `{"enabled":false,"required_headers":[]}`<br/>||
 |[**graphiql**](#graphiql)|`object`|Configuration for the GraphiQL interface.<br/>Default: `{"enabled":true}`<br/>||
@@ -22,7 +22,7 @@
 **Example**
 
 ```yaml
-authentication:
+authorization:
   directives:
     enabled: true
     unauthorized:
@@ -119,14 +119,14 @@ traffic_shaping:
 
 ```
 
-<a name="authentication"></a>
-## authentication: object
+<a name="authorization"></a>
+## authorization: object
 
 **Properties**
 
 |Name|Type|Description|Required|
 |----|----|-----------|--------|
-|[**directives**](#authenticationdirectives)|`object`||yes|
+|[**directives**](#authorizationdirectives)|`object`||yes|
 
 **Additional Properties:** not allowed  
 **Example**
@@ -139,15 +139,15 @@ directives:
 
 ```
 
-<a name="authenticationdirectives"></a>
-### authentication\.directives: object
+<a name="authorizationdirectives"></a>
+### authorization\.directives: object
 
 **Properties**
 
 |Name|Type|Description|Required|
 |----|----|-----------|--------|
 |**enabled**|`boolean`|Default: `true`<br/>||
-|[**unauthorized**](#authenticationdirectivesunauthorized)|`object`|Default: `{"mode":"filter"}`<br/>||
+|[**unauthorized**](#authorizationdirectivesunauthorized)|`object`|Default: `{"mode":"filter"}`<br/>||
 
 **Additional Properties:** not allowed  
 **Example**
@@ -159,8 +159,8 @@ unauthorized:
 
 ```
 
-<a name="authenticationdirectivesunauthorized"></a>
-#### authentication\.directives\.unauthorized: object
+<a name="authorizationdirectivesunauthorized"></a>
+#### authorization\.directives\.unauthorized: object
 
 **Properties**
 
--- a/e2e/configs/jwt_auth.directives.reject.router.yaml
+++ b/e2e/configs/jwt_auth.directives.reject.router.yaml
@@ -8,7 +8,7 @@ jwt:
   jwks_providers:
     - source: file
       path: ../jwks.rsa512.json
-authentication:
+authorization:
   directives:
     unauthorized:
       mode: reject
--- a/lib/router-config/src/authorization.rs
+++ b/lib/router-config/src/authorization.rs
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Default, Deserialize, Serialize, JsonSchema, Clone)]
 #[serde(deny_unknown_fields)]
-pub struct AuthenticationConfig {
+pub struct AuthorizationConfig {
     pub directives: AuthorizationDirectivesConfig,
 }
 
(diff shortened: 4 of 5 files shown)


bin/router/src/pipeline/execution.rs
```diff
@@ -24,15 +25,21 @@ enum ExposeQueryPlanMode {
     DryRun,
 }
 
+pub struct PlannedRequest<'req> {
```
dotansimha, line 28: Nice

bin/router/src/pipeline/execution.rs
```diff
@@ -24,15 +25,21 @@ enum ExposeQueryPlanMode {
...
+    pub normalized_payload: &'req GraphQLNormalizationPayload,
+    pub query_plan_payload: &'req Arc<QueryPlan>,
+    pub variable_payload: &'req CoerceVariablesPayload,
+    pub client_request_details: &'req ClientRequestDetails<'req, 'req>,
+    pub authorization_errors: &'req [AuthorizationError],
+}
+
+// TODO: less args
```
dotansimha: // TODO can be removed now, i think?

bin/router/src/pipeline/authorization.rs
```diff
@@ -0,0 +1,855 @@
...
+    /// Implies authentication as well.
+    RequiresScopes(RequiredScopes),
+}
+
+type TypeRulesMap = HashMap<String, AuthorizationRule>;
+type FieldRulesMap = HashMap<String, AuthorizationRule>;
+type TypeFieldRulesMap = HashMap<String, FieldRulesMap>;
+
+/// A container for all pre-computed authorization metadata.
+/// This is built once at startup and shared across all requests.
```
dotansimha: * startup -> schema (re)loading?

bin/router/src/pipeline/authorization.rs
```diff
@@ -0,0 +1,855 @@
...
+#[derive(Debug)]
+pub struct AuthorizationMetadata {
+    /// Rules applied to types
+    pub type_rules: TypeRulesMap,
+    /// Rules applied to specific fields - (TypeName, FieldName) tuple.
+    pub field_rules: TypeFieldRulesMap,
```
dotansimha: I might have missed something here, but I noticed that `TypeFieldRulesMap`'s key is `String` and not a `(String, String)`. Also when created, it uses `field_rules.insert(type_name.clone(), type_field_rules);` and I expected the field also to be there?

bin/router/src/pipeline/authorization.rs
```diff
@@ -0,0 +1,855 @@
+use crate::pipeline::coerce_variables::CoerceVariablesPayload;
```
dotansimha: nit: this file grew a lot and I can see a several areas where it can be splited.

lib/query-planner/src/ast/value.rs
```diff
@@ -22,6 +22,34 @@ pub enum Value {
...
+// We manually implement `PartialEq` and `Eq` for `Value` because the `f64` type,
+// used in the `Float` variant, does not implement `Eq`. According to the IEEE 754
+// standard, `f64::NAN != f64::NAN`, which violates the reflexivity rule
+// required for `Eq` (where `a == a` must always be true).
+//
+// To work around this, we compare the underlying bit patterns of the `f64` values
+// using `to_bits()`. This method provides a total ordering for all `f64` values,
+// including `NAN`, and ensures that our equality logic is consistent with the
+// `Hash` implementation, which uses the same technique.
+impl PartialEq for Value {
```
dotansimha, line 34: nice catch

lib/query-planner/src/federation_spec/authorization.rs
```diff
@@ -0,0 +1,89 @@
...
+    where
+        Self: Sized,
+    {
+        let mut result = Self {
+            scopes: AstValue::Null,
+        };
+
+        for (arg_name, arg_value) in &directive.arguments {
+            if arg_name.eq("scopes") {
+                // I don't like the fact it's converted from graphql_parser::Value
```
dotansimha, line 62: Yeah, I agree this part of using AST `Value` is not ideal. 

What prevents us from storing here as raw `Vec<Vec<String>>` (the way it's defined in the direcrtive)

bin/router/src/pipeline/mod.rs
```diff
@@ -169,6 +171,43 @@ pub async fn execute_pipeline(
...
+    );
+
+    let (normalize_payload, authorization_errors) = match decision {
+        AuthorizationDecision::NoChange => (normalize_payload.clone(), vec![]),
+        AuthorizationDecision::Modified {
+            new_operation_definition,
+            errors,
+        } => {
+            (
+                Arc::new(GraphQLNormalizationPayload {
```
dotansimha, line 191: nit: I wonder if this should be "cleared" from here and be something like `normalize_payload.next(...)` that will deal with the clones?

bin/router/src/schema_state.rs
```diff
@@ -118,6 +125,8 @@ impl SchemaState {
         let supergraph_state = SupergraphState::new(&parsed_supergraph_sdl);
         let planner = Planner::new_from_supergraph(&parsed_supergraph_sdl)?;
         let metadata = planner.consumer_schema.schema_metadata();
+        // TODO: use supergraph
```
dotansimha: TODO

bin/router/src/pipeline/authorization.rs
kamilkisiela: old comment, it's still a draft

bin/router/src/pipeline/mod.rs
kamilkisiela: but why `next`? It lacks meaning and context

bin/router/src/schema_state.rs
kamilkisiela: It's a draft :) You have no idea how many more `TODO:` comments I have locally still

lib/query-planner/src/federation_spec/authorization.rs
kamilkisiela: Lack of error handling, if something is not "right" then we can't error. See the rest of the FederationDirective impls.

The error handling is done when reading the extracted directives.

bin/router/src/pipeline/authorization.rs
kamilkisiela: as well, yes

bin/router/src/pipeline/authorization/metadata.rs
```diff
@@ -0,0 +1,515 @@
...
+                (None, Some(rule)) => {
+                    // Interface has no rule, implementor has rule
+                    combined_rule = Some(rule);
+                }
+                (Some(_), None) => {
+                    // Interface has rule, implementor has no rule
+                    // No auth on implementor means interface field is accessible without auth
+                    // (AND semantics: if any implementor allows, we can't require auth)
+                    return None;
+                }
```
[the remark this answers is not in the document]
kamilkisiela: @gemini-code-assist fixed

bin/router/src/pipeline/authorization/collector.rs
```diff
@@ -0,0 +1,519 @@
...
+                        FieldAuthStatus::UnauthorizedNonNullable
+                    } else {
+                        FieldAuthStatus::UnauthorizedNullable
+                    };
+
+                    if status == FieldAuthStatus::UnauthorizedNonNullable
+                        && has_non_null_unauthorized == &false
+                    {
+                        *has_non_null_unauthorized = true;
+                    }
```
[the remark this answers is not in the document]
kamilkisiela: done

bin/router/src/pipeline/authorization/collector.rs
```diff
@@ -0,0 +1,519 @@
...
+                        });
+                    }
+                }
+                SelectionItem::FragmentSpread(_) => {
+                    // Fragment spreads are inlined during normalization, so we can skip them here.
+                }
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
kamilkisiela: done

bench/subgraphs/accounts.rs
```diff
@@ -42,14 +44,71 @@ lazy_static! {
...
+                    self.username.as_ref().unwrap_or(&"unknown".to_string())
+                ),
+                handle: self
+                    .username
+                    .as_ref()
+                    .unwrap_or(&"unknown".to_string())
+                    .clone(),
+                repo_count: 42,
+            }),
+        ]
```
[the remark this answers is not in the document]
kamilkisiela: Ignored

PR #540, opened 2025-10-31
Shared utilities to handle VRL expressions & some other improvements

ardatan:
Removed `pool_idle_timeout_seconds` from `traffic_shaping`, instead use `pool_idle_timeout` with duration format.

```diff
traffic_shaping:
-  pool_idle_timeout_seconds: 30
+  pool_idle_timeout: 30s
```

Related documentation changes PR -> https://github.com/graphql-hive/console/pull/7207

Created some helper functions to deduplicate the code to compile and resolve VRL expressions

### DIFF
--- a/lib/executor/src/executors/map.rs
+++ b/lib/executor/src/executors/map.rs
@@ -154,7 +154,12 @@ impl SubgraphExecutorMap {
                     client_request,
                 )
             })
-            .unwrap_or_else(|| self.get_executor_from_static_endpoint(subgraph_name))
+            .unwrap_or_else(|| {
+                self.get_executor_from_static_endpoint(subgraph_name)
+                    .ok_or_else(|| {
+                        SubgraphExecutorError::StaticEndpointNotFound(subgraph_name.to_string())
+                    })
+            })
     }
 
     /// Looks up a subgraph executor,
@@ -207,7 +212,7 @@ impl SubgraphExecutorMap {
     fn get_executor_from_static_endpoint(
         &self,
         subgraph_name: &str,
-    ) -> Result<SubgraphExecutorBoxedArc, SubgraphExecutorError> {
+    ) -> Option<SubgraphExecutorBoxedArc> {
         self.static_endpoints_by_subgraph
             .get(subgraph_name)
             .and_then(|endpoint_ref| {
@@ -216,7 +221,6 @@ impl SubgraphExecutorMap {
                     .get(subgraph_name)
                     .and_then(|endpoints| endpoints.get(endpoint_str).map(|e| e.clone()))
             })
-            .ok_or_else(|| SubgraphExecutorError::StaticEndpointNotFound(subgraph_name.to_string()))
     }
 
     /// Registers a VRL expression for the given subgraph name.


lib/router-config/src/primitives/expression.rs
```diff
@@ -0,0 +1,144 @@
...
+};
+
+#[derive(Debug, Clone)]
+pub struct Expression {
+    expression: String,
+    program: Box<VrlProgram>,
+}
+
+static VRL_FUNCTIONS: Lazy<Vec<Box<dyn Function>>> = Lazy::new(vrl_build_functions);
+static VRL_TIMEZONE: Lazy<VrlTimeZone> = Lazy::new(VrlTimeZone::default);
```
kamilkisiela: Let's not do it this way, where we centralize the function building.
Yeah, right now we pass all functions that vrl provides, but if we ever want to add one made by us or filter some, for certain expressions, we would have to opt out and, we won't be able to do it without basically reverting your changes

ardatan: I just didn't want to copy/paste the same logic again like we did 3 different places before. How do you think I should do that instead?

kamilkisiela: You can create a function and reuse it. If we ever want to opt-out, we either slightly refactor it or duplicate.

if you move this logic to the config crate then we’re not able to do those small adjustments in future

ardatan: Looks better now?

lib/router-config/src/override_labels.rs
```diff
@@ -15,10 +15,7 @@ pub enum LabelOverrideValue {
     /// A static boolean value to enable or disable the label.
     Boolean(bool),
     /// A dynamic value computed by an expression.
-    Expression {
-        /// An expression that must evaluate to a boolean. If true, the label will be applied.
-        expression: String,
-    },
+    Expression { expression: String },
```
kamilkisiela: original comment is missing

lib/executor/Cargo.toml
```diff
@@ -49,6 +49,8 @@ itoa = "1.0.15"
 ryu = "1.0.20"
 indexmap = "2.10.0"
 bumpalo = "3.19.0"
+once_cell = "1.21.3"
+schemars = "1.0.4"
```
kamilkisiela: you sure we need schemars in executor?

lib/router-config/src/override_labels.rs
ardatan: Reverted 👍

lib/executor/Cargo.toml
ardatan: Removed 👍

lib/router-config/src/primitives/expression.rs
kamilkisiela: Yes, it's much better now. Thanks

If we ever want to use different set of function, we can now do:

```
compile_expression(expression, Some(fns))
```

and add pre-built functions anywhere in code, where it makes sense.

kamilkisiela (approved):

PR #542, opened 2025-10-31
feat(router): HMAC based Subgraph Auth

ardatan:
Ref ROUTER-123

Documentation -> https://github.com/graphql-hive/console/pull/7215

Following the same principle in Hive Gateway. So the existing Yoga and Apollo recipes can be used for subgraphs.

```yaml
hmac_signature:
   enabled: false #default
   enabled: true # signs all requests
   enabled:
       expression: |
      # enable only for products
      if .subgraph.name == "products"
         true
      else
        false
   secret: MY_SECRET # required
   extension_name: hmac-signature # by default
```


dotansimha:
Closing for now while keeping the branch, doing some PR housekeeping

PR #543, opened 2025-10-31
feat(router): AWS Sigv4

ardatan:
Ref [ROUTER-123](https://linear.app/the-guild/issue/ROUTER-123)
Similar to https://www.apollographql.com/docs/graphos/routing/security/subgraph-authentication#:~:text=The%20GraphOS%20Router%20and%20Apollo,hasn't%20been%20tampered%20with.

AWS's official SDK was too complex. I used https://github.com/apache/opendal-reqsign of Apache which supports other cloud providers, too. In the future, we can add support for these too, maybe?

Documentation -> https://github.com/graphql-hive/console/pull/7235


dotansimha:
Closing for now while keeping the branch, doing some PR housekeeping

Issue #579, opened 2025-11-20
Issue @include @skip directive, filtered out field requested to subgraphs


arsonik:
Hey 

While attempting to migrate from Apollo rover to Hive Router in our CI.

I came across a weird error.
The Query planner or I dont know who is in charge of that
Is actually removing / including fields filtered by @include/@skip (ok for that)

But it is doing it after calling the related subgraph without filtering the fields first !
Conclusion the router is querying the unnecessary fields from the subgraphs.

I can observe the behavior by adding a breakpoint (or logging the `info.variableValues` - 4th parameter) in the concerned field resolver

https://github.com/graphql-hive/router/pull/238

Example query : 
```graphql
query user($slug: ID!, $include: Boolean!, $skip: Boolean!) {
    user(slug: $slug) {
        __typename
        ... on User {
            slug
            userId
            displayName @include(if: $include)
            score @skip(if: $skip)
        }
    }
}
```

running with these variables 
```
 {
                        slug: "user-1",
                        include: false,
                        skip: true,
                    },
```
querying against the hive router info.variableValues return an object with slug only
querying against the subgraph directly info.variableValues return an object with slug, include, skip

Any thoughts ?

Best

PR #591, opened 2025-11-28
fix: support `@include` and `@skip` in initial fetch node

kamilkisiela:
Fixed an issue where `@skip` and `@include` directives were incorrectly removed from the initial Fetch of the Query Plan.

References #579

### DIFF
--- a/lib/query-planner/src/ast/type_aware_selection.rs
+++ b/lib/query-planner/src/ast/type_aware_selection.rs
@@ -89,10 +89,6 @@ impl TypeAwareSelection {
 
             Ok(())
         } else {
-            println!(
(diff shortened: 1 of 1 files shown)


lib/query-planner/src/ast/type_aware_selection.rs
```diff
@@ -89,6 +89,10 @@ impl TypeAwareSelection {
 
             Ok(())
         } else {
+            println!(
+                "Failed to find path '{}' in selection set '{}'",
+                add_at_fetch_path, self.selection_set
+            );
```
kamilkisiela:
```suggestion
```

PR #602, opened 2025-12-04
fix(query-planner): prevent planner failure when combining conditional directives and interfaces

ardatan:
Fixed a bug where the query planner failed to handle the combination of conditional directives (`@include`/`@skip`) and the automatic `__typename` injection required for abstract types.

The condition was passed to the fetch path and response path, but it shouldn't.


ardatan:
@gemini-code-assist try to fix it

kamilkisiela (approved):

Issue #604, opened 2025-12-08
Introspection query fails


rickbijkerk:
<img width="583" height="816" alt="Image" src="https://github.com/user-attachments/assets/3dc21753-e486-4445-9155-fe7b26b983f5" />

most basic hive router setup with our graphql schema isnt working 

config:
```
# router.config.yaml
supergraph:
  source: file
  path: ./supergraph.graphql

```

(best to contact me directly through dm's and ill be able to share some info on our schema with you )

rickbijkerk:
Yes other supergraph do work it's just our supergraph that isnt working.

network tab on the graphql request 
<img width="1368" height="22" alt="Image" src="https://github.com/user-attachments/assets/c7535617-48bb-4d1a-808f-20b24ce3f71c" />

let me know if you need more info

ardatan:
Could you run the router with `LOG_LEVEL=trace` environment variable, make it fail and share the log?

ardatan:
I did some small changes to learn more about the issue;
Could you download the alpha release (released binaries from https://github.com/graphql-hive/router/actions/runs/20058123964?pr=606)  and try with it?

[hive_router_linux_amd64](https://github.com/graphql-hive/router/actions/runs/20058123964/artifacts/4808501035)
[hive_router_linux_arm64](https://github.com/graphql-hive/router/actions/runs/20058123964/artifacts/4808502301)
[hive_router_macos_amd64](https://github.com/graphql-hive/router/actions/runs/20058123964/artifacts/4808489498)
[hive_router_macos_arm64](https://github.com/graphql-hive/router/actions/runs/20058123964/artifacts/4808512546)

rickbijkerk:
we're getting somewhere, previous error disappeared, now im seeing

<img width="349" height="262" alt="Image" src="https://github.com/user-attachments/assets/41d9a425-b156-4610-ae8f-b0e68ed47fdb" />

/graphql response seems to be empty

<img width="1357" height="24" alt="Image" src="https://github.com/user-attachments/assets/51f87a92-a929-4ec9-b726-d768e02dc405" />

ardatan:
Hmm I am not sure if it is empty. Because;
- `Decorated type deeper than introspection query.` error comes from GraphiQL while parsing the introspection result.
- Also it has `200` with some response size.

I have a few questions;
- Could you check the response on the right view for the entry you shared in the screenshot in `Network` tab?
[...]

PR #606, opened 2025-12-09
fix(executor): strip `@join__directive` and `join__DirectiveArguments` for the public consumer schema

ardatan:
Strip `@join__directive` and `join__DirectiveArguments` internal types while creating the consumer/public schema

When join__directive remains in the schema, it has `join__Graph` in it and introspection was failing because join__Graph is stripped already.

Fixes https://github.com/graphql-hive/router/issues/604

### DIFF
--- a/lib/executor/src/introspection/resolve.rs
+++ b/lib/executor/src/introspection/resolve.rs
@@ -400,7 +400,7 @@ fn resolve_type<'exec, 'schema: 'exec>(
     match t {
         Type::NamedType(name) => {
             let type_def = ctx.schema.type_by_name(name).unwrap_or_else(|| {
-                unreachable!(
+                panic!(
                     "Type '{}' not found in the schema unexpectedly during introspection",
                     name
                 );


.changeset/avoid-panic-introspection.md
```diff
@@ -0,0 +1,6 @@
+---
+executor: patch
+router: patch
+---
+
+Avoid panicking when a type reference in another definition cannot be found in the schema during introspection. Instead, log a trace message and return only `name` field for that reference and `null` for the rest.
```
kamilkisiela: As I said on Slack, we should either keep the panic here or fail the whole query when an error in the introspection happens due to invalid state of the schema, and introduce a validation logic of the public schema state (https://linear.app/the-guild/issue/ROUTER-225/validate-public-schema-when-supergraph-is-loaded).

Can you revert those changes and keep only the join direct ive stripping part?

kamilkisiela (changes requested):

PR #620, opened 2025-12-13
Subscriptions

### LINKED ISSUE #313: RFC: GraphQL Subscriptions
- transports (to/from client, to/from subgraphs) 
- configurations 
- technical implementation aspect (perf? memory?)

enisdenjo:
Closes #313
Ref router-165

Implement [SSE](https://github.com/enisdenjo/graphql-sse/blob/918b9d414a4f938c715a8f8bc13184c0849eac50/PROTOCOL.md#distinct-connections-mode), [Incremental Delivery over HTTP](https://github.com/graphql/graphql-over-http/blob/d312e43384006fa323b918d49cfd9fbd76ac1257/rfcs/IncrementalDelivery.md), [Apollo's Multipart HTTP](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol), [WebSocket](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) and [HTTP Callback](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/callback-protocol) specs for subscriptions when communicating with subgraphs or clients with entity resolution capabilities.

# Documentation

[See PR in docs repo](https://github.com/graphql-hive/docs/pull/66).

# Streaming execution result

```
lib/executor/src/execution/plan.rs@QueryPlanExecutionResult
```

The execution pipeline returns `QueryPlanExecutionResult` - an enum that can be either a single response or a stream:

```rust
pub enum QueryPlanExecutionResult {
    Single(PlanExecutionOutput),
    Stream(PlanSubscriptionOutput),
}
```

This separation allows the query planner and executor to operate independently from transport concerns, making future improvements (like connection repair and silent retries) easier to implement.

# Owned query plan execution context values for streaming

```
lib/executor/src/execution/plan.rs
```

Subscriptions require long-lived contexts that outlive request lifetimes. The subscription stream closure requires `'static` data, so `QueryPlanExecutionOpts` fields captured by the stream are `Arc`-wrapped so they can be cheaply cloned per event without deep-copying:

- `operation_for_plan: Arc<OperationDefinition>`
- `variable_values: Arc<Option<HashMap<String, sonic_rs::Value>>>`
- `client_request: Arc<ClientRequestDetails>`
- `introspection_context: Arc<IntrospectionContext>`
- `executors`, `projection_plan`, `headers_plan` are `Arc` at the call site

`IntrospectionContext` carries no lifetime parameter - `query` is `Option<Arc<OperationDefinition>>` and `schema` is `Arc<Document>`, both sourced from already-`Arc`-wrapped data in `SupergraphData` (`planner.consumer_schema.document`) and `SchemaMetadata`.

`SupergraphData.subgraph_executor_map`, `SupergraphData.metadata`, and `RouterSharedState.headers_plan` are `Arc` at the source and are threaded through `QueryPlanExecutionOpts` and `IntrospectionContext` as `Arc` instead of borrowed references. `SubgraphExecutorMap` contains multiple `DashMap`s so keeping it behind `Arc` is especially important. The subscription path does cheap ref-count bumps. The query/mutation path just derefs.

This enables entity resolution to happen independently for each subscription event and allows remaining plan nodes to be processed after the subscription fetch.

# Unified execution inner path for subscriptions

```
lib/executor/src/execution/plan.rs@execute_query_plan_with_data
```

The post-fetch logic (plugin hooks, plan node execution, projection, error aggregation) lives in a private `execute_query_plan_with_data(data, opts)` function shared by both the query/mutation path and the subscription stream. Each subscription event reconstructs an `opts` from cloned `Arc`s and calls `execute_query_plan_with_data`, meaning subscription events go through the full execution pipeline including plugins.

# Error serialization helper

```
lib/executor/src/execution/plan.rs@FailedExecutionResult
```

`FailedExecutionResult` is a serializable struct with a `serialize()` method (including a safe fallback if serialization itself fails) used in both the subscription stream and the pipeline error handler.

# Error messages in the subscription stream

When a subscription subgraph stream yields an error it is converted through `with_plan_context` (which attaches the subgraph name) and then `From<PlanExecutionError> for GraphQLError`, producing a message that includes the actual underlying error detail. For example:

```
"Error reading SSE subscription stream: Stream read error: error reading a body from connection"
```

# Subscription handlers

The router respects the client's `Accept` header to determine response format:

- `text/event-stream` → SSE responses
- `multipart/mixed` → Incremental Delivery over HTTP
- `multipart/mixed;subscriptionSpec="1.0"` → Apollo multipart HTTP
- Returns `406 Not Acceptable` if subscription is requested over unsupported transport
- Handles errors by emitting an error event and completing the stream
- Heartbeats every 10 seconds (except for incremental delivery, doesn't have it)
- Of course protocols used between router and subgraphs and clients can be different

Same behavior is expected when communicating with subgraphs.

## SSE

```
lib/executor/src/executors/sse.rs
```

Implements the [GraphQL over SSE spec distinct connection mode](https://github.com/enisdenjo/graphql-sse/blob/918b9d414a4f938c715a8f8bc13184c0849eac50/PROTOCOL.md#distinct-connections-mode).

## Multipart protocol

```
lib/executor/src/executors/multipart_subscribe.rs
```

Implements [Apollo's multipart HTTP spec](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol) and [GraphQL's Incremental Delivery over HTTP RFC](https://github.com/graphql/graphql-over-http/blob/d312e43384006fa323b918d49cfd9fbd76ac1257/rfcs/IncrementalDelivery.md).

# WebSockets

Read stacked PR #738

## Plugin system

The WS server assembles a synthetic http request for every ws operation message allowing the plugin system to run as usual for the supported execution paths. (No subscriptions because the plugin system generally does not support subscriptions - yet).

# HTTP Callbacks

Read stacked PR #831

# Entity resolution

```
lib/executor/src/execution/plan.rs@execute_query_plan
```

When a subscription emits data that references entities from other subgraphs, the router:

1. Receives subscription event from primary subgraph
2. Executes remaining plan nodes (flatten, fetch, etc.) to populate missing fields
3. Projects the complete response
4. Streams the full response to client

# Subscription node in query plan

In `PlanNode::from_fetch_step` in `plan_nodes.rs`, when building a node from a fetch step, if the step has no response path and its operation kind is Subscription, it directly produces a `PlanNode::Subscription` instead of `PlanNode::Fetch`.

# Subgraph executor subscribe method

```
lib/executor/src/executors/common.rs@subscribe
```

The HTTP executor has a `subscribe()` method that:

- Negotiates content-type (prefers multipart, falls back to SSE)
- Establishes long-lived connections to subgraphs
- Returns a `BoxStream<HttpExecutionResponse>` for downstream processing

# WebSocket executor

```
lib/executor/src/executors/websocket.rs
lib/executor/src/executors/graphql_transport_ws.rs
```

Request preparation (building a `SubscribePayload` and optional `ConnectionInitPayload` from a `SubgraphExecutionRequest`) is handled by a shared `build_subscribe_payload` function. `WsClient::subscribe` takes a single `SubscribePayload` argument. `SubscribePayload` derives `Default`.

# HTTP callback executor

```
lib/executor/src/executors/http_callback.rs
```

The `subscription` extension is injected by mutating `execution_request.extensions` in place via `get_or_insert_default()` before delegating to the shared `build_request_body`.

# Configure to only enable/disable subscriptions

The supported subscription protocols are inherently HTTP and do not need a "protocol" configuration option. Hive Router sends an accept header listing all supported protocols for subscriptions over HTTP and the subgraph is free to choose whichever one it supports.

Whether we really _want_ to limit specific protocols is up to discussion but objectively there is no benefit since they're all streamed HTTP.

Hence, you can only:

```yaml
supergraph:
  source: file
  path: supergraph.graphql
subscriptions:
  enabled: true
```

P.S. if the subscriptions are disabled, the router will respond when receiving a subscription request:

```
HTTP/1.1 415 Unsupported Media Type
Content-Type: application/json

{"errors":[{"message":"Subscriptions are not supported","extensions":{"code":"SUBSCRIPTIONS_NOT_SUPPORTED"}}]}
```

# No Silent Retries & Upstream Connection Repair

Most GraphQL subgraph implementations are stateless for subscriptions and have no concept of "continuing" from where they left off after a connection loss. Implementing retry logic on the router side would create false expectations - users would assume all events are delivered, but some would be lost when the subgraph creates a fresh subscription.

This is fundamentally why the EDFS (Event-Driven Federated Subscriptions) and callback protocols exist. To avoid misleading behavior and keep the PR focused on the core functionality, connection repair is not implemented.

# TODO

- [x] [documentation](https://github.com/graphql-hive/docs/pull/66)
- [x] changesets
- [x] e2e tests for error handling
- [x] e2e tests for header propagation
- [x] use details from the actual client request for entity resolution towards subgraphs of events
- [x] test subgraph disconnect propagation
- [ ] use apollo's multipart http payload for transport erros
- [ ] ~~test client disconnect propagation~~
  - hard to test, would need changing the source code because async being state machines allows it just to stop execution in any suspended state
- [x] #921
- [x] entity resolution failures dont end the stream, should they? a subgraph might recover, usually you dont end the whole stream. ~~**they should not end the stream**~~ actually they SHOULD because the router might not guarantee the subgraph will recover and the client might ignore the error keeping the gateway in an infinite error loop wasting resources
- [x] configure enable/disable subscriptions
- [ ] ~~configure accepted protocols~~
  - no need, all these here fall under "http" protocols and can be negotiated over http. read description
- [x] stream errors when accepting only streaming content types
- [ ] validate config allowing only one subscriptions transport
- [ ] #922
- [x] remove ntex patch https://github.com/ntex-rs/ntex/pull/790 and after main is rebased because there are some other necessary refactors for latest ntex v3


lib/executor/src/executors/http.rs
```diff
@@ -306,4 +309,230 @@ impl SubgraphExecutor for HTTPSubgraphExecutor {
...
+
+                            let graphql_error: GraphQLError = error.into();
+                            let graphql_error = graphql_error.add_subgraph_name(&subgraph_name);
+                            let errors = vec![graphql_error];
+                            let errors_bytes = sonic_rs::to_vec(&errors).unwrap();
+                            let mut buffer = BytesMut::new();
+                            buffer.put_slice(b"{\"errors\":");
+                            buffer.put_slice(&errors_bytes);
+                            buffer.put_slice(b"}");
+                            let error_bytes = buffer.freeze();
```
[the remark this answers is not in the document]
enisdenjo: That's fine.

lib/executor/src/executors/http.rs
```diff
@@ -306,4 +309,230 @@ impl SubgraphExecutor for HTTPSubgraphExecutor {
...
```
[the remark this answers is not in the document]
enisdenjo: that's fine.

bench/subgraphs/reviews.rs
```diff
@@ -167,8 +168,76 @@ impl Query {
...
+    }
+
+    async fn review_added_for_product(
+        &self,
+        product_upc: String,
+        #[graphql(default = 1_000)] interval_in_ms: u64,
+    ) -> impl Stream<Item = Review> {
+        let reviews_for_product: Vec<Review> = REVIEWS
+            .iter()
+            .filter(move |r| r.product.as_ref().unwrap().upc == product_upc)
```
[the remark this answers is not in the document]
enisdenjo: It's for tests.

lib/executor/src/execution/plan.rs
```diff
@@ -127,6 +406,16 @@ pub async fn execute_query_plan<'exec, 'req>(
...
 
+/// Format an error response as JSON bytes
+fn format_error_response(errors: &[GraphQLError]) -> Vec<u8> {
+    let errors_json: Vec<String> = errors
+        .iter()
+        .map(|e| format!(r#"{{"message":"{}"}}"#, e.message.replace('"', "\\\"")))
+        .collect();
+
+    format!(r#"{{"data":null,"errors":[{}]}}"#, errors_json.join(",")).into_bytes()
+}
```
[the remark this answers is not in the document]
enisdenjo: It's ok.

bin/router/src/pipeline/multipart_subscribe.rs
```diff
@@ -0,0 +1,49 @@
...
+                                    // As per the spec.
+                                    yield Ok(Bytes::from(format!(r#"{{"payload":{}}}"#, json_str)));
+                                    yield Ok(Bytes::from("\r\n"));
+                                }
+                                Err(e) => {
+                                    // TODO: transport level errors as per spec
+                                    yield Err(std::io::Error::new(std::io::ErrorKind::InvalidData, e));
+                                    break;
+                                }
+                            }
```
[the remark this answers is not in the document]
enisdenjo: Thanks, done.

bin/router/src/pipeline/sse.rs
```diff
@@ -0,0 +1,44 @@
...
+                        Some(resp) => {
+                            match String::from_utf8(resp) {
+                                Ok(json_str) => {
+                                    yield Ok(Bytes::from(format!("event: next\ndata: {}\n\n", json_str)));
+                                }
+                                Err(e) => {
+                                    yield Err(std::io::Error::new(std::io::ErrorKind::InvalidData, e));
+                                    break;
+                                }
+                            }
```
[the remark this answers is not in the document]
enisdenjo: Thanks, done.

lib/executor/src/response/subgraph_response.rs
```diff
@@ -89,6 +94,57 @@ impl<'de> de::Deserialize<'de> for SubgraphResponse<'de> {
     }
 }
 
+impl GraphQLError {
+    pub fn to_subgraph_response(self, subgraph_name: &str) -> SubgraphResponse<'static> {
```
ardatan: For a future reference; (not related to your PR directly but I don't like this), I hope we get the following PR soon to get this fixed so that we will handle errors as usual;
https://github.com/graphql-hive/router/pull/659

lib/executor/src/execution/error.rs
```diff
@@ -14,6 +14,10 @@ pub enum PlanExecutionErrorKind {
     #[error(transparent)]
     #[strum(serialize = "HEADER_PROPAGATION_FAILURE")]
     HeaderPropagation(#[from] HeaderRuleRuntimeError),
+
+    #[error("{0}")]
+    #[strum(serialize = "INTERNAL_ERROR")]
+    Internal(String),
```
ardatan: Why do we need this?

lib/executor/src/execution/jwt_forward.rs
```diff
@@ -2,7 +2,7 @@ use sonic_rs::Value;
 
 use crate::execution::client_request_details::JwtRequestDetails;
 
-#[derive(Default)]
+#[derive(Default, Clone)]
```
ardatan: Is it possible to avoid cloning with `Arc`?

lib/executor/src/execution/error.rs
enisdenjo: experimenting leftovers... https://github.com/graphql-hive/router/pull/620/commits/a5df42516c0053608fd5ece4fe93dd095d9f623a

lib/executor/src/execution/jwt_forward.rs
enisdenjo: It's done only once when setting up the subscriptions, then the JWT (and all other necessary parts) are moved into the stream closure and then passed by reference to other parts as with regular query/mutation execution.

I didn't want to optimize as this seemed like an overkill, atomically counting references vs. cloning two strings once.

What do you think?

enisdenjo: Note that heavy things from the qp executor context are `Arc`ed. Things like the projection plan, executors, schema metadata. Structs containing a handful of primitive values are cloned.

Also note that this is done only once during subscription setup.

lib/query-planner/src/planner/plan_nodes.rs
```diff
@@ -359,7 +359,8 @@ pub struct ValueSetter {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct SubscriptionNode {
-    pub primary: Box<PlanNode>, // Use Box to prevent size issues
+    // A subscription node can only really have a primary fetch node.
+    pub primary: FetchNode,
```
enisdenjo, line 363: FetchNode is a concrete type and not an enum so I dropped the box because the compiler will know the size. Right?

ardatan (approved):

lib/executor/src/executors/http.rs
```diff
@@ -481,6 +481,138 @@ impl SubgraphExecutor for HTTPSubgraphExecutor {
...
+        let (parts, body_stream) = res.into_parts();
+
+        let content_type = parts
+            .headers
+            .get(http::header::CONTENT_TYPE)
+            .and_then(|v| v.to_str().ok())
+            .unwrap_or("");
+
+        let is_multipart = content_type.starts_with("multipart/mixed");
+        let is_sse = content_type == "text/event-stream";
```
kamilkisiela: @enisdenjo This strict match rejects `text/event-stream; charset=utf-8`. Is it fine?

lib/executor/src/executors/map.rs
```diff
@@ -378,27 +402,121 @@ impl SubgraphExecutorMap {
...
+                    semaphore,
+                    subgraph_config.dedupe_enabled,
+                    self.in_flight_requests.clone(),
+                    self.telemetry_context.clone(),
+                    self.config.clone(),
+                )
+                .to_boxed_arc();
 
-        let executor_arc = executor.to_boxed_arc();
+                self.http_executors_by_subgraph
```
kamilkisiela, line 429: @enisdenjo is it fine that `HTTP` variant is written to `http_executors_by_subgraph`, but `WebSocket` and `HTTPCallback` to `subscription_executors_by_subgraph`?

The `get_or_create_subscription_executor` is reading from `subscription_executors_by_subgraph`

lib/executor/src/executors/http.rs
enisdenjo: SSE event streams _must_ be encoded in utf-8, the charset would not need to be specified and thefore redundant. But it _can_ happen coming from weird clients - but I am not aware of a real life scenario.

P.S. charsets in accept headers are quite rare. They are, however, common in content-type headers.

lib/executor/src/executors/map.rs
enisdenjo: Yeah, that was intentional. It was in the PR description once I think, but got lost a while back.

The idea is that http executors can be both streaming and single shot (queries, mutations and subscriptions), while subscription executors can be only streaming (and hence only for subscriptions).

bin/router/src/lib.rs
```diff
@@ -338,11 +420,23 @@ impl RouterPaths {
...
+    cfg.route(&callback_route, web::post().to(handler));
+}
+
 pub fn configure_ntex_app(
     cfg: &mut web::ServiceConfig,
     paths: &RouterPaths,
     prometheus: Option<PrometheusAttached>,
 ) {
+    if let Some(websocket) = &paths.websocket {
+        cfg.route(websocket.as_str(), web::get().to(ws_index));
```
kamilkisiela: @enisdenjo If the subscription path is the same as graphql endpoint, won't the WS GET put a shadow over the normal GraphQL `GET` request, so that the `GET /graphql?query=...` won't reach the GraphQL handler?

bin/router/src/pipeline/error.rs
```diff
@@ -291,5 +294,42 @@ pub fn handle_pipeline_error(
             .record_errors(|| errors.iter().map(|error| error.extensions.code.as_deref()));
     }
 
-    res.json(&FailedExecutionResult { errors })
+    let data = FailedExecutionResult { errors }.serialize();
+
+    match response_mode {
+        ResponseMode::SingleOnly(_) | ResponseMode::Dual(_, _) => res
+            .header(header::CONTENT_TYPE, "application/json")
```
kamilkisiela: @enisdenjo what about `application/graphql-response+json` when requested

enisdenjo: https://github.com/graphql-hive/router/pull/620/commits/aa7aacb712b8e40c7344b6d4850c5a41fe638cf1

bin/router/src/lib.rs
enisdenjo: yeah right! fixing

lib/executor/src/executors/http.rs
kamilkisiela: Maybe we shouldn't be that strict?

lib/executor/src/executors/map.rs
kamilkisiela: I see

bin/router/src/lib.rs
enisdenjo: https://github.com/graphql-hive/router/pull/620/commits/bf1bd96bfc741019600277e5bc12cf050e54e82d

lib/executor/src/executors/http_callback.rs
```diff
@@ -0,0 +1,285 @@
...
+    ) -> Result<
+        BoxStream<'static, Result<SubgraphResponse<'static>, SubgraphExecutorError>>,
+        SubgraphExecutorError,
+    > {
+        let subscription_id = Uuid::new_v4().to_string();
+        let verifier = Uuid::new_v4().to_string();
+
+        let body = self.build_request_body(&mut execution_request, &subscription_id, &verifier)?;
+
+        let (tx, mut rx) = mpsc::unbounded_channel::<CallbackMessage>();
```
kamilkisiela: @enisdenjo 
Unbounded channels can cause memory issues. When producer outpaces consumer, the memory grows and grows until OOM.

I'm not sure how possible it is in this case, but I will let you judge it :)

lib/executor/src/executors/websocket.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    async fn subscribe<'a>(
+        &self,
+        execution_request: SubgraphExecutionRequest<'a>,
+        _timeout: Option<Duration>,
+    ) -> Result<
+        BoxStream<'static, Result<SubgraphResponse<'static>, SubgraphExecutorError>>,
+        SubgraphExecutorError,
+    > {
+        let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel::<
```
kamilkisiela: @enisdenjo

Similar story to http_callback.rs

> Unbounded channels can cause memory issues. When producer outpaces consumer, the memory grows and grows until OOM.
> 
> I'm not sure how possible it is in this case, but I will let you judge it :)

lib/executor/src/executors/websocket_client.rs
```diff
@@ -0,0 +1,480 @@
...
+    /// Multiple subscriptions can be active simultaneously on the same connection.
+    pub async fn subscribe(
+        &mut self,
+        subscribe_payload: SubscribePayload,
+    ) -> LocalBoxStream<'static, SubgraphResponse<'static>> {
+        let subscribe_id = self.next_subscription_id();
+
+        let _ = self
+            .sink
+            .send(ClientMessage::subscribe(
```
kamilkisiela, line 281: @enisdenjo 
Can there be a race condition here? We send something and only then `self.state.borrow_mut().subscriptions.insert()` happens.

The `dispatch_loop` already runs, I think, and `handle_text_frame` ignores messages from unknown subscriptions (id may not be registered yet).

enisdenjo: the WsClient is singlethreaded so immediately after the message sent await we populate the subscriptions map - so _technically_ this is safe. but, it's not obivious enough to notice as soon as you look at the code, so I changed the order. good catch

https://github.com/graphql-hive/router/pull/620/commits/bd178821e37b9a8e1316cbb9c511f399a69596ee

lib/executor/src/executors/http_callback.rs
enisdenjo: https://github.com/graphql-hive/router/pull/620/commits/b4ea23158161f09757110bbc2bbe6d4a8bab5092

lib/executor/src/executors/websocket.rs
enisdenjo: https://github.com/graphql-hive/router/pull/620/commits/b4ea23158161f09757110bbc2bbe6d4a8bab5092

lib/executor/src/executors/http.rs
enisdenjo: agreed. https://github.com/graphql-hive/router/pull/620/commits/82ab610184bbc3dcd4228f734975cc02b565583c

dotansimha (approved):

PR #623, opened 2025-12-15
feat(router): Query Complexity: Max Depth, Max Directives, Max Tokens

ardatan:
# New Query Complexity Configuration 

This PR adds the following configurable validation rules to prevent overly complex queries to be sent

## Max Depth

This is the validation rule to calculate the depth of the operation, and compare it the limits, then throws if it exceeds

```yaml
limits:
  max_depth:
    n: 10  # Set the maximum allowed depth for queries
```

Inspired from https://escape.tech/graphql-armor/docs/plugins/max-depth/

## Max Directives

New validation rule to limit the number of directives can be sent inside a GraphQL operation

```yaml
limits:
  max_directives:
    n: 10  # Set the maximum allowed depth for queries
```

Inspired from https://escape.tech/graphql-armor/docs/plugins/max-directives/

## Max Tokens

With a few changes in parser, it is possible to limit the number tokens parsed

```yaml
limits:
  max_tokens:
    n: 10
```

Inspired from https://escape.tech/graphql-armor/docs/plugins/max-tokens/

Documentation is WIP
E2E tests maybe?

### Related changes
In order to cover this -> https://github.com/graphql-hive/router/pull/623#discussion_r2703983088
I had to refactor `ParseError`. So it now contains the original internal error object from the underlying `consume` library. So that we can detect if the parse error is token limit exceeding error. Then we can use `ValidationError` instead of ParsingError from PipelineError.

### DIFF
--- a/bench/subgraphs/accounts.rs
+++ b/bench/subgraphs/accounts.rs
@@ -1,5 +1,5 @@
 use async_graphql::{
-    ComplexObject, EmptyMutation, EmptySubscription, ID, Interface, Object, Schema, SimpleObject
+    ComplexObject, EmptyMutation, EmptySubscription, Interface, Object, Schema, SimpleObject, ID,
 };
 use lazy_static::lazy_static;
 
--- a/e2e/src/max_depth.rs
+++ b/e2e/src/max_depth.rs
@@ -4,9 +4,9 @@ mod max_depth_e2e_tests {
     use sonic_rs::{from_slice, to_string_pretty, Value};
 
     use crate::testkit::{
-        SubgraphsServer, init_graphql_request, init_router_from_config_inline, wait_for_readiness
+        init_graphql_request, init_router_from_config_inline, wait_for_readiness, SubgraphsServer,
     };
-        const QUERY: &'static str = r#"
+    const QUERY: &'static str = r#"
             query {
                 me {
                     name
@@ -20,18 +20,17 @@ mod max_depth_e2e_tests {
     #[ntex::test]
     async fn allows_query_within_max_depth() {
         let _subgraphs = SubgraphsServer::start().await;
-        let app = init_router_from_config_inline(r#"
+        let app = init_router_from_config_inline(
+            r#"
         limits:
             max_depth:
                 n: 3
-        "#)
-            .await
-            .unwrap();
+        "#,
+        )
+        .await
+        .unwrap();
         wait_for_readiness(&app.app).await;
-        let req = init_graphql_request(
-            QUERY,
-            None,
-        );
+        let req = init_graphql_request(QUERY, None);
 
         let resp = test::call_service(&app.app, req.to_request()).await;
         let body = test::read_body(resp).await;
@@ -58,19 +57,18 @@ mod max_depth_e2e_tests {
     #[ntex::test]
     async fn rejects_query_exceeding_max_depth() {
         let _subgraphs = SubgraphsServer::start().await;
-        let app = init_router_from_config_inline(r#"
+        let app = init_router_from_config_inline(
+            r#"
         limits:
             max_depth:
                 n: 1
-        "#)
-            .await
-            .unwrap();
+        "#,
+        )
+        .await
+        .unwrap();
         wait_for_readiness(&app.app).await;
 
-        let req = init_graphql_request(
-            QUERY,
-            None,
-        );
+        let req = init_graphql_request(QUERY, None);
         let resp = test::call_service(&app.app, req.to_request()).await;
         let body = test::read_body(resp).await;
         let json_body: Value = from_slice(&body).unwrap();
@@ -91,13 +89,15 @@ mod max_depth_e2e_tests {
     #[ntex::test]
     async fn unknown_fragments() {
         let _subgraphs = SubgraphsServer::start().await;
-        let app = init_router_from_config_inline(r#"
+        let app = init_router_from_config_inline(
+            r#"
         limits:
             max_depth:
                 n: 3
-        "#)
-            .await
-            .unwrap();
+        "#,
+        )
+        .await
+        .unwrap();
         wait_for_readiness(&app.app).await;
         let req = init_graphql_request(
             r#"
--- a/e2e/src/max_directives.rs
+++ b/e2e/src/max_directives.rs
@@ -4,19 +4,21 @@ mod max_directives_e2e_tests {
     use sonic_rs::{from_slice, to_string_pretty, Value};
 
     use crate::testkit::{
-        SubgraphsServer, init_graphql_request, init_router_from_config_inline, wait_for_readiness
+        init_graphql_request, init_router_from_config_inline, wait_for_readiness, SubgraphsServer,
     };
 
     #[ntex::test]
     async fn allows_query_within_max_directives() {
         let _subgraphs = SubgraphsServer::start().await;
-        let app = init_router_from_config_inline(r#"
+        let app = init_router_from_config_inline(
+            r#"
         limits:
             max_directives:
                 n: 8
-        "#)
-            .await
-            .unwrap();
+        "#,
+        )
+        .await
+        .unwrap();
         wait_for_readiness(&app.app).await;
         let req = init_graphql_request(
             "query { 
@@ -60,13 +62,15 @@ mod max_directives_e2e_tests {
     #[ntex::test]
     async fn rejects_query_exceeding_max_directives() {
         let _subgraphs = SubgraphsServer::start().await;
-        let app = init_router_from_config_inline(r#"
+        let app = init_router_from_config_inline(
+            r#"
         limits:
             max_directives:
                 n: 5
-        "#)
-            .await
-            .unwrap();
+        "#,
+        )
+        .await
+        .unwrap();
         wait_for_readiness(&app.app).await;
 
         let req = init_graphql_request(


kamilkisiela:
After #647 gets in, we can do this parser thing where we track parsed tokens and put the limit

kamilkisiela:
We could do
```yaml
limits:
  max_depth: ...
```

instead of `query_complexity`, as we may want to also include limits related to http requests (body size, max amount of headers etc) in future.

bin/router/src/pipeline/validation/introspection_query.graphql
```diff
@@ -0,0 +1,107 @@
+
+    query IntrospectionQuery {
```
kamilkisiela: Let's make it more obvious that this file is used only in tests. I was confused why it exists and how it relates to the new validation rules, I thought you touched introspection in Router or something.

bin/router/src/lib.rs
```diff
@@ -126,10 +130,23 @@ pub async fn configure_app_from_config(
     let schema_state =
         SchemaState::new_from_config(bg_tasks_manager, router_config_arc.clone()).await?;
     let schema_state_arc = Arc::new(schema_state);
+    let mut validation_plan = default_rules_validation_plan();
+    if let Some(max_depth_config) = &router_config_arc.limits.max_depth {
+        if max_depth_config.is_enabled() {
+            info!(
```
kamilkisiela: Not sure we should log those, for sure not as `info` level, `debug` maybe

lib/router-config/src/limits.rs
```diff
@@ -0,0 +1,123 @@
...
+            n: default_max_depth(),
+            ignore_introspection: default_ignore_introspection(),
+            flatten_fragments: default_flatten_fragments(),
+            expose_limits: default_expose_limits(),
+        }
+    }
+}
+
+fn default_max_depth() -> usize {
+    6
```
kamilkisiela: I'm not sure we should offer default values, I believe it's better to leave it to the user to configure. We make the user think about the value, what it represents etc.

Having a default can be surprising to many, and we're simply guessing here.

@dotansimha wdyt?

kamilkisiela: Same applies to all `n`s

bin/router/src/pipeline/validation/shared.rs
```diff
@@ -0,0 +1,53 @@
+use graphql_tools::{
+    ast::OperationDefinitionExtension,
+    static_graphql::query::{
+        Field, FragmentDefinition, FragmentSpread, InlineFragment, OperationDefinition, Selection,
+        SelectionSet,
+    },
+};
+
+pub enum CountableNode<'a> {
```
kamilkisiela, line 23: The name doesn't tell what "countable" really means. Looking at usage, these are nodes in a GraphQL query we traverse and analyze. With one exception being the fragment spread (that has no explanation of why it's ignored).

bin/router/src/pipeline/validation/max_directives_rule.rs
```diff
@@ -0,0 +1,284 @@
...
+                OperationDefinition::SelectionSet(_) => None,
+            },
+            CountableNode::FragmentSpread(fragment_spread) => Some(&fragment_spread.directives),
+        }
+    }
+}
+
+impl<'a> MaxDirectivesVisitor<'a, '_> {
+    fn count_directives(&mut self, countable_node: CountableNode<'a>) -> i32 {
+        let mut directive_cnt: i32 = 0;
```
kamilkisiela: Why not just `directive_count`?

bin/router/src/pipeline/validation/max_depth_rule.rs
```diff
@@ -0,0 +1,537 @@
...
+            if !self.config.flatten_fragments {
+                parent_depth += 1;
+            }
+
+            let fragment_name = node.fragment_name.as_str();
+            let visited_fragment = self.visited_fragments.get(fragment_name);
+            if let Some(visited_fragment_depth) = visited_fragment {
+                return parent_depth + visited_fragment_depth;
+            } else {
+                self.visited_fragments.insert(fragment_name, -1);
```
kamilkisiela: why `-1`?

bin/router/src/pipeline/validation/max_directives_rule.rs
```diff
@@ -0,0 +1,284 @@
...
+                let mut visitor = MaxDirectivesVisitor {
+                    visited_fragments: HashMap::new(),
+                    ctx,
+                };
+                let directives = visitor.count_directives(op.into());
+                if directives > self.config.n as i32 {
+                    let message = if self.config.expose_limits {
+                        format!(
+                            "Directives limit of {} exceeded, found {}",
+                            directives, self.config.n
```
kamilkisiela: I think it should be swapped. FIrst config then directives.

bin/router/src/pipeline/validation/max_directives_rule.rs
```diff
@@ -0,0 +1,284 @@
...
+            config: MaxDirectivesRuleConfig {
+                n: 3,
+                expose_limits: true,
+            },
+        })]);
+
+        let errors = validate(&schema, &query, &validation_plan);
+
+        assert_eq!(errors.len(), 1);
+        assert_eq!(errors[0].message, "Directives limit of 4 exceeded, found 3");
```
kamilkisiela: Yeah, they need to be swapped and updated

lib/router-config/src/limits.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub flatten_fragments: bool,
+
+    #[serde(default = "default_expose_limits")]
+    /// Whether to expose the limits in the error message.
+    pub expose_limits: bool,
+}
+
+impl MaxDepthRuleConfig {
+    pub fn is_enabled(&self) -> bool {
+        self.n > 0
```
kamilkisiela: I think `enabled: boolean` is a better choice, but I don't think we should have an option to enable/disable those limits, at least not yet. My guess is that people will ask for it, but not for a boolean or `n>0`, they'll probably want to apply limits on certain apps and disable the limits for trusted queries/apps (meaning expression or a higher-level feature).

kamilkisiela: Another issue here is that we don't verify the `n`.
if it's `-123` then the rule is disabled now, but imo it should error, as it could be a typo.

bin/router/src/pipeline/validation/max_directives_rule.rs
```diff
@@ -0,0 +1,284 @@
...
+    books: [Book]
+  }
+"#;
+
+    const QUERY: &'static str = r#"
+  query {
+    __typename @a @a @a @a
+  }
+"#;
+
```
kamilkisiela, line 147:
```suggestion
#[test]
    fn bug_i_think() {
        let schema = parse_schema(TYPE_DEFS)
            .expect("Failed to parse schema")
            .into_static();
        let query = parse_query(
            r#"
        query {
          ...A
        }

        fragment A on Query {
          ...B @directive1 @directive2
        }

        fragment B on Query {
          ...A @directive3 @directive4
        }
      "#,
        )
        .expect("Failed to parse query")
        .into_static();

        let validation_plan = ValidationPlan::from(vec![Box::new(MaxDirectivesRule {
            config: MaxDirectivesRuleConfig {
                n: 1,
                expose_limits: false,
            },
        })]);

        let errors = validate(&schema, &query, &validation_plan);

        assert!(!errors.is_empty(), "There are 4 directives, should error");
    }
```

lib/router-config/src/limits.rs
ardatan: It is an unsigned integer, it can't be negative.

kamilkisiela: That's correct, yeah and this is the error user will get:
```
Failed to start Hive Router:
  Failed to load configuration: invalid type: 64-bit integer `-2`, expected an unsigned 64 bit or less integer for key `limits.max_depth.n` in router.config.yaml
  ```
  
  I raised two points here, negative value is one of them.

File: bin/router/src/pipeline/validation/shared.rs (hunk 4 above)
Comment (ardatan) [reply bin/router/src/pipeline/validation/shared.rs:23]: Fragment spreads are not exception, it was a bug I fixed. I also added some comments to describe why we need this enum.

File: bin/router/src/pipeline/validation/max_directives_rule.rs (hunk 5 above)
Comment (ardatan) [reply bin/router/src/pipeline/validation/max_directives_rule.rs]: Fixed

File: bin/router/src/pipeline/validation/max_depth_rule.rs (hunk 6 above)
Comment (ardatan) [reply bin/router/src/pipeline/validation/max_depth_rule.rs]: The idea is to avoid infinity recursions in case of recursive fragment spreads like you also shared.

File: bin/router/src/pipeline/validation/max_directives_rule.rs (hunk 7 above)
Comment (ardatan) [reply bin/router/src/pipeline/validation/max_directives_rule.rs]: Fixed!

File: bin/router/src/pipeline/validation/max_directives_rule.rs (hunk 8 above)
Comment (ardatan) [reply bin/router/src/pipeline/validation/max_directives_rule.rs]: Fixed!

File: lib/router-config/src/limits.rs (hunk 9 above)
Comment (ardatan) [reply lib/router-config/src/limits.rs]: I also added `enabled`. Forgot to tell.

File: bin/router/src/pipeline/validation/max_directives_rule.rs (hunk 10 above)
Comment (ardatan) [reply bin/router/src/pipeline/validation/max_directives_rule.rs:147]: Fixed!

File: lib/router-config/src/limits.rs (hunk 9 above)
Comment (kamilkisiela) [reply lib/router-config/src/limits.rs]: hmm but should we have `enabled` option for those? I said that `enabled` is a better choice, but imo it's better to have `Option<T>` so if there are no settings for max_depth, then it's disabled, and they are, they are explicitly defined by the user (then the user understands or at least have a chance to understand what the values mean).

File: bin/router/src/lib.rs (hunk 2 above)
Comment (ardatan) [reply bin/router/src/lib.rs]: That's an extra. I might have forgot it after debugging. Removed!

File: bin/router/src/pipeline/validation/introspection_query.graphql (hunk 1 above)
Comment (ardatan) [reply bin/router/src/pipeline/validation/introspection_query.graphql]: I created a `test_fixtures` directory and put it there with the name `introspection_query.fixture.graphql`.
Looks better?

File: lib/router-config/src/limits.rs (hunk 9 above)
Comment (ardatan) [reply lib/router-config/src/limits.rs]: They were already `Option<MaxTokensRuleConfig>` etc. I didn't understand what I should change.
On the other hand, do you think I should add expression support now?

File: bin/router/src/pipeline/validation/shared.rs (hunk 4 above)
Comment (kamilkisiela) [reply bin/router/src/pipeline/validation/shared.rs:23]: Fragment spreads are an exception, as they are the only variant in `selection_set` that is not counted, and it's still missing the description why.

If it's only to count directives, then why not make the struct name specific to directives then? 

If it's meant to be reused, then why fragment spread is skipped.

Comment (ardatan) [reply bin/router/src/pipeline/validation/shared.rs:23]: The struct is not only for directives. It is used in max depth too. For directive counting, spreads are not meant to be skipped. There was a bug you found out :) I fixed it and added the test you shared.

Comment (ardatan) [reply bin/router/src/pipeline/validation/shared.rs:23]: I added some comments to the code, and also moved `get_directives` method to the original `impl`  to make its purpose clear

File: bin/router/src/pipeline/validation/max_depth_rule.rs (hunk 6 above)
Comment (ardatan) [reply bin/router/src/pipeline/validation/max_depth_rule.rs]: I replaced it with `enum`s now there will be `VisitedFragment::Visiting` or `VisitedFragment::Counted(n)`.
So first one will be set initially to avoid infinite loops while the latter will be set whenever the count calculation is finished.

File: .vscode/settings.json [hunk 11]
```diff
@@ -0,0 +1,3 @@
+{
+    "cmake.sourceDirectory": "/home/ardatan/Guild/query-planner-rs/node_modules/emnapi"
```
kamilkisiela: ⚔️

lib/router-config/src/limits.rs
kamilkisiela: Waiting for @dotansimha to chip in.

If he says "I agree", then let's drop those `default` functions

.vscode/settings.json
ardatan: Oh VSCode... Deleted!

lib/router-config/src/limits.rs
dotansimha: Hmm, I think this needs to be explicitly enabled, as it adds functionality on top of what's built-in with GraphQL. 

So for the question above, I think all of the graphql-security-features needs to be enabled explicitly by the user, so we won't surprise the user in production.

dotansimha: fyi @ardatan

lib/router-config/src/limits.rs
dotansimha: As mentioned above, let's not enable by default. 

If the user doesn't explicitly configured `n: N`, then the plugin is disabled. If they do, it it's enabled. 

Later we might consider to add `enabled` as standalone, but i feel like here it might not be needed at the moment? since the user needs an explicit action in order to activate this.

ardatan: Ok now `n: N` should be explicitly defined in order to enable these. Otherwise, they are all disabled by default.

lib/router-config/src/limits.rs
ardatan: https://github.com/graphql-hive/router/pull/623#discussion_r2695110269

lib/router-config/src/limits.rs
dotansimha: perfect

bin/router/src/pipeline/validation/max_depth_rule.rs
```diff
@@ -0,0 +1,567 @@
...
+            let Definition::Operation(op) = definition else {
+                continue;
+            };
+
+            let mut visitor = MaxDepthVisitor {
+                config: &self.config,
+                visited_fragments: HashMap::new(),
+                ctx,
+            };
+            let depth = visitor.count_depth(op.into(), None);
```
dotansimha: I think the logic here has a tiny flaw, a performance issue, and a possible security concern, due to the fact you are using recursion to calculate the depth. 

You are recursing `count_depth` and only checking the result when you are done, exposing the router to a potential ddos in case of a super deep query sent in order to exhaust the validation rule / Rust stack. 

A simple solution on thre application layer, would be to give up on the need to calcuate the total depth, and fail early if the depth limit is reached. This also means that you don't need to return `found N` in the error message.

bin/router/src/pipeline/validation/max_depth_rule.rs
```diff
@@ -0,0 +1,567 @@
...
+        &self,
+        ctx: &mut OperationVisitorContext<'_>,
+        error_collector: &mut ValidationErrorContext,
+    ) {
+        for definition in &ctx.operation.definitions {
+            let Definition::Operation(op) = definition else {
+                continue;
+            };
+
+            let mut visitor = MaxDepthVisitor {
```
dotansimha: Any reason for initialize this inside the `for` loop? I think it should be outside?

bin/router/src/pipeline/validation/max_depth_rule.rs
```diff
@@ -0,0 +1,567 @@
...
+            }
+
+            // If not, mark it as Visiting initially to avoid infinite loops,
+            // because fragments can refer itself recursively at some point.
+            // See the tests at the bottom of this file to understand the use cases fully.
+            self.visited_fragments
+                .insert(fragment_name, VisitedFragment::Visiting);
+
+            // Look up the fragment definition by its name
+            if let Some(fragment) = self.ctx.known_fragments.get(fragment_name) {
```
dotansimha, line 131: I assume that the end case of non-existing fragment is checked by another rule, right? 

Can we just double check and confirm that the other rule runs before? (https://github.com/graphql-hive/router/blob/main/lib/graphql-tools/src/validation/rules/known_fragment_names.rs  right?)

bin/router/src/pipeline/validation/max_depth_rule.rs
```diff
@@ -0,0 +1,567 @@
...
+            };
+
+            let mut visitor = MaxDepthVisitor {
+                config: &self.config,
+                visited_fragments: HashMap::new(),
+                ctx,
+            };
+            let depth = visitor.count_depth(op.into(), None);
+
+            if depth <= self.config.n {
```
dotansimha: If we do `<=`, let's make sure that the docs mentions that the limit is inclusive? so it will be clear that `n: 5` allows up to or 5.

bin/router/src/pipeline/validation/max_depth_rule.rs
```diff
@@ -0,0 +1,567 @@
...
+        let error = &errors[0];
+        assert_eq!(error.message, "Query depth limit of 1 exceeded, found 3.");
+    }
+
+    #[test]
+    fn rejects_fragment_exceeding_max_depth() {
+        let validation_plan = ValidationPlan::from(vec![Box::new(MaxDepthRule {
+            config: MaxDepthRuleConfig {
+                n: 4,
+                ..Default::default()
```
dotansimha: this fails now because this type doesn't have `Default` implemented

dotansimha: Image

bin/router/src/pipeline/validation/max_depth_rule.rs
```diff
@@ -0,0 +1,567 @@
...
+        error_collector: &mut ValidationErrorContext,
+    ) {
+        for definition in &ctx.operation.definitions {
+            let Definition::Operation(op) = definition else {
+                continue;
+            };
+
+            let mut visitor = MaxDepthVisitor {
+                config: &self.config,
+                visited_fragments: HashMap::new(),
```
dotansimha: Should this `HashMap` have a size limit (with_capacity)? I have in mind a scenario where a malicious query with tons of fragments is sent, which could be valid (doesn't have depth issues), but still iterate tons of times here and grow this hashmap? 

It could also be some other limit, or something that limits the fragments count iteration?

bin/router/src/pipeline/validation/max_directives_rule.rs
```diff
@@ -0,0 +1,322 @@
...
+        ctx: &mut OperationVisitorContext<'_>,
+        error_collector: &mut ValidationErrorContext,
+    ) {
+        for definition in &ctx.operation.definitions {
+            let Definition::Operation(op) = definition else {
+                continue;
+            };
+
+            let mut visitor = MaxDirectivesVisitor {
+                visited_fragments: HashMap::new(),
```
dotansimha: See my comment in depth limit rule, about the `HashMap` and possible need to configure some limit here.

bin/router/src/pipeline/validation/max_directives_rule.rs
```diff
@@ -0,0 +1,322 @@
...
+        &self,
+        ctx: &mut OperationVisitorContext<'_>,
+        error_collector: &mut ValidationErrorContext,
+    ) {
+        for definition in &ctx.operation.definitions {
+            let Definition::Operation(op) = definition else {
+                continue;
+            };
+
+            let mut visitor = MaxDirectivesVisitor {
```
dotansimha: Also, should `MaxDirectivesVisitor` be created outside the loop?

bin/router/src/pipeline/validation/max_directives_rule.rs
```diff
@@ -0,0 +1,322 @@
...
+    }
+}
+
+struct MaxDirectivesVisitor<'a, 'b> {
+    visited_fragments: HashMap<&'a str, VisitedFragment>,
+    ctx: &'b mut OperationVisitorContext<'a>,
+}
+
+impl<'a> MaxDirectivesVisitor<'a, '_> {
+    fn count_directives(&mut self, countable_node: CountableNode<'a>) -> usize {
```
dotansimha: Similar to depth limit, please check me comment about recursion and possible issues with counting it all and then returning. I think it might be better to cut early in case limit has reached. 

No need to count them all.

lib/graphql-tools/src/parser/tokenizer.rs
```diff
@@ -154,10 +169,26 @@ impl<'a> TokenStream<'a> {
...
+        // Should be counted?
+        if self.token_limit.is_some() {
+            self.token_count += 1;
+        }
         Ok((kind, size))
     }
 
     fn take_token(&mut self) -> Result<(Kind, usize), Error<Token<'a>, Token<'a>>> {
+        if let Some(limit) = self.token_limit {
+            if self.token_count >= limit {
```
dotansimha, line 178: This one is a good example on how cutting early can help us. The parser cut early in case token limit reached, before counting it all.

lib/router-config/src/limits.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// It is used to prevent too many directives that could lead to overfetching or DOS attacks.
+    #[serde(default)]
+    pub max_directives: Option<MaxDirectivesRuleConfig>,
+
+    /// Configuration of limiting the number of tokens in the incoming GraphQL operations.
+    /// If not specified, token limiting is disabled.
+    ///
+    /// It is used to prevent too large queries that could lead to overfetching or DOS attacks.
+    #[serde(default)]
+    pub max_tokens: Option<MaxTokensRuleConfig>,
```
dotansimha, line 25: in the docs, let's explain what's a token and give example on what's being count and what not.

.changeset/query_complexity.md
```diff
@@ -0,0 +1,50 @@
...
+
+```yaml
+limits:
+  max_directives:
+    n: 5  # Set the maximum allowed number of directives
+```
+
+This configuration helps to prevent excessive use of directives in queries, which can lead to performance issues.
+
+# Max Tokens
```
dotansimha: `##`

docs/README.md
```diff
@@ -90,6 +91,10 @@ jwt:
     - name: authorization
       prefix: Bearer
       source: header
+limits:
```
dotansimha: @ardatan I think something in the config causes it to be generated as `null`. Probably some serde attribute missing in order to avoid this?

dotansimha:
also, rebase + start a docs PR

bin/router/src/pipeline/validation/max_depth_rule.rs
ardatan: Limit means the maximum amount it allows, no? But we can still specify it in the docs.

docs/README.md
ardatan: By default they are all disabled. So they are null. Not sure if I understand the issue here?

bin/router/src/pipeline/validation/max_directives_rule.rs
ardatan: But it starts counting from an operation, not all definitions. But I'll move it anyway :)

bin/router/src/pipeline/validation/max_depth_rule.rs
ardatan: We add this rule after default validation rules. And also I am not sure if it really matters once the referenced fragment is missing. It will always prevent execution on validation.

bin/router/src/pipeline/validation/max_depth_rule.rs
ardatan: I did a refactor on it. So it checks on every change in the counter.

bin/router/src/pipeline/validation/max_depth_rule.rs
ardatan: Moved outside

bin/router/src/pipeline/validation/max_depth_rule.rs
ardatan: Updated

bin/router/src/pipeline/validation/max_directives_rule.rs
ardatan: Done

.changeset/query_complexity.md
ardatan: Done

docs/README.md
dotansimha: I think you can add `#[serde(skip_serializing_if = "Option::is_none")]` to the fields in `LimitsConfig`, then it will not print the `null` in the config

ardatan: Fixed

bin/router/src/pipeline/validation/max_depth_rule.rs
```diff
@@ -35,27 +35,16 @@ impl ValidationRule for MaxDepthRule {
                 config: &self.config,
                 visited_fragments: HashMap::new(),
                 ctx,
+                limit_checker: LimitChecker {
+                    limit: self.config.n,
+                    limit_name: "Query depth",
+                    expose_limits: self.config.expose_limits,
```
dotansimha: Does `LimitChecker` really needs `limit_name` and `expose_limits` and `error_code` ? 
I mean, the whole purpose of it is to track some `u64` and return some error when it exceeds. I feel like the logic of creating the final error string and returning it is not something that should be done inside `LimitChecker`.

ardatan: This struct doesn't track anything. It just checks if the sent value is more than `limit` or not. If not, it returns an error by generating it using `error_code`, `limit_name` and `expose_limits`.

dotansimha: I feel like it's a bit weird that we need to pass a `limit_name` and more stuff to an external struct, just for formatting a message or building the error. 
Building the error is a clear role of the `ValidationRule`, while `LimitChecker` needs to check the limit (is it just a condition? so maybe we don't need a while struct for that?) 

Even if it's a few extra line of code, I don't see this as an optimization or an improvement to the code.

bin/router/src/pipeline/validation/max_depth_rule.rs
ardatan: I'll remove the found one then

bin/router/src/pipeline/validation/max_depth_rule.rs
ardatan: Removed it

dotansimha: Thanks!

bin/router/src/pipeline/validation/max_depth_rule.rs
dotansimha: great, thanks. Thinking a bit more about this, I even think we might not need to expose the limit? What do you think?

ardatan: Removed

e2e/src/max_tokens.rs
```diff
@@ -0,0 +1,52 @@
...
+            limits:
+                max_tokens:
+                    n: 100
+            "#,
+        )
+        .await
+        .unwrap();
+        wait_for_readiness(&app.app).await;
+
+        let req = init_graphql_request("{ a a a a a a a }", None);
```
dotansimha: this one passes on the tokens limit, but really fails on a validation issue `Cannot query field \"a\" on type \"Query\".`. Can you please add a test that fully passes?

My main reason for asking, is because we can't tell if validation failed because of incorrect fields, or because of tokens limit, and if order of the validation matters, it might impact things.

e2e/src/max_tokens.rs
```diff
@@ -0,0 +1,52 @@
...
+        )
+        .await
+        .unwrap();
+        wait_for_readiness(&app.app).await;
+
+        let req = init_graphql_request("query { a a a a a a }", None);
+        let resp = test::call_service(&app.app, req.to_request()).await;
+        let body_bytes = test::read_body(resp).await;
+        let body_str = std::str::from_utf8(&body_bytes).unwrap();
+        assert!(body_str.contains("Token limit exceeded"));
```
dotansimha: Can you please confirm the full response body here?

e2e/src/max_tokens.rs
```diff
@@ -0,0 +1,52 @@
...
+            "#,
+        )
+        .await
+        .unwrap();
+        wait_for_readiness(&app.app).await;
+
+        let req = init_graphql_request("query { a a a a a a }", None);
+        let resp = test::call_service(&app.app, req.to_request()).await;
+        let body_bytes = test::read_body(resp).await;
+        let body_str = std::str::from_utf8(&body_bytes).unwrap();
```
dotansimha: Hmm, this one fails with: 
```
{"errors":[{"message":"Failed to parse GraphQL operation: query parse error: Parse error at 1:13\nExpected }\nToken limit exceeded\n","extensions":{"code":"GRAPHQL_PARSE_FAILED"}}]}
```

While this is valid to fail, the `code` is about a parsing error - should this be a validation error instead? (token limit is indeed implemented in the parser, but i wonder if it should be considered a validation issue?)

ardatan: Fixed!

e2e/src/max_tokens.rs
ardatan: Fixed!

e2e/src/max_tokens.rs
ardatan: Fixed!

e2e/src/max_tokens.rs
ardatan: Also added e2es @dotansimha

dotansimha (approved):

PR #624, opened 2025-12-16
perf(query-planner): optimize InDegree with Vec-based storage

rhzs:
- Replaces `HashMap` with a boxed slice using `in_degree+1` encoding, enabling calloc-style zero-initialization.
- Converts panics to Result-based error handling.
- Avoids `String` allocation in `BestPathTracker` by using borrowed `&str` references.

### DIFF
--- a/lib/query-planner/src/planner/query_plan.rs
+++ b/lib/query-planner/src/planner/query_plan.rs
@@ -254,7 +254,6 @@ fn are_conditions_compatible(c1: &ConditionNode, c2: &ConditionNode) -> bool {
 }
 
 fn merge_two_condition_nodes(mut a: ConditionNode, mut b: ConditionNode) -> PlanNode {
-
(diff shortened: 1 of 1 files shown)


dotansimha (commented):
Thank you for your contribution @rhzs ! Can you please share the background for this PR? Did you encounter some performance issues? 

> Replaces HashMap with a boxed slice using in_degree+1 encoding, enabling calloc-style zero-initialization.

While this PR seems appealing, I need your help with a few things to clear on this. 
I do understand the benefit of using zero allocations, but I'm not sure if the benefit is so significant in that area of the code, especially because it adds some additional complexity. 

Can you please elaborate on how you tested the performance improvement? It will help me to understand if the added complexity is worth it. 

> Converts panics to Result-based error handling.

✅ 

> Avoids String allocation in BestPathTracker by using borrowed &str references.

✅

PR #625, opened 2025-12-16
fix(jwt): make JWKS algorithm optional and fallback to token alg if set

### LINKED ISSUE #621: JWT validation for Entra tokens breaks at JWK check
Hi!

It seems the code for validating tokens from Entra expects the JWK from Entra to contain the algorithm. Microsoft does not publish this field and I think it is optional in the spec.

I have tried to configure Entra auth for Hive 0.0.23 for both versions of tokens Entra publishes, I've also tried them separately. They always fail with `403`:

error
```json
{
  "errors": [
    {
      "message": "all jwk failed to decode token: [JwkMissingAlgorithm, JwkMissingAlgorithm, JwkMissingAlgorithm, JwkMissingAlgorithm, JwkMissingAlgorithm]",
      "extensions": {
        "code": "MISSING_JWT"
      }
    }
  ]
}
```

router.config.yaml
```yaml
jwt:
  enabled: true
  require_authentication: true
  allowed_algorithms:
    - RS256
  jwks_providers:
    - source: remote
      url: "https://login.microsoftonline.com/<tenant id>/discovery/v2.0/keys"
      polling_interval: "1h"
      prefetch: true
    - source: remote
      url: "https://login.microsoftonline.com/<tenant id>/discovery/keys"
      polling_interval: "1h"
      prefetch: true
  issuers:
    - "https://login.microsoftonline.com/<tenant id>/v2.0"
    - "https://sts.windows.net/<tenant id>/"
  audiences:
    - "api://<client id>"
    - "<client id>"
```

I'm not familiar with rust but this piece of code seems to expect the algorithm is defined in the JWK:

```rust
fn try_decode_from_jwk(&self, token: &str, jwk: &Jwk) -> Result<JwtTokenPayload, JwtError> {
    let decoding_key = DecodingKey::from_jwk(jwk).map_err(JwtError::InvalidDecodingKey)?;
    let key_alg = jwk
        .common
        .key_algorithm
        .ok_or(JwtError::JwkMissingAlgorithm)?;
```

Since the field is optional, perhaps a fallback could be to fetch the algorithm from the JWT header where it is defined. A quick visit to Google shows many libraries have run into this issue through the years.

I've added a local file instead where I added `alg` manually to the jwks as a workaround, but keeping a local file updated will be a pain. Let me know if you need more information. I'm attaching the jwks for both Entra endpoints so you can see the details.

[v1.jwk.json](https://github.com/user-attachments/files/24146445/v1.jwk.json)
[v2.jwk.json](https://github.com/user-attachments/files/24146444/v2.jwk.json)
RFC specifying OPTIONAL (https://datatracker.ietf.org/doc/html/rfc7517#section-4.4)

dkmio:
This PR makes the JWK algorithm optional as it is defined as such in the RFC. To handle a missing algorithm, we fall back to reading the algorithm from the user JWT. To protect against forged tokens, we add a validation that the algorithm in the token is part of the `allowed_algorithms`. Since `JwkMissingAlgorithm` is not longer an error, the field is removed.

Build have been tested locally but not with a docker image since I could not get the docker build to work properly.

Fixes #621

### DIFF
--- a/bin/router/src/jwt/mod.rs
+++ b/bin/router/src/jwt/mod.rs
@@ -169,7 +169,7 @@ impl JwtAuthRuntime {
                 let header = decode_header(&token).map_err(JwtError::InvalidJwtHeader)?;
                 let jwk = self.find_matching_jwks(&header, jwks)?;
 
-                self.decode_and_validate_token(&token, &jwk.keys)
+                self.decode_and_validate_token(&header, &token, &jwk.keys)
                     .map(|token_data| (token_data, maybe_prefix, token))
             }
             Err(e) => {
@@ -182,10 +182,13 @@ impl JwtAuthRuntime {
 
     fn decode_and_validate_token(
         &self,
+        header: &Header,
         token: &str,
         jwks: &[Jwk],
     ) -> Result<JwtTokenPayload, JwtError> {
-        let decode_attempts = jwks.iter().map(|jwk| self.try_decode_from_jwk(token, jwk));
+        let decode_attempts = jwks
+            .iter()
+            .map(|jwk| self.try_decode_from_jwk(header, token, jwk));
 
         if let Some(success) = decode_attempts.clone().find(|result| result.is_ok()) {
             return success;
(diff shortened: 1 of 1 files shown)


dotansimha:
This makes perfect sense. Thanks for fixing @dkmio ! I'm taking this one to the finish line and will cut a release soon.

dotansimha:
Ok I've adjust the code a bit, just to avoid calling `decode_header` twice. Rest of the code remains the same. 

Thanks @dkmio !

PR #628, opened 2025-12-19
feat(router): plugin system

ardatan:
Documentation > https://github.com/graphql-hive/console/pull/7387
Preview of the docs -> https://pr-7387.hive-landing-page.pages.dev/docs/router/guides/extending-the-router
Example plugins -> https://github.com/graphql-hive/router/tree/plugin-system/plugin_examples

TODO:
- Document why API is implemented in this way and that way (payload.initialize_plugins, OnPluginInitResult type e.g.)
- Documentation updates


e2e/src/plugins/one_of.rs
```diff
@@ -0,0 +1,336 @@
...
+    fn on_supergraph_reload<'exec>(
+        &'exec self,
+        start_payload: OnSupergraphLoadStartHookPayload,
+    ) -> StartHookResult<'exec, OnSupergraphLoadStartHookPayload, OnSupergraphLoadEndHookPayload>
+    {
+        for def in start_payload.new_ast.definitions.iter() {
+            if let Definition::TypeDefinition(TypeDefinition::InputObject(input_obj)) = def {
+                for directive in input_obj.directives.iter() {
+                    if directive.name == "oneOf" {
+                        self.one_of_types
```
dotansimha: This one is tricky. 
The whole supergraph swap is managed using `ArcSwap` in order to have a way to swap the supergraph without impacting in-flight requests.
So a request will get the supergraph when it starts, and it guarentee that it won't be replaced or changed during the request handling. 

Since here we want a similar behaviour, we want to have a similar setup with `ArcSwap` and not use `RwLock`, this way the state of the plugin guarentee to remain the same during the time of the request, and not to change if a supergraph changes mid-flight.

ardatan: Does this look better?
https://github.com/graphql-hive/router/pull/628/commits/a37fcd36f0b8ef2bc32f770a2a6b9ba0d38265ba

bin/router/src/pipeline/mod.rs
```diff
@@ -169,18 +207,91 @@ pub async fn execute_pipeline(
...
+
+    validate_operation_with_cache(
         supergraph,
         schema_state,
-        &execution_request,
+        shared_state,
         &parser_payload,
+        &plugin_req_state,
     )
     .await?;
```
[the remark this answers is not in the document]
ardatan: Fixed

lib/router-config/src/lib.rs
```diff
@@ -111,6 +111,10 @@ pub struct HiveRouterConfig {
     /// Configuration to enable or disable introspection queries.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub introspection: Option<IntrospectionPermissionConfig>,
+
+    /// Configuration for custom plugins
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub plugins: HashMap<String, serde_json::Value>,
```
dotansimha: Should we allow any `serde_json::Value` at the root level, or enforce something like an object? 

I mean, this is considered valid, or do we to push users to atlesat have an object at the root level of each plugin?

```yaml
plugins:
  my_plugin: true
```

```yaml
plugins:
  my_plugin:
    my_config: true
```

bin/router/src/lib.rs
```diff
@@ -5,6 +5,7 @@ mod http_utils;
 mod jwt;
 mod logger;
 pub mod pipeline;
+pub mod plugins;
```
dotansimha, line 6: @ardatan please check my comment here regarding simplicity of use and what needs to be exported:
https://github.com/graphql-hive/console/pull/7387#discussion_r2735172510

dotansimha: similar to what we did with `init_rustls_crypto_provider`

lib/router-config/src/lib.rs
```diff
@@ -111,6 +111,10 @@ pub struct HiveRouterConfig {
     /// Configuration to enable or disable introspection queries.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub introspection: Option<IntrospectionPermissionConfig>,
+
+    /// Configuration for custom plugins
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
```
dotansimha, line 117: At the moment, the configuration entrypoint is directly passed to the plugin, and I wanted to suggest the following: 

```
plugins: 
  my_plugin:
     enabled: boolean/expression # this is not part of the plugin, but something we manage 
     config: {} # this is a custom Value or Object we allow the user to have here 
```

This is also related to another comment I made on docs repo, that explains that a plugin that was registered but doesn't have a configuration, might have weird way setup in order to enable it. 

Also, we don't want users to need to implement enabled/disable for their plugin, I think this is a fairly simple and explicit way to tell users to enable plugins once registered.

lib/router-config/src/lib.rs
dotansimha: I'm not sure about `HashMap` here. Should this be a `Vec` so users will configure it as an array with an explicit order? 

Also, what about a scenario where plugin could be added multiple times? is that something we consider or out of scope? (and then `Vec` solves it)

lib/internal/src/authorization/metadata.rs
```diff
@@ -0,0 +1,47 @@
+use ahash::HashMap;
```
dotansimha, line 1: why this one has been moved?

lib/executor/src/plugins/plugin_trait.rs
```diff
@@ -0,0 +1,322 @@
...
+        &'exec self,
+        start_payload: OnSupergraphLoadStartHookPayload,
+    ) -> OnSupergraphLoadStartHookResult<'exec> {
+        start_payload.proceed()
+    }
+    async fn on_shutdown<'exec>(&'exec self) {}
+}
+
+#[async_trait::async_trait]
+pub trait DynRouterPlugin: Send + Sync + 'static {
```
dotansimha, line 233: I think this and `RouterPluginBoxed` and not intended to be used by a user, right? they are both exported and public, so please change.

lib/executor/src/plugins/plugin_trait.rs
```diff
@@ -0,0 +1,322 @@
...
+            .status(status_code)
+            .body(body)
+    }
+}
+
+#[async_trait::async_trait]
+pub trait RouterPlugin: Send + Sync + 'static {
+    fn plugin_name() -> &'static str;
+
+    type Config: DeserializeOwned + Sync;
```
dotansimha, line 155: Just so it's clear, if a plugin doesn't have config, then it needs to use `Config = ()`, right?

lib/executor/src/plugins/plugin_trait.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    fn from_config(config: Self::Config) -> Result<Option<Self>, BoxError>
```
dotansimha: why `Result<Option<>>` here? if we add `enabled:` to every plugin, then this makes no sense and needs to be just `Result<Self>`

bin/router/src/main.rs
```diff
@@ -7,7 +7,7 @@ static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
 async fn main() -> Result<(), RouterInitError> {
     init_rustls_crypto_provider();
 
-    match router_entrypoint().await {
+    match router_entrypoint(None).await {
```
dotansimha: Also on this, do you think it should just accept a plugin registey instance? it has 0 cost really to init an empty one and drop the `Option` here

lib/executor/src/plugins/plugin_trait.rs
dotansimha: Let's discuss `from_config`; in short: it has multiple purposes that needs to be split (configure/construct the plugin, and initialize the plugin). 

This is a helper function that creates the instance of the plugin, from a given config. But, in all examples and also in practice, you are using function to do more. Let's say you need to load a file, or load something from remote service, or even start a thread - is it the right place? 

My suggestion is to create some sort of a startup hook that will allow users to configure, create the in-memory struct representation, and later to do initialize other things that might be needed (`on_router_startup` maybe?).

lib/executor/src/plugins/plugin_context.rs
```diff
@@ -0,0 +1,159 @@
...
+            .map(|entry| PluginContextMutEntry {
+                entry,
+                phantom: std::marker::PhantomData,
+            })
+    }
+}
+
+pub struct PluginRequestState<'req> {
+    pub plugins: Arc<Vec<RouterPluginBoxed>>,
+    pub router_http_request: RouterHttpRequest<'req>,
```
dotansimha, line 134: I wonder if this needs to be `incoming_http_request`?

lib/executor/src/response/value.rs
```diff
@@ -43,7 +43,7 @@ impl Hash for Value<'_> {
 }
 
 impl<'a> Value<'a> {
-    pub fn take_entities(&mut self) -> Option<Vec<Value<'a>>> {
+    pub fn take_entities(&'a mut self) -> Option<Vec<Value<'a>>> {
```
dotansimha: why?

lib/executor/src/context.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
 
 pub struct ExecutionContext<'a> {
     pub response_storage: ResponsesStorage,
-    pub final_response: Value<'a>,
+    pub data: Value<'a>,
```
dotansimha, line 18: why?

lib/executor/src/context.rs
```diff
@@ -27,24 +27,19 @@ impl<'a> Default for ExecutionContext<'a> {
...
     }
 }
 
 impl<'a> ExecutionContext<'a> {
-    pub fn new(
-        query_plan: &'a QueryPlan,
-        init_final_response: Value<'a>,
-        init_errors: Vec<GraphQLError>,
-    ) -> Self {
+    pub fn new(query_plan: &'a QueryPlan, data: Value<'a>, errors: Vec<GraphQLError>) -> Self {
```
dotansimha, line 37: Same, I understand the refactor, but why here?

lib/executor/src/projection/response.rs
```diff
@@ -80,7 +80,7 @@ impl<'a> TypeName<'a> {
 pub fn project_by_operation(
     data: &Value,
     errors: Vec<GraphQLError>,
-    extensions: &Option<HashMap<String, sonic_rs::Value>>,
+    extensions: &HashMap<String, sonic_rs::Value>,
```
dotansimha, line 83: why?

lib/executor/src/plugins/hooks/on_subgraph_http_request.rs
```diff
@@ -0,0 +1,54 @@
...
+        from_graphql_error_to_bytes, EndHookPayload, FromGraphQLErrorToResponse, StartHookPayload,
+    },
+    response::graphql_error::GraphQLError,
+};
+
+pub struct OnSubgraphHttpRequestHookPayload<'exec> {
+    pub subgraph_name: &'exec str,
+
+    pub endpoint: &'exec http::Uri,
+    pub method: http::Method,
```
dotansimha, line 19: can't be a ref? `&`

ardatan: Method is something like numbers or duration etc so ref is not needed here(also doesn't work i guess).

lib/executor/src/projection/response.rs
ardatan: Looks cleaner, and in the hooks, it is cleaner to deal with an empty map when you add something.

lib/executor/src/context.rs
ardatan: Looks cleaner to be honest. Final response sounds weird. It is actually "data" not "response"

lib/executor/src/plugins/plugin_context.rs
ardatan: Why incoming? It is the request of the router. I mean it is a matter of preference for sure but router(which Apollo also uses this term) looks more explicit no?

lib/executor/src/plugins/plugin_trait.rs
ardatan: I didn't want to overcomplicate the initialization logic. This is the place where you decide to enable/disable the plugin, initialize the internals etc. For me, it was much more clear and spreading into multiple places for initialization and configuration make it overcomplicated.

bin/router/src/main.rs
ardatan: I am not sure if it really matters to add "Some"

lib/executor/src/plugins/plugin_trait.rs
ardatan: I didn't want to be strict on that point and let them decide when the plug-in should be enabled

lib/executor/src/plugins/plugin_trait.rs
ardatan: They won't be public from hive router package. Executor package is internal that we use. So they are exported from here to the router not from the router.

lib/internal/src/authorization/metadata.rs
ardatan: Because AuthorizationMetadata is part of the payload. And it was under router. It is moved there so we can use it in the payload struct.

lib/router-config/src/lib.rs
ardatan: That would enforce an object as a configuration which wasn't my goal. It can be a Boolean or another primitive in current case.

lib/router-config/src/lib.rs
ardatan: I am not fan of complicating this. With the current "from_config", the implementor decides on when to enable the plugin based on the flags or any other cases.

bin/router/src/lib.rs
ardatan: All the stuff need including the stuff for customizations like ValidationRule etc should also be exported, I agree.

lib/router-config/src/lib.rs
ardatan: It doesn't have to be object actually. It can be any serializable like boolean then that value is deserialized and passed to the from_config. It can be a boolean, integer or an object. That's also another reason why i am not fan of enforcing "enabled".

lib/router-config/src/lib.rs
dotansimha: Yeah, but I feel like we need some layer between the custom user config, and our config, and now it's just doesn't exists. 
If plugins should be enabled/disabled, let's add that and also do a standard for the `config` field. Then a user can decide what to do with it. 
From my point of view, it does not complicates thing, it just leaves us some room for future changes

ardatan: Adding a standard config field complicates things. Today we have just one function that enables/disables the plugin, handles initialization errors etc. Why would we have a config? Do you have any example? How would adding `enabled` can be simpler than just a boolean?

```yaml
plugins:
# Current one
   my-plugin: true
# Yours
plugins:
   my-plugin:
      enabled: true
      my-config???
```

lib/executor/src/plugins/plugin_context.rs
dotansimha: For me `incoming` is more clear, but I hear you, `router_` is also ok

lib/router-config/src/lib.rs
dotansimha:
```
plugins:
  my_plugin: true
```

```
plugins:
  my_plugin: 10
```

```
plugins:
  my_plugin: my_file.json
```

all of these feels so wrong

ardatan: It is a matter of preference so you can have a struct with `enabled` flag

lib/executor/src/context.rs
dotansimha: `data` is too vague, I mean, is it the data from the incoming request? the data from the subgraphs? the data we accumulate? 
`final_response` represented here better the actual content of the variable.

lib/router-config/src/lib.rs
dotansimha: I wanted to have some barrier between our config and the user's. Adding `enabled` and scoping the user's config to `config:` field solves the barrier issue, and also moves the control on enabling plugins to us. I prefer that.

lib/router-config/src/lib.rs
dotansimha: No, i meant: 
```
plugins:
   my-plugin: # user's plugin name, scoped by an object we control and can later change/extend
      enabled: true # enabled/disabled controlled by the router
      config:  # user config is scopd as object here
         my_config: something 
```

lib/executor/src/context.rs
ardatan: `data`, `errors` and `extensions`. Those are the part of the GraphQL response object. Using these names are more accurate. This might sound like nitpicking. I can also revert this change but I am 100% sure that `final_response` is misleading and `data` is more accurate. Because the response is `{ data, errors, extensions }`, and that `Value` is definitely `data` not the whole response.

lib/router-config/src/lib.rs
ardatan: What is our config? Why do we need to decide on enabling the plugin? Just for `enabled` flag, is it worth to have an extra nested structure? Do we have a plan with "our config"? I can change it for sure. Just wanted to understand if you have a plan regarding to our config. 
With taking the control of enabling plugin, the implementor won't be able to control when to enable the plugin. This is ok?

dotansimha: I think it's ok to control and have `enabled` for the user, and let's set it to `true` by default. 

Then, `config` will be there to accomodate the user's config. 

I've mentioned in our DM that I think we should make it easy for developers and offer them a good coverage on things: that's why we'll re-export many things, and also offer things for them. 

Having that additional layer of configuration will allow us later to extend plugin funtionality in case we'll need it.

dotansimha: Ok sure, let's give the user full control over the config itself, let's just scope it under `config` (as we discussed in my other comment on this line).

bin/router/src/pipeline/parser.rs
```diff
@@ -20,54 +26,114 @@ pub struct GraphQLParserPayload {
...
 
     let parsed_operation = if let Some(cached) = app_state.parse_cache.get(&cache_key).await {
         trace!("Found cached parsed operation for query");
         cached
     } else {
-        let parsed = match app_state.router_config.limits.max_tokens.as_ref() {
-            Some(cfg) => safe_parse_operation_with_token_limit(&execution_params.query, cfg.n),
-            _ => safe_parse_operation(&execution_params.query),
+        let mut document = None;
+        let mut on_end_callbacks = vec![];
```
dotansimha, line 98: @ardatan I noticed that every area where plugins involved, have a snippent similar to this one (on_end_callbacks -> build the payload -> call the plugins -> control flow -> on_end_callbacks -> decision). 

I do'nt think we need to generalize this, but would appreicate if we can have the plugin system code encapsulated a bit more, meaning something like: 

```
process_plugin_system_parse(plugin_req_state, graphql_params, document)
```

And then collect the result and do the decision making here. I feel like it could improve readability and have the plugin system code grouped somewhere.

bin/router/src/pipeline/parser.rs
```diff
@@ -20,54 +26,114 @@ pub struct GraphQLParserPayload {
...
+        if let Some(plugin_req_state) = plugin_req_state.as_ref() {
+            /* Handle on_graphql_parse hook in the plugins - START */
+            let mut start_payload = OnGraphQLParseStartHookPayload {
```
dotansimha, line 100: nit: `start_payload` -> `initial_payload` or `rolling_payload`

bin/router/src/pipeline/parser.rs
```diff
@@ -20,54 +26,114 @@ pub struct GraphQLParserPayload {
     pub cache_key: u64,
 }
 
+pub enum ParseResult {
+    Payload(GraphQLParserPayload),
+    Response(ntex::http::Response),
```
dotansimha: Should this be `ShortCircuitResponse`?

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -66,7 +64,7 @@ pub async fn collect_usage_report<'a>(
     client_request_details: &ClientRequestDetails<'a>,
     hive_usage_agent: &UsageAgent,
     usage_config: &UsageReportingConfig,
-    execution_result: &PlanExecutionOutput,
+    error_count: usize,
```
dotansimha, line 94: good catch, we don't need to whole thing passed

bin/router/src/pipeline/parser.rs
ardatan: I preferred `start_` and `end_` prefixes which points to the before and after the process. `Initial` can be a bit misleading because it is the payload of the `start`

bin/router/src/pipeline/parser.rs
ardatan: I thought to isolate the stuff but as you can see there are variables set and passed around(give and take back the ownership etc). So I'm afraid it might be overcomplicate the code there(not the logic but the syntax-wise). So return values etc will cause functions to have looong return types and arguments etc.

bin/router/src/pipeline/parser.rs
ardatan: It can be but it looked a bit long to me. Just preference. I can rename it.

dotansimha: Not a must, but it is more readable and helps to understand what kind of response we are talking about

bin/router/src/pipeline/parser.rs
dotansimha: ok!

bin/router/src/pipeline/parser.rs
dotansimha: I'm ok with complicated code in the plugin system, as long as the hotpath remains a bit simpler. 
I had in mind something like `let document = execute_parser_plugins(document, ...);` and then it just returns ownership of the result, or fallbacks if needed (using `match` here is also ok, just to decide on the final result).

lib/executor/src/context.rs
dotansimha: I get what you are saying, but this is `ExecutionContext`, and not `GraphQLResponse`. In other places, sure, but here it's an internal struct we use to build the final response

lib/executor/src/plugins/plugin_trait.rs
```diff
@@ -0,0 +1,322 @@
...
+    fn from_graphql_error_to_response(error: GraphQLError, status_code: http::StatusCode) -> Self {
+        let body = from_graphql_error_to_bytes(error);
+        ntex::http::Response::build(ntex::http::StatusCode::OK)
+            .content_type("application/json")
+            .status(status_code)
+            .body(body)
+    }
+}
+
+#[async_trait::async_trait]
```
dotansimha, line 151: Can we avoid `async_trait` and just use the `async` trait of Rust? it should be (almost) ready to use.

lib/executor/src/plugins/plugin_trait.rs
```diff
@@ -0,0 +1,322 @@
...
+    where
+        Self: Sized;
+
+    fn on_http_request<'req>(
+        &'req self,
+        start_payload: OnHttpRequestHookPayload<'req>,
+    ) -> OnHttpRequestHookResult<'req> {
+        start_payload.proceed()
+    }
+    async fn on_graphql_params<'exec>(
```
dotansimha, line 169: Since some hooks are `async`, and we are using `tokio`, it might lead to work-stealing and jumping between threads when async hook are called. 

I think `LocalSet` should be able to allow you to run the data on the same thread, then you might also need to do `Send` on the trait itself, and maybe even `!Send`?

lib/executor/src/plugins/plugin_trait.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    fn from_config(config: Self::Config) -> Result<Option<Self>, BoxError>
+    where
+        Self: Sized;
+
+    fn on_http_request<'req>(
+        &'req self,
+        start_payload: OnHttpRequestHookPayload<'req>,
+    ) -> OnHttpRequestHookResult<'req> {
+        start_payload.proceed()
```
dotansimha, line 166: For the default functions of the trait, let's use `#[inline]` to make sure they are not just doing noop function calls.

lib/executor/src/plugins/plugin_trait.rs
dotansimha: As a user, I see this: 

Image

ardatan: Hmm ok DynRouterPlugin should be private. I'll handle it.

bin/router/src/pipeline/parser.rs
ardatan: Sometimes we have more than one owned values there. It would be more complicated like a function with all types passed then all types in the return type etc.

dotansimha: we can do tuples for returned value, also for the input if it makes sense.

ardatan: Imagine `(data, variables, extensions, query_plan, ...) = some_new_fn(data, variables, extensions, query_plan, ...)` call and 
```rust
fn some_new_fn(data: DataType, variables: HashMap<...>, extensions: HashMap<..>, ...) -> (data: DataType, variables: HashMap<..> {}
```

Is it really worth it to do the things above? Repeating types over and over again just to force an extra function there?

bin/router/src/pipeline/parser.rs
ardatan: Looks better?
https://github.com/graphql-hive/router/pull/628/changes/fcb68335aded7a02d36a1c22de76ec9eb4c0964c#diff-39e2f6b493c02e71ca61e1820dd0b158a8c2c354517c9315b7c842551c266144R74

bin/router/src/pipeline/parser.rs
dotansimha: I think it does, I think it is worth grouping the plugin system code somewhere, including each hook's pipeline. 

I think `(data, variables, extensions, query_plan, ...) = some_new_fn(data, variables, extensions, query_plan, ...)` is good enough, but we can always improve it with a state-struct for grouping and taking back ownership.

lib/executor/src/plugins/plugin_trait.rs
ardatan: Discussed in DM

lib/executor/src/plugins/plugin_trait.rs
ardatan: Done!
https://github.com/graphql-hive/router/pull/628/changes/fcb68335aded7a02d36a1c22de76ec9eb4c0964c#diff-f20a79ad216bcb5831f25e048021535b671514ce82bf1aac4d36b7b90c4200cbR273

bin/router/src/pipeline/parser.rs
dotansimha: perfect

lib/executor/src/plugins/plugin_trait.rs
ardatan: Correct! 
See https://github.com/graphql-hive/router/pull/628/changes/fcb68335aded7a02d36a1c22de76ec9eb4c0964c#diff-f1ae98dc6b38bf8733269983dba53a31c4a7dfcf386e9f4c49bd31ea56f12503R20

lib/router-config/src/lib.rs
ardatan: Done!
https://github.com/graphql-hive/router/pull/628/changes/fcb68335aded7a02d36a1c22de76ec9eb4c0964c#diff-6ff9e4051e9431ff74fdf0979f043e92c25bcdc4cb1499ae5b9a41ef4aaa4d32

lib/router-config/src/lib.rs
ardatan: Done!
https://github.com/graphql-hive/router/pull/628/changes/fcb68335aded7a02d36a1c22de76ec9eb4c0964c#diff-6ff9e4051e9431ff74fdf0979f043e92c25bcdc4cb1499ae5b9a41ef4aaa4d32

bin/router/src/lib.rs
ardatan: Done!

ardatan:
@gemini-code-assist review

plugin_examples/incoming_request_deduplication/src/plugin.rs
```diff
@@ -0,0 +1,155 @@
...
+    async fn on_graphql_params<'exec>(
+        &'exec self,
+        payload: OnGraphQLParamsStartHookPayload<'exec>,
+    ) -> OnGraphQLParamsStartHookResult<'exec> {
+        let fingerprint = self.get_fingerprint(payload.router_http_request, &payload.body);
+
+        if self.in_flight_requests.contains_key(&fingerprint) {
+            let receiver = self.receiver.clone();
+            receiver.subscribe(fingerprint);
+            let (response_key, shared_response) = receiver.recv().await.unwrap();
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist Fixed!

plugin_examples/one_of/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+                            return payload.end_with_graphql_error(GraphQLError::from_message_and_code(
+                                format!(
+                                    "Variable '${}' of input object type '{}' with @oneOf directive has multiple fields set: {:?}. Only one field must be set.",
+                                    var_name,
+                                    variable_named_type,
+                                    keys_num
+                                ),
+                                "TOO_MANY_FIELDS_SET_IN_ONEOF",
+                            ),
+                            http::StatusCode::PAYLOAD_TOO_LARGE
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist Fixed!

plugin_examples/one_of/src/lib.rs
```diff
@@ -0,0 +1,322 @@
...
+    }
+}
+
+struct OneOfValidationRule {
+    one_of_types: Arc<Vec<String>>,
+}
+
+impl ValidationRule for OneOfValidationRule {
+    fn error_code<'a>(&self) -> &'a str {
+        "TOO_MANY_ROOT_FIELDS"
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist Fixed!

bin/router/src/pipeline/parser.rs
```diff
@@ -41,35 +76,79 @@ impl<'a> From<&'a GraphQLParserPayload> for GraphQLSpanOperationIdentity<'a> {
...
+                    }
+                }
+            }
+            overridden_document = start_payload.document;
+        }
+
         let cache_key = {
             let mut hasher = Xxh3::new();
-            execution_params.query.hash(&mut hasher);
+            graphql_params.query.hash(&mut hasher);
```
kamilkisiela, line 125: Is it possible that the `on_graphql_parse` hook modifies or replaces the Document (AST)?

If so then the `cache_key` will be incorrect.

Let's say the client sends query `Q1`, and the plugin rewrites/replaces it with `Q2`.
Won't the cache key represent the `Q1` in this case?
This could cause an issue with the rest of the pipeline.

The cache key is used for parse, validation and normalization caches.

kamilkisiela: To explain it in an easier way:
1. Client sends `Q1`
2. Plugin rewrites to `Q2`
3. `cache_key` still represents `Q1`
or not really?

bin/router/src/plugins/registry.rs
```diff
@@ -0,0 +1,104 @@
...
+
+impl PluginRegistry {
+    pub fn new() -> Self {
+        Self {
+            map: HashMap::new(),
+        }
+    }
+    pub fn register<P: RouterPlugin>(mut self) -> Self {
+        let plugin_name = P::plugin_name();
+        self.map.insert(
```
kamilkisiela: I think plugin execution order may currently be nondeterministic, because of `HashMap`.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=16953f1e02f60e758434c0f7b160e910

lib/executor/src/executors/http.rs
```diff
@@ -141,110 +149,197 @@ impl HTTPSubgraphExecutor {
...
+            execution_request,
+            context: &plugin_req_state.context,
+        };
+        for plugin in plugin_req_state.plugins.as_ref() {
+            let result = plugin.on_subgraph_http_request(start_payload).await;
+            start_payload = result.payload;
+            match result.control_flow {
+                StartControlFlow::Proceed => { /* continue to next plugin */ }
+                StartControlFlow::EndWithResponse(early_response) => {
+                    response = Some(early_response);
```
kamilkisiela: Maybe we should warn or error or whatever when more than one plugin tries to set an early response? This question applies to other hooks too.

lib/executor/src/executors/http.rs
```diff
@@ -141,110 +149,197 @@ impl HTTPSubgraphExecutor {
...
+            .await?;
+
             return shared_response.deserialize_http_response();
         }
 
-        let fingerprint = request_fingerprint(&http::Method::POST, &self.endpoint, &headers, &body);
-        let inflight_span =
-            HttpInflightRequestSpan::new(&http::Method::POST, &self.endpoint, &headers, &body);
+        let fingerprint =
+            request_fingerprint(&method, &self.endpoint, &execution_request.headers, &body);
```
kamilkisiela: The `request_fingerprint` is triggered before `on_subgraph_http_request` hooks run, and these hooks can mutate request fields.

I think this can incorrectly merge different requests into one shared in-flight response.

bin/router/src/plugins/registry.rs
```diff
@@ -0,0 +1,104 @@
...
+                        if let Some(plugin) = maybe_plugin {
+                            info!("Plugin '{}' successfully enabled", plugin_name);
+                            plugins.push(plugin);
+                        } else {
+                            warn!("Plugin '{}' disabled during initialization", plugin_name);
+                        }
+                    }
+                }
+            } else {
+                return Err(PluginRegistryError::MissingInRegistry(
```
kamilkisiela, line 96: When a plugin is not registered, but defined in the configuration as disabled `my_plugin: { enabled: false }`, then Router will crash the startup with "missing plugin" error.

Is it intentional? Should we do it like that?

lib/graphql-tools/src/ast/schema_visitor.rs
```diff
@@ -5,171 +5,364 @@ use crate::static_graphql::schema::{
...
-                                    field,
-                                    input_object,
-                                    _visitor_context,
-                                );
-                                self.leave_input_object_type_field(
-                                    field,
-                                    input_object,
-                                    _visitor_context,
-                                );
+    fn visit_schema_document(
```
kamilkisiela: I'm not a fan of this change.
I know GraphQL-JS visitor can mutate schemas, but visitor here is read-only.

We do have a separate trait for transforming operations. We have it so if users want to visit the document/operation, it's read-only and fast with 0 allocations.

If we turn the visitor into mutable visitor, then there's no way to visit the schema with no allocations.

I suggest to have a similar trait to the one for transforming operations that will accept 3 result variants:
```
TransformedValue::Keeep
TransformedValue::Replace(new_value)
TransformedValue::Drop
```

bin/router/src/pipeline/parser.rs
ardatan: If you set `document` before it starts, it returns that document without caching.
If you change the `document` in `on_end`, it doesn't touch the original one in the cache and returns the modified.

lib/executor/src/executors/http.rs
ardatan: Hmm actually there is `break;` after this line so not sure if more than one plugin can set an early response.

lib/executor/src/executors/error.rs
```diff
@@ -1,46 +1,55 @@
...
+    EndpointParseFailure(String, InvalidUri),
     #[error("Failed to compile VRL expression for subgraph '{0}'. Please check your VRL expression for syntax errors. Diagnostic: {1}")]
     #[strum(serialize = "SUBGRAPH_ENDPOINT_EXPRESSION_BUILD_FAILURE")]
     EndpointExpressionBuild(String, String),
-    #[error("Failed to resolve VRL expression for subgraph '{0}'. Runtime error: {1}")]
+    #[error("Failed to resolve VRL expression. Runtime error: {0}")]
     #[strum(serialize = "SUBGRAPH_ENDPOINT_EXPRESSION_RESOLUTION_FAILURE")]
-    EndpointExpressionResolutionFailure(String, String),
-    #[error("VRL expression for subgraph '{0}' resolved to a non-string value.")]
+    EndpointExpressionResolutionFailure(String),
```
kamilkisiela, line 14: I see you removed info about subgraphs and vrl errors. Is there a way for a developer to know why something broke and for which subgraph?

lib/query-planner/src/consumer_schema/mod.rs
```diff
@@ -1,22 +1,41 @@
...
-    pub document: Document<'static, String>,
+    pub document: Arc<Document>,
+    pub hash: u64,
+}
+
+impl From<Arc<Document>> for ConsumerSchema {
+    fn from(document: Arc<Document>) -> Self {
+        let hash = {
+            let mut hasher = Xxh3::new();
+            let sdl = document.to_string();
```
kamilkisiela: I know it's not super critical performance-wise, as it does not happen often, but it's just inefficient. It could have `Hash` impl that is order-independent.

kamilkisiela: This way we could avoid allocating potentially 50MB schema (or more when people have big supergraphs) and spend a lot of synchronous time on hashing big blob of text.

lib/executor/src/executors/error.rs
ardatan: `extensions.serviceName` is still there actually. That's why I removed it. I think I forgot to extract it to another PR. But I can revert this change.

bin/router/src/plugins/registry.rs
ardatan: Moved `enabled` check on top. Good catch!

bin/router/src/plugins/registry.rs
ardatan: I changed it to BTreeMap.

lib/query-planner/src/consumer_schema/mod.rs
ardatan: Could you explain your point a bit more?

bin/router/src/plugins/registry.rs
kamilkisiela: Are you sure `BTreeMap` solves the problem, did you verify?
I expect the plugins to run in the same order as they are registered

bin/router/src/pipeline/parser.rs
kamilkisiela: I'm not sure what you mean by "it", is it the `on_graphql_parse` call? And if so, you mean that it may return the document without using `on_end`?

Another question: there's no way to modify documents and cache them with the router's cache?

lib/executor/src/executors/error.rs
kamilkisiela: The `serviceName` is fine, as long as it's visible in stdout/stderror, and not only in the http response sent to the client. Without it logged in the terminal, developers won't know about the error.

There's also the second thing I asked about, the error created by vrl. Do we log it?

lib/query-planner/src/consumer_schema/mod.rs
kamilkisiela: When you call `document.to_string()` it uses `Display` implementation to allocate an intiial String (buffer), then it iterates over all types and fields and their arguments, to write more and more str to that buffer.

When it's finished, you end up with a big String that is sent to a hashing function, that will byte by byte create a hash, and at the end you drop (deallocate) the big String and return `u64`.

What I mean is to implement hashing, that avoids `Display`'s overhead and does not allocate a String.
You can `impl Hash` for all AST nodes and iterate definitions in the schema Document, and go deeper and deeper and call AST node's `hash` function.
When doing it, the hash should be order-indepenedent, meaning `type O { foo: String bar: String }` should have the same hash as `type O { bar: String foo: String }` as they are exact same schemas, just the fields are in different order.

lib/executor/src/executors/http.rs
ardatan: Ok I changed it in the similar way we have in parser, validation and query plan hooks.
So first we invoke `on_subgraph_http_request` with all parameters, then in case of deduplication, fingerprint is generated after that.
And in `on_end`, we have deduplication hint to let the user know about the deduplication.

lib/executor/src/executors/error.rs
ardatan: I added logging for the execution errors here!
https://github.com/graphql-hive/router/pull/628/changes#diff-33c9e06e738dde0f63ef177987a5821d328c78989ca484ed2654a29c3a4a293fR455

bin/router/src/pipeline/parser.rs
ardatan: By it, I am referring to the default behavior. `on_end` is called always with the result either from plugins or the actual parser we have.
You can modify the final document in `on_end`.

So,
- If you set an early document result before any cache checks, parser job etc happens, the document is set, and on_end callbacks are called without any caching stuff.
- If you change the document in `on_end`, it doesn't modify the copy in the cache. It just affects that request pipeline.

Letting user control the cache is tricky because the user can have values depending on the request or some other parameters.

lib/query-planner/src/consumer_schema/mod.rs
ardatan: https://github.com/graphql-hive/router/pull/772
Implemented it here. Let me know what you think.

bin/router/src/plugins/registry.rs
ardatan: https://github.com/graphql-hive/router/pull/628/commits/e805f258cde483adb0daa2550eb21b7489082380
I changed the map in the registry to `BTreeMap` so the order of registration is kept. After initiating plugins, I re-create the plugins vec by using the order of the registration. So plugins are invoked in the same order they are registered

lib/executor/src/context.rs
ardatan: It is not touched by this PR anymore

lib/executor/src/context.rs
ardatan: It is not touched by this PR anymore

lib/executor/src/projection/response.rs
ardatan: It is not touched by this PR anymore

lib/executor/src/response/value.rs
ardatan: Removed the diff

lib/graphql-tools/src/ast/schema_visitor.rs
ardatan: Removed those changes. I don't touch the schema visitor anymore.

bin/router/src/pipeline/parser.rs
ardatan: Removed this functionality as discussed on Slack.

bin/router/src/plugins/registry.rs
kamilkisiela: Are you 100% sure? https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=50de49e9f21faa6d9c8d66314378976d

Here I insert things to BTreeMap, then `iter()` and they are yielded in sorted a-z order, not insertion order

kamilkisiela: and why not use a vector here? Then the order is super simple to achieve

dotansimha: `Vec` seems to make sense here, more than a map.

dotansimha: And if you nee the plugin name, do a `Vec<(name, plugin)>`?

ardatan: Implemented it with tests
https://github.com/graphql-hive/router/pull/628/commits/2b281458e9b10ed0ad6ace48baedfd2030d595d7

.github/workflows/ci.yaml
```diff
@@ -128,22 +144,41 @@ jobs:
...
+            package: "hive-router"
+            binary: "hive_router"
+            config: "bench/configs/default.config.yaml"
+            compare_with_default: true
           - name: "no-dedupe"
             args: ""
+            package: "hive-router"
+            binary: "hive_router"
             config: bench/configs/no-deduplication.config.yaml
+            compare_with_default: false
```
kamilkisiela, line 161: I would just keep CI job failing till merged into `main`. Can you do a follow-up PR (after this one gets merged) to remove that logic, as it's not needed, I think.

lib/graphql-tools/src/ast/schema_visitor.rs
kamilkisiela: Can you explain what you did instead?

ardatan: https://github.com/graphql-hive/router/blob/plugin-system/plugin_examples/feature_flags/src/plugin.rs#L60
I just implemented a basic example that removes the definitions if needed.

.github/workflows/ci.yaml
ardatan: Hmm my idea is actually to compare the results between the case when the plugin is enabled and when it is not.

kamilkisiela: Ok, then let's leave it as is, but when we add more logic to the measured plugin at some point, we would need to switch to plugin vs plugin. Today we measure 3 noop function calls.

PR #629, opened 2025-12-22
fix: handle projection conditions per item in an array, and merge plans nestedly

ardatan:
Ref ROUTER-235

User reported that they have a schema like following;

```graphql
interface IContent {
  id: ID!
}

interface IElementContent {
  id: ID!
}

type ContentAChild {
  title: String!
}

type ContentA implements IContent & IElementContent {
  id: ID!
  contentChildren: [ContentAChild!]!
}

type ContentBChild {
  title: String!
}

type ContentB implements IContent & IElementContent {
  id: ID!
  contentChildren: [ContentBChild!]!
}

type Query {
  contentPage: ContentPage!
}

type ContentPage {
  contentBody: [ContentContainer!]!
}

type ContentContainer {
  id: ID!
  section: IContent
}
```

And when they send the following query;

```graphql
query {
  contentPage {
    contentBody {
      section {
        ...ContentAData
        ...ContentBData
      }
    }
  }
}

fragment ContentAData on ContentA {
  contentChildren {
    title
  }
}

fragment ContentBData on ContentB {
  contentChildren {
    title
  }
}
```

The subgraph returns the following data;
```json
{
            "contentPage": {
                    "contentBody": [
                        {
                            "id": "container1",
                            "section": {
                                "__typename": "ContentA",
                                "id": "contentA1",
                                "contentChildren": []
                            }
                        },
                        {
                            "id": "container2",
                            "section": {
                                "__typename": "ContentB",
                                "id": "contentB1",
                                "contentChildren": [
                                    {
                                        "title": "contentBChild1"
                                    }
                                ]
                            }
                        }
                    ]
                }
        }
```

But the router responds like below;

```json
{
            "contentPage": {
                    "contentBody": [
                        {
                            "id": "container1",
                            "section": {
                                "__typename": "ContentA",
                                "id": "contentA1",
                                "contentChildren": []
                            }
                        },
                        {
                            "id": "container2",
                            "section": {
                                "__typename": "ContentB",
                                "id": "contentB1",
                                "contentChildren": null
                            }
                        }
                    ]
                }
        }
```

Initially I assumed that the subgraph returns the data with `__typenames` like below;

```json

            "__typename": "Query",
            "contentPage": [
                {
                    "__typename": "ContentPage",
                    "contentBody": [
                        {
                            "__typename": "ContentContainer",
                            "id": "container1",
                            "section": {
                                "__typename": "ContentA",
                                "contentChildren": []
                            }
                        },
                        {
                            "__typename": "ContentContainer",
                            "id": "container2",
                            "section": {
                                "__typename": "ContentB",
                                "contentChildren": [
                                    {
                                        "__typename": "ContentBChild",
                                        "title": "contentBChild1"
                                    }
                                ]
                            }
                        }
                    ]
                }
            ]
        }
  ```
  
 And I saw that projection puts `null` to `contentChildren` on the second item because it expects `ContentBChild` but it gets `ContentAChild`. I realized that something is wrong with the plan merging logic when it is nested, it was duplicating the fields in the plan so for each possible type, the field is serialized like `"id": "value", "id": "value"` because it wasn't merging the same fields when they are nested (so it can have multiple plans for the same field).
 So I did a refactor to apply the plan merging nestedly but in order to do that I needed to convert `Vec` usage to `IndexMap` which is needed by merging logic. Then this ended up changing all `Vec<Plan>` types to `IndexMap`s.
  
The merging was fine but it was still getting `ContentBChild`s typename as `ContentAChild` because it couldn't get `__typename` from the array and it was using `plan.field_type` as a fallback incorrectly. Then I did another refactor to apply condition checking for arrays item-by-item instead of using the fallback value directly, because fallback value was always the first possible item of the abstract field type (ContentAChild) in our case while it can be `ContentBChild`.

Then the first test started passing as expected but the user's issue is not still solved. Why? Because it was never sending __typename s for non-abstract values because planner didn't need it to be sent. So I dropped the fallback logic for __typename, and check for __typenames only when it is there which made fallback logic unnecessary and I dropped `field_type` from the projection plan entirely.
So the second test case which is user's actual issue started passing.

While working on this, I dropped unnecessary Arc usage, and added a helper function `get_value_by_key` to reduce some duplicate logic in the code, I also improved tracing messages so we can get more information about projection condition checks.

### DIFF
--- a/.changeset/nested-array.md
+++ b/.changeset/nested-array.md
@@ -1,0 +1,139 @@
+---
+router: patch
+executor: patch
+---
+
+Handle nested abstract items while projecting correctly;
+
+For the following schema;
+
+```graphql
+interface IContent {
+  id: ID!
+}
+
+interface IElementContent {
+  id: ID!
+}
+
+type ContentAChild {
+  title: String!
+}
+
+type ContentA implements IContent & IElementContent {
+  id: ID!
+  contentChildren: [ContentAChild!]!
+}
+
+type ContentBChild {
+  title: String!
+}
+
+type ContentB implements IContent & IElementContent {
+  id: ID!
+  contentChildren: [ContentBChild!]!
+}
+
+type Query {
+  contentPage: ContentPage!
+}
+
+type ContentPage {
+  contentBody: [ContentContainer!]!
+}
+
+type ContentContainer {
+  id: ID!
+  section: IContent
+}
+```
+
+```graphql
+query {
+  contentPage {
+    contentBody {
+      section {
+        ...ContentAData
+        ...ContentBData
+      }
+    }
+  }
+}
+
+fragment ContentAData on ContentA {
+  contentChildren {
+    title
+  }
+}
+
+fragment ContentBData on ContentB {
+  contentChildren {
+    title
+  }
+}
+```
+
+If a query like above is executed, the projection plan should be able to handle nested abstract types correctly.
+
+For the following subgraph response, array items should be handled by their own `__typename` values individually;
+
+```json
+{
+  "__typename": "Query",
+  "contentPage": [
+    {
+      "__typename": "ContentPage",
+      "contentBody": [
+        {
+          "__typename": "ContentContainer",
+          "id": "container1",
+          "section": {
+            "__typename": "ContentA",
+            "contentChildren": []
+          }
+        },
+        {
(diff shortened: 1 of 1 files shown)


dotansimha:
Replaced by https://github.com/graphql-hive/router/pull/633

Issue #645, opened 2026-01-08
Ability to specify an endpoint other than /graphql


mikeq:
I have had a look through the configuration options, and had a look through the code (I am not a rust developer).

Is the graphql endpoint `/graphql' hard coded and cannot be specified in the router config?

I think this line of code https://github.com/graphql-hive/router/blob/main/bin/router/src/lib.rs#L137

Mike

PR #649, opened 2026-01-11
feat(router): allow to customize gql endpoint

dotansimha:
Fixes https://github.com/graphql-hive/router/issues/645 / ROUTER-238

### DIFF
--- a/bin/router/src/http_utils/landing_page.rs
+++ b/bin/router/src/http_utils/landing_page.rs
@@ -4,9 +4,9 @@ use ntex::{http::ResponseBuilder, web::Responder};
 static LANDING_PAGE_HTML: &str = include_str!("../../static/landing-page.html");
 static PRODUCT_LOGO_SVG: &str = include_str!("../../static/product_logo.svg");
 
-pub async fn landing_page_handler() -> impl Responder {
+pub async fn landing_page_handler(graphql_endpoint: String) -> impl Responder {
     let rendered_html = LANDING_PAGE_HTML
-        .replace("__GRAPHIQL_LINK__", "/graphql")
+        .replace("__GRAPHIQL_LINK__", &graphql_endpoint)
         .replace("__PRODUCT_NAME__", "Hive Router")
         .replace(
             "__PRODUCT_DESCRIPTION__",
--- a/bin/router/src/lib.rs
+++ b/bin/router/src/lib.rs
@@ -80,17 +80,19 @@ pub async fn router_entrypoint() -> Result<(), Box<dyn std::error::Error>> {
     let router_config = load_config(config_path)?;
     configure_logging(&router_config.log);
     info!("hive-router@{} starting...", ROUTER_VERSION);
+    let http_config = router_config.http.clone();
     let addr = router_config.http.address();
     let mut bg_tasks_manager = BackgroundTasksManager::new();
     let (shared_state, schema_state) =
         configure_app_from_config(router_config, &mut bg_tasks_manager).await?;
 
     let maybe_error = web::HttpServer::new(move || {
+        let lp_gql_path = http_config.graphql_endpoint().to_string();
         web::App::new()
             .state(shared_state.clone())
             .state(schema_state.clone())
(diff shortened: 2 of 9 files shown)


bin/router/src/lib.rs
```diff
@@ -80,17 +80,19 @@ pub async fn router_entrypoint() -> Result<(), Box<dyn std::error::Error>> {
...
 
     let maybe_error = web::HttpServer::new(move || {
+        let lp_gql_path = http_config.graphql_path().to_string();
         web::App::new()
             .state(shared_state.clone())
             .state(schema_state.clone())
-            .configure(configure_ntex_app)
-            .default_service(web::to(landing_page_handler))
+            .configure(|m| configure_ntex_app(m, http_config.graphql_path()))
+            .default_service(web::to(move || landing_page_handler(lp_gql_path.clone())))
```
ardatan, line 95:
```suggestion
            .default_service(web::to(move || landing_page_handler(http_config.graphql_path())))
```
We can send `&str` to here still?

lib/router-config/src/http_server.rs
```diff
@@ -1,9 +1,13 @@
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct HttpServerConfig {
+    /// The endpoint to serve GraphQL requests. By default, `/graphql` is used.
+    #[serde(default = "graphql_path_default")]
+    path: String,
```
ardatan:
```suggestion
    graphql_endpoint: String,
```
For the future path configurations for health etc, being more explicit like `graphql_endpoint` can be better, no?

bin/router/src/http_utils/landing_page.rs
```diff
@@ -4,9 +4,9 @@ use ntex::{http::ResponseBuilder, web::Responder};
 static LANDING_PAGE_HTML: &str = include_str!("../../static/landing-page.html");
 static PRODUCT_LOGO_SVG: &str = include_str!("../../static/product_logo.svg");
 
-pub async fn landing_page_handler() -> impl Responder {
+pub async fn landing_page_handler(graphql_path: String) -> impl Responder {
```
ardatan:
```suggestion
pub async fn landing_page_handler(graphql_path: &str) -> impl Responder {
```

bin/router/src/lib.rs
dotansimha: not really, because it's already used before and it needs to be for `'static`

lib/router-config/src/http_server.rs
dotansimha: sure, makes sense!

docs/README.md
```diff
@@ -65,6 +65,7 @@ headers:
             named: x-tenant-id
             rename: x-acct-tenant
 http:
+  graphql_endpoint: /graphql
```
kamilkisiela, line 68: how do you imagine configuring ws endpoint for subscriptions?

dotansimha: I think eventually we'll need to talk about where the WS config goes. I think it might go into a root config field called `ws:`?

dotansimha: Just because the current one is `http`. We can decide to make a breaking change here and do `server:` for both, but that's really open for discussion.

ardatan: Why should WS endpoint be different?

dotansimha: It shouldn't in most cases, but I think it makes sense to allow to customize it separately

ardatan (approved):

docs/README.md
kamilkisiela: nobody said it should be different, I asked this question because other routers allow for such customization, and we should keep that on mind and make a room for it. If we name things badly, it may bite us in future.

It's always good to think about broader context, than this narrow graphql endpoint scope...

PR #652, opened 2026-01-12
refactor(executor): handle header propagation outside of the plan executor

ardatan:

Combining with https://github.com/graphql-hive/router/pull/651 , the aim is to reduce the number of extra structs.

Today, `PlanExecutionOutput` which is the return type of `execute_query_plan` in `executor/execution/plan.rs`, returns `body: Vec<u8>` and `headers: http::HeaderMap`.
Then, in the `router/pipeline`, `ResponseBuilder` is called with `Bytes::from` of `PlanExecutionOutput.body` and `PlanExecutionOutput.headers` is converted to `ntex::HeaderMap`, then `content_type` is set here.

With this PR, the response aggregator is sent with `PlanExecutionOutput`, and during the response build, the response headers aggregator sets the headers into `ResponseBuilder` directly.

So now `modify_client_headers` helper accepts `ResponseBuilder`, not `http:HeaderMap`. And since this throws `HeaderPropagationError` in the pipeline executor, I had to add `HeaderPropagationError` to `PipelineError`.

I also extracted `PipelineError` to `FailedExecutionResult` logic to a `From` trait so that other future implementations that serializes and sets the body in a different way can use it and serialize `FailedExecutionResult` on their own.
Currently it is used here only -> https://github.com/graphql-hive/router/pull/652/changes#diff-f3dc2f2d579374645281d6950414266d8984195dbf45c00f0a754ceca8550aadR95

----
Update:
After recent changes, I see that `project_by_operation` is used for `dry_run` but it is actually unnecessary as we just need to serialize `extensions` which has nothing to do with projection since it uses `sonic_rs::Value`.

### DIFF
--- a/bin/router/src/pipeline/error.rs
+++ b/bin/router/src/pipeline/error.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use graphql_tools::validation::utils::ValidationError;
 use hive_router_plan_executor::{
     execution::{error::PlanExecutionError, jwt_forward::JwtForwardingError},
+    headers::errors::HeaderRuleRuntimeError,
     response::graphql_error::GraphQLError,
 };
 use hive_router_query_planner::{
@@ -118,6 +119,10 @@ pub enum PipelineError {
     #[error("Subscriptions are not supported over accepted transport(s)")]
     #[strum(serialize = "SUBSCRIPTIONS_TRANSPORT_NOT_SUPPORTED")]
     SubscriptionsTransportNotSupported,
+
+    #[error(transparent)]
+    #[strum(serialize = "HEADER_PROPAGATION_FAILURE")]
+    HeaderPropagation(#[from] HeaderRuleRuntimeError),
 }
 
 impl PipelineError {
@@ -169,6 +174,7 @@ impl PipelineError {
             (Self::IntrospectionDisabled, _) => StatusCode::FORBIDDEN,
             (Self::SubscriptionsNotSupported, _) => StatusCode::UNSUPPORTED_MEDIA_TYPE,
             (Self::SubscriptionsTransportNotSupported, _) => StatusCode::NOT_ACCEPTABLE,
+            (Self::HeaderPropagation(_), _) => StatusCode::INTERNAL_SERVER_ERROR,
         }
     }
 
--- a/bin/router/src/pipeline/execution.rs
+++ b/bin/router/src/pipeline/execution.rs
@@ -31,7 +31,6 @@ pub struct PlannedRequest<'req> {
     pub variable_payload: &'req CoerceVariablesPayload,
     pub client_request_details: &'req ClientRequestDetails<'req>,
     pub authorization_errors: Vec<AuthorizationError>,
-    pub response_content_type: &'static str,
 }
 
 #[inline]
@@ -63,12 +62,15 @@ pub async fn execute_plan(
     };
 
     if matches!(expose_query_plan, ExposeQueryPlanMode::DryRun) {
-        let response = ntex::http::Response::Ok().json(&json!({
+        let body_json = json!({
             "extensions": extensions,
-        }));
+        });
+
+        let body = sonic_rs::to_vec(&body_json).unwrap();
 
         return Ok(PlanExecutionOutput {
-            response,
+            body,
+            response_header_aggregator: None,
             error_count: 0,
         });
     }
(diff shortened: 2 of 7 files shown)


lib/executor/src/execution/plan.rs
```diff
@@ -120,9 +113,18 @@ pub async fn execute_query_plan<'exec, 'req>(
         affected_path: || None,
     })?;
 
+    let mut response = ntex::http::Response::Ok()
+        .content_type(ctx.response_content_type)
+        .body(body);
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist Done!

bin/router/src/pipeline/error.rs
```diff
@@ -142,3 +142,39 @@ pub struct FailedExecutionResult {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub errors: Option<Vec<GraphQLError>>,
 }
+
+// This allows easy conversion from PipelineError to FailedExecutionResult
+// FailedExecutionResult is agnostic to HTTP so later on we can use it with different serialization formats
+impl From<PipelineError> for FailedExecutionResult {
```
kamilkisiela, line 205: it wasn't here no?

ardatan: True, it wasn't. I moved error to result logic to a trait so that other handlers (non-HTTP ones etc) can use it too.
I forgot to add it to the description, just added it.
https://github.com/graphql-hive/router/pull/652/changes#diff-f3dc2f2d579374645281d6950414266d8984195dbf45c00f0a754ceca8550aadR95

lib/executor/src/execution/plan.rs
```diff
@@ -117,9 +111,18 @@ pub async fn execute_query_plan<'exec>(
...
+
+    modify_client_response_headers(exec_ctx.response_headers_aggregator, response.headers_mut())
```
kamilkisiela: It gives users the power to modify the content-type header, we're okay with it?

ardatan: I can prevent that inside `modify_client_response_headers` ?

kamilkisiela: I'm just asking for opinion, if we should do allow it or not

ardatan: It is also a header at the end of the day. Maybe user wants to have a special json type? I think it is fine to allow it.

bin/router/src/pipeline/execution.rs
```diff
@@ -64,31 +63,12 @@ pub async fn execute_plan(
...
-        .map_err(|err| {
-            tracing::error!(
-                "Failed to project query plan to extensions during dry-run: {}",
-                err
-            );
-            PipelineError::PlanExecutionError(err)
-        })?;
+        let response = ntex::http::Response::Ok().json(&json!({
+            "extensions": extensions,
+        }));
```
enisdenjo: Love this.

lib/executor/src/execution/plan.rs
enisdenjo: Client should not impact the calls to subgraphs, and neither should subgraphs to clients. There should be complete separation of client<->router and router<->subgraph - the only thing that can propagate are the allowed headers as per the config - excluding `content-type` and `accept`.

bin/router/src/pipeline/mod.rs
```diff
@@ -219,6 +208,7 @@ pub async fn execute_pipeline<'exec>(
     supergraph: &SupergraphData,
     shared_state: &Arc<RouterSharedState>,
     schema_state: &Arc<SchemaState>,
+    response_content_type: &'static str,
```
enisdenjo: Why does the execution_pipeline need to know about the response `content-type`? As I said in previous comment, they should be separate an the client should in no way influence the content-type of internal subgraph calls. Subscriptions, for example, would not have a content-type at all when dealing with WebSockets. 

The execution_pipeline should be oblivious to the _why or for whom_ it's resolving something.

ardatan: I moved the response header propagation to the pipeline executor. My point is to avoid creating a new HeaderMap in the execution then re-iterate it while creating `ntex:http::Response`.  I think with my last commit, the plan executor becomes fully HTTP independent as you wish, right?

enisdenjo: It's looking great! ~~I am just wondering why not simply HeaderMap? Why an abstraction over it? At the end of theday, there is no hiding from the fact that the executing a query plan needs headers, so why not just HeaderMap?~~

enisdenjo (approved):

Issue #646, opened 2026-01-09
Disable/protect introspection for production environments


bugy:
Hi all, thanks for the great project!

I'm curious if you considered a setting to disable introspection or hide it behind an additional authorization (i.e. only for specific user roles or with some additional api key) 
It's quite a [common practice ](https://www.apollographql.com/blog/why-you-should-disable-graphql-introspection-in-production) to hide introspection from users, so potential attackers wouldn't know all possible queries/mutations and fields. Of course, this is kind of "security through obscurity", but it's still an additional line of defense. And some other routers (apollo, cosmo) provide such a setting.

dotansimha:
Hi @bugy , thanks for reaching out. 

> I'm curious if you considered a setting to disable introspection

Yeah, this is a very common practice, and we do have plans to improve the readiness of the Router for production envs (@ardatan was working on https://github.com/graphql-hive/router/pull/623 to also improve it). 

> hide it behind an additional authorization (i.e. only for specific user roles or with some additional api key)

I think ideally, we can use the `expressions` (https://the-guild.dev/graphql/hive/docs/router/configuration/expressions) in order to allow some kind of conditional mechanism for deciding if introspection needs to be disabled or not. How does that sounds?

dotansimha:
Thanks for sharing! <br>NODE_ENV is a bit tricky as we are no in NodeJS env, and at the moment we don't have way to identify a production environment, so we might just need the user to decide if they want to disable/enable introspection, using the YAML config file. Same for GraphiQL endpoint. <br><br>Regarding the expressions, I defenitly think that it needs to be in addition to a simple boolean - it's only for cases where you might want to disable introspection dynamically based on some attributes from the incoming request.

PR #655, opened 2026-01-12
feat(router): Enable/Disable Introspection with `introspection`

ardatan:
Closes https://github.com/graphql-hive/router/issues/646
Ref ROUTER-239

Documentation -> https://github.com/graphql-hive/console/pull/7483

Adds a new option `introspection` on the root level so you can enable/disable introspection either statically with a boolean or dynamically with expressions (headers, env vars)

By default, introspection is allowed. However if an expression or boolean is present and it is resolved to `false` which means "introspection should be disabled".

In case of `introspection` is resolved to `false`
```graphql
query {
  __schema {  ... }
}
```
will be resolved as;
```js
{
   "errors": [
     {
        "message": "Introspection queries are disabled",
        "extensions": {
             "code": "INTROSPECTION_DISABLED"
         }
     }
   ]
}
```

### DIFF
--- a/bin/router/src/shared_state.rs
+++ b/bin/router/src/shared_state.rs
@@ -117,6 +117,6 @@ pub enum SharedStateError {
     OverrideLabelsCompile(#[from] Box<OverrideLabelsCompileError>),
     #[error("error creating hive usage agent: {0}")]
     UsageAgent(#[from] Box<AgentError>),
-    #[error("invalid disable introspection config: {0}")]
-    DisableIntrospectionCompile(#[from] Box<ExpressionCompileError>),
+    #[error("invalid introspection config: {0}")]
+    IntrospectionPolicyCompile(#[from] Box<ExpressionCompileError>),
 }


lib/internal/src/expressions/values/boolean.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use crate::expressions::{FromVrlValue, ValueOrProgram};
+use vrl::core::Value as VrlValue;
+
+/// Type alias for a Boolean that can be either static or computed via expression
+pub type BooleanOrProgram = ValueOrProgram<bool>;
+
+/// Error type for String conversion failures
+#[derive(Debug, thiserror::Error, Clone)]
+pub enum StringConversionError {
```
kamilkisiela: I think it shouldn't be **String**ConversionError, but Boolean

lib/internal/src/expressions/values/boolean.rs
```diff
@@ -0,0 +1,35 @@
+use std::string::FromUtf8Error;
+
+use crate::expressions::{FromVrlValue, ValueOrProgram};
+use vrl::core::Value as VrlValue;
+
+/// Type alias for a Boolean that can be either static or computed via expression
+pub type BooleanOrProgram = ValueOrProgram<bool>;
+
+/// Error type for String conversion failures
```
kamilkisiela:
```suggestion
/// Error type for Boolean conversion failures
```

lib/internal/src/expressions/values/boolean.rs
```diff
@@ -0,0 +1,35 @@
...
+#[derive(Debug, thiserror::Error, Clone)]
+pub enum StringConversionError {
+    #[error("Failed to convert bytes to UTF-8 string: {0}")]
+    InvalidUtf8(#[from] FromUtf8Error),
```
kamilkisiela: This variant is not used

bin/router/src/pipeline/allow_introspection.rs
```diff
@@ -0,0 +1,60 @@
...
+    allow_introspection_program: &BooleanOrProgram,
+    introspection_context: &mut IntrospectionContext,
+    client_request_details: &client_request_details::ClientRequestDetails<'_, '_>,
+    default_value: bool,
+    initial_errors: &mut Vec<GraphQLError>,
+) -> Result<(), PipelineErrorVariant> {
+    let mut context_map = BTreeMap::new();
+    context_map.insert("request".into(), client_request_details.into());
+
+    context_map.insert("default".into(), VrlValue::Boolean(default_value));
```
kamilkisiela: Should we expose `.default`? I think it's weird to have:

```
introspection: if ... { true } else { .default }
```

bin/router/src/shared_state.rs
```diff
@@ -110,4 +116,6 @@ pub enum SharedStateError {
     OverrideLabelsCompile(#[from] Box<OverrideLabelsCompileError>),
     #[error("error creating hive usage agent: {0}")]
     UsageAgent(#[from] Box<AgentError>),
+    #[error("invalid disable introspection config: {0}")]
+    DisableIntrospectionCompile(#[from] Box<ExpressionCompileError>),
```
kamilkisiela: I think it's outdated, we don't have `disable_introspection`

bin/router/src/shared_state.rs
```diff
@@ -96,6 +100,8 @@ impl RouterSharedState {
             .map_err(Box::new)?,
             jwt_auth_runtime,
             hive_usage_agent,
+            allow_introspection: compile_allow_introspection(&router_config.introspection)
```
kamilkisiela: Maybe it would be better to name it: `introspection_policy` or `introspection_enabled`.

bin/router/src/pipeline/mod.rs
```diff
@@ -36,6 +36,7 @@ use crate::{
     shared_state::RouterSharedState,
 };
 
+pub mod allow_introspection;
```
kamilkisiela: it could be just `introspection`

bin/router/src/pipeline/error.rs
```diff
@@ -150,6 +153,7 @@ impl PipelineErrorVariant {
             (Self::MissingContentTypeHeader, _) => StatusCode::NOT_ACCEPTABLE,
             (Self::UnsupportedContentType, _) => StatusCode::UNSUPPORTED_MEDIA_TYPE,
             (Self::CsrfPreventionFailed, _) => StatusCode::FORBIDDEN,
+            (Self::AllowIntrospectionEvaluationError(_), _) => StatusCode::INTERNAL_SERVER_ERROR,
```
kamilkisiela: AllowIntrospection sounds like an action to take, it could be named differently, based on my comment regarding naming of the app state and the module.

bin/router/src/pipeline/allow_introspection.rs
```diff
@@ -0,0 +1,60 @@
...
+        .map(|config| match config {
+            AllowIntrospectionConfig::Boolean(b) => Ok(BooleanOrProgram::Value(*b)),
+            AllowIntrospectionConfig::Expression { expression } => expression
+                .compile_expression(None)
+                .map(|program| BooleanOrProgram::Program(Box::new(program))),
+        })
+        .transpose()
+}
+
+pub fn handle_allow_introspection(
```
kamilkisiela: Could be named `verify_introspection_access` or something similar

bin/router/src/shared_state.rs
```diff
@@ -70,6 +73,7 @@ pub struct RouterSharedState {
     pub jwt_claims_cache: JwtClaimsCache,
     pub jwt_auth_runtime: Option<JwtAuthRuntime>,
     pub hive_usage_agent: Option<UsageAgent>,
+    pub allow_introspection: Option<BooleanOrProgram>,
```
kamilkisiela: Not sure why we have an Option here, could be just `BooleanOrProgram` and by default it's `true`.

bin/router/src/pipeline/allow_introspection.rs
```diff
@@ -0,0 +1,60 @@
...
+    initial_errors: &mut Vec<GraphQLError>,
+) -> Result<(), PipelineErrorVariant> {
+    let mut context_map = BTreeMap::new();
+    context_map.insert("request".into(), client_request_details.into());
+
+    context_map.insert("default".into(), VrlValue::Boolean(default_value));
+
+    let context = VrlValue::Object(context_map);
+    let is_enabled = allow_introspection_program
+        .resolve(context)
```
kamilkisiela: I will make a Pull Request to change `resolve` function to accept a function that returns the context, so it's lazy and will improve performance on the hot path.

kamilkisiela: https://github.com/graphql-hive/router/pull/667

lib/internal/src/expressions/values/boolean.rs
ardatan: Yeah you're right! I had a different thought but fixed it anyways!

lib/internal/src/expressions/values/boolean.rs
ardatan: Fixed!

lib/internal/src/expressions/values/boolean.rs
ardatan: Removed!

bin/router/src/pipeline/allow_introspection.rs
ardatan: Fixed

bin/router/src/shared_state.rs
ardatan: Fixed!

bin/router/src/shared_state.rs
ardatan: Renamed to `introspection_policy`!

bin/router/src/pipeline/mod.rs
ardatan: Renamed to `introspection_policy` matching the rest. Sounds good?

bin/router/src/pipeline/error.rs
ardatan: Fixed!

bin/router/src/shared_state.rs
ardatan: Fixed!

bin/router/src/pipeline/allow_introspection.rs
ardatan: Adapted the changes!

bin/router/src/pipeline/allow_introspection.rs
ardatan: Looks better now?

bin/router/src/pipeline/introspection_policy.rs
```diff
@@ -0,0 +1,54 @@
...
+        .resolve(|| {
+            let mut context_map = BTreeMap::new();
+            context_map.insert("request".into(), client_request_details.into());
+
+            VrlValue::Object(context_map)
+        })
+        .map_err(|e| PipelineError::IntrospectionPermissionEvaluationError(e.to_string()))?;
+
+    if !is_enabled {
+        introspection_context.query = None;
```
kamilkisiela: When I look at popular GraphQL server implementations, when introspection is disabled, but the query contains `__schema` or other introspection fields, the entire query is rejected.

Why you decided to allow partial results?

ardatan: Changed it. It no longer allows partial results. It rejects the entire query.

bin/router/Cargo.toml
```diff
@@ -60,6 +60,7 @@ cookie = "0.18.1"
 arc-swap = "1.7.1"
 lasso2 = "0.8.2"
 anyhow = { workspace = true }
+insta.workspace = true
```
dotansimha: nit: `insta = { workspace = true }`

bin/router/src/pipeline/introspection_policy.rs
kamilkisiela: We should move this up the pipeline, to not perform query planning or authorization of fields and things like that, when we know the query contains introspection and the introspection is forbidden, so it will be rejected anyway.

ardatan: I moved it after the normalization

PR #659, opened 2026-01-12
enhance(executor): simplify subgraph execution error handling

ardatan:

Now `SubgraphExecutor` can throw `SubgraphExecutorError` which will be coerced to `PlanExecutorError`.
So that the execution related errors and the errors propagated from the subgraph can be distinguished.
Combining this with https://github.com/graphql-hive/router/pull/651, we no longer need to generate dummy subgraph response objects with `GraphQLError`s. It was a bit complicated before because in different levels, helpers like `error_to_graphql_bytes` and `internal_server_error_response` were handling those errors in different places, so it was hard to debug it.

Right now the subgraph response will be the actual subgraph response while the rest of the execution errors thrown by us will be handled as `PlanExecutionError`.

Of course an error in a plan node will not stop the request entirely for such cases like a subgraph is down etc which is tested in an E2E test added with this PR.

To summarize the idea is to take the responsibility of `error` to `response` conversion from the subgraph executor to the plan executor.

I also removed unused methods;
- HeaderRuleCompileError.new_expression_build
- HeaderRuleRuntimeError.new_value_conversion

And removed unnecessary methods;
- HeaderRuleRuntimeError.new_expression_evaluation

### DIFF
--- a/e2e/src/override_subgraph_urls.rs
+++ b/e2e/src/override_subgraph_urls.rs
@@ -77,8 +77,11 @@ mod override_subgraph_urls_e2e_tests {
         let body = test::read_body(resp).await;
         let json_body: Value = from_slice(&body).unwrap();
 
-        insta::assert_snapshot!(sonic_rs::to_string_pretty(&json_body).unwrap(), @r###"
+        insta::assert_snapshot!(sonic_rs::to_string_pretty(&json_body).unwrap(), @r#"
         {
+          "data": {
+            "users": null
+          },
           "errors": [
             {
               "message": "Failed to send request to subgraph \"http://0.0.0.0:4200/accounts\": client error (Connect)",
@@ -88,7 +91,8 @@ mod override_subgraph_urls_e2e_tests {
               }
             }
           ]
-        }"###);
+        }
+        "#);
 
         let subgraph_requests = subgraphs_server
             .get_subgraph_requests_log("accounts")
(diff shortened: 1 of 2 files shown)


e2e/src/error_handling.rs
```diff
@@ -0,0 +1,98 @@
...
+        wait_for_readiness(&app.app).await;
+
+        let req = init_graphql_request("{ me { reviews { id product { upc name } } } }", None);
+        let resp = test::call_service(&app.app, req.to_request()).await;
+
+        assert!(resp.status().is_success(), "Expected 200 OK");
+        let resp_body_bytes = test::read_body(resp).await;
+        let resp_json: Value = from_slice(&resp_body_bytes).expect("expected valid JSON response");
+
+        let data = resp_json
```
kamilkisiela: why not insta snapshot assertions? This way we'd know what the output is. By reading this code I have no idea how it looks like

ardatan: I updated it with snapshots. Do you think it looks better now?

e2e/src/override_subgraph_urls.rs
```diff
@@ -1,7 +1,7 @@
 #[cfg(test)]
 mod override_subgraph_urls_e2e_tests {
     use ntex::web::test;
-    use sonic_rs::{from_slice, Value};
+    use sonic_rs::{from_slice, JsonValueTrait as _, Value};
```
kamilkisiela: I think it's unused no?

ardatan: `.as_str()` is imported from this;
https://github.com/graphql-hive/router/pull/659/changes#diff-6c0acb32874b6fb95dd74796c446bbb447f8362c2bca12c31e1d8221ddf8d0baR82

But I changed them to use snapshots anyways

Issue #305, opened 2025-08-07
RFC: Plugin system


dotansimha:
[TBD] 

Make sure to cover:
- Plugins API 
- Lifecycle of each phase 
- Reasoning for each phase (what plugins will need that, ...)

PR #665, opened 2026-01-12
refactor: Execution pipeline without HTTP request dependency and better accept header parsing

enisdenjo:
Ref ROUTER-240

`execution_pipeline` does not depend on the HTTP anymore, `graphql_request_handler` does all of the HTTP request handling.

As a result, this allows the `execution_pipeline` and all of its dependencies to return HTTP request independant errors, ~simplifying the error handling and removing the `PipelineError` leaving only `PipelineErrorVariant`~ (mostly already done in #[637](https://github.com/graphql-hive/router/pull/637)). 

Also, this PR allows the upcoming subscriptions PR (#620) to include also WebSockets implementation[^1] and not be too huge for review. Because of this PR being a prerequisite, there's some subscriptions stuff with a `// coming soon` just to expidite the process of subscriptions and to have better understanding on whats coming.

Besides the WS stuff, removing the HTTP request dependency from `execution_pipeline` allows us to have onit tests for the pipeline as well as mocking (which might be a thing when we get the #305).

Furthermore we now use [`headers-accept` crate](https://crates.io/crates/headers-accept) to negotiate the content type as per the official HTTP spec (correct parameter handling and q-weighting).

Bonus points go to: expose query plan dry-run is now really dry. Also added tests.

### Why do we return both single and streaming accepted content type when parsing the accept header?

Some clients ([like Apollo](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol#executing-a-subscription)) sends `Accept: multipart/mixed;subscriptionSpec="1.0", application/json` headers indicating that the client can also stream. This enables router to send single results as JSON responses but stream multiple results as multipart. Router will inteligently choose one or the other.

As a bonus, it also allows the router to stream single responses.

### TODOs

- [x] Changesets

[^1]: At the time of writing #620 does not implement WebSockets, but merging this PR will allow it inthe PR.

### DIFF
--- a/bin/router/src/pipeline/header.rs
+++ b/bin/router/src/pipeline/header.rs
@@ -38,14 +38,19 @@ impl TryFrom<&MediaType<'_>> for SingleContentType {
 
     /// The only thing where the conversion can fail is if the media type is not supported.
     fn try_from(media_type: &MediaType) -> Result<Self, Self::Error> {
-        let essence_media_type_string = media_type.essence().to_string();
-        if essence_media_type_string == SingleContentType::GraphQLResponseJSON.as_ref() {
-            Ok(SingleContentType::GraphQLResponseJSON)
-        } else if essence_media_type_string == SingleContentType::JSON.as_ref() {
-            Ok(SingleContentType::JSON)
-        } else {
-            Err("Unsupported single content type")
+        let ty = media_type.ty.as_str();
+        let subty = media_type.subty.as_str();
+        let suffix = media_type.suffix.as_ref().map(|s| s.as_str());
+
+        if ty == "application" {
+            if subty == "graphql-response" && suffix == Some("json") {
+                return Ok(SingleContentType::GraphQLResponseJSON);
+            } else if subty == "json" && suffix.is_none() {
+                return Ok(SingleContentType::JSON);
+            }
         }
+
+        Err("Unsupported single content type")
     }
 }
 
@@ -103,21 +108,23 @@ impl TryFrom<&MediaType<'_>> for StreamContentType {
 
     /// The only thing where the conversion can fail is if the media type is not supported.
     fn try_from(media_type: &MediaType) -> Result<Self, Self::Error> {
-        let essence_media_type_string = media_type.essence().to_string();
-        if essence_media_type_string == StreamContentType::IncrementalDelivery.as_ref() {
+        let ty = media_type.ty.as_str();
+        let subty = media_type.subty.as_str();
+
+        if ty == "multipart" && subty == "mixed" {
             if media_type
                 .get_param(Name::new_unchecked("subscriptionSpec"))
                 .is_some_and(|s| s.unquoted_str() == "1.0")
             {
-                Ok(StreamContentType::ApolloMultipartHTTP)
+                return Ok(StreamContentType::ApolloMultipartHTTP);
             } else {
-                Ok(StreamContentType::IncrementalDelivery)
+                return Ok(StreamContentType::IncrementalDelivery);
             }
-        } else if essence_media_type_string == StreamContentType::SSE.as_ref() {
-            Ok(StreamContentType::SSE)
-        } else {
-            Err("Unsupported stream content type")
+        } else if ty == "text" && subty == "event-stream" {
+            return Ok(StreamContentType::SSE);
         }
+
+        Err("Unsupported stream content type")
     }
 }
 


bin/router/src/pipeline/error.rs
```diff
@@ -72,6 +78,12 @@ pub enum PipelineError {
     JwtError(JwtError),
     #[error("Failed to forward jwt: {0}")]
     JwtForwardingError(JwtForwardingError),
+
+    // Subscription-related errors
+    #[error("Subscriptions are not supported")]
+    SubscriptionsNotSupport,
```
kamilkisiela: misses `ed`, a typo I guess?

bin/router/src/pipeline/error.rs
```diff
@@ -95,6 +107,8 @@ impl PipelineError {
                 "OPERATION_RESOLUTION_FAILURE"
             }
             Self::JwtError(err) => err.error_code(),
+            Self::SubscriptionsNotSupport => "SUBSCRIPTIONS_NOT_SUPPORT",
```
kamilkisiela: same typo

bin/router/src/pipeline/execution.rs
```diff
@@ -35,27 +34,13 @@ pub struct PlannedRequest<'req> {
...
-                "true" => expose_query_plan = ExposeQueryPlanMode::Yes,
-                "dry-run" => expose_query_plan = ExposeQueryPlanMode::DryRun,
-                _ => {}
-            }
-        }
-    }
-
-    let extensions = if expose_query_plan == ExposeQueryPlanMode::Yes
-        || expose_query_plan == ExposeQueryPlanMode::DryRun
+    let extensions = if *expose_query_plan == ExposeQueryPlanMode::Yes
```
kamilkisiela: You could use something like:
```
if matches!(
    expose_query_plan,
    ExposeQueryPlanMode::Yes | ExposeQueryPlanMode::DryRun
) {
```

It reads better

bin/router/src/pipeline/header.rs
```diff
@@ -2,27 +2,434 @@ use http::{header::ACCEPT, HeaderValue};
...
+    ///
+    /// Read more: https://graphql.github.io/graphql-over-http
+    GraphQLResponseJSON,
+    /// Legacy GraphQL over HTTP (`application/json`)
+    #[default]
+    JSON,
+}
+
+impl SingleContentType {
+    pub fn as_str(&self) -> &'static str {
```
kamilkisiela: it's all const and static, so we could make it a `const fn`.
It gets compiled nicely and could be used (in future) in other const values, like:
```
const DEFAULT_CONTENT_TYPE: &str = SingleContentType::JSON.as_str();
```

if we ever need it, just a nit.

bin/router/src/pipeline/header.rs
```diff
@@ -2,27 +2,434 @@ use http::{header::ACCEPT, HeaderValue};
...
+            SingleContentType::GraphQLResponseJSON => &APPLICATION_GRAPHQL_RESPONSE_JSON_STR,
+            SingleContentType::JSON => &APPLICATION_JSON_STR,
+        }
+    }
+}
+
+/// Streamable content types for GraphQL responses.
+#[derive(Default, Debug)]
+pub enum StreamContentType {
+    /// Incremental Delivery over HTTP (`multipart/mixed`)
```
kamilkisiela, line 84: Great you add comments and document what and why. I chase it :)

bin/router/src/pipeline/header.rs
```diff
@@ -2,27 +2,434 @@ use http::{header::ACCEPT, HeaderValue};
...
+    /// Read more: https://github.com/graphql/graphql-over-http/blob/d285c9f31897ea51e231ebfe8dcb481a354431c9/rfcs/GraphQLOverSSE.md
+    SSE,
+    /// Apollo Multipart HTTP protocol (`multipart/mixed;subscriptionSpec="1.0"`)
+    ///
+    /// Read more: https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol
+    ApolloMultipartHTTP,
+}
+
+impl StreamContentType {
+    pub fn as_str(&self) -> &'static str {
```
kamilkisiela: same `const` nit

bin/router/src/pipeline/header.rs
```diff
@@ -2,27 +2,434 @@ use http::{header::ACCEPT, HeaderValue};
...
+
+#[derive(Debug)]
+enum SupportedContentType {
+    Single(SingleContentType),
+    Stream(StreamContentType),
+}
+
+impl SupportedContentType {
+    fn parse(content_type: &str) -> Option<SupportedContentType> {
+        if content_type == *APPLICATION_GRAPHQL_RESPONSE_JSON_STR {
```
kamilkisiela: Maybe it's worth using [mime](https://crates.io/crates/mime) crate and do pattern matching instead?

We compare header values by `eq` operator, but what if they contain additional params separated with `;`?

bin/router/src/pipeline/header.rs
```diff
@@ -2,27 +2,434 @@ use http::{header::ACCEPT, HeaderValue};
...
+    fn accepted_content_type(
+        &self,
+    ) -> Result<(Option<SingleContentType>, Option<StreamContentType>), PipelineError> {
+        let content_types = self
+            .headers()
+            .get(ACCEPT)
+            .and_then(|value| value.to_str().ok())
+            .unwrap_or("");
+
+        let (single_content_type, stream_content_type) =
```
kamilkisiela: I think you could do:
```rust
let parsed @ (single, stream) = SupportedContentType::parse_header(content_types);
```

bin/router/src/pipeline/header.rs
```diff
@@ -2,27 +2,434 @@ use http::{header::ACCEPT, HeaderValue};
...
+
+        // at this point we treat no content type as "user explicitly does not support any known types"
+        // this is because only empty accept header or */* is treated as "accept everything" and we check
+        // that above
+        match (single_content_type, stream_content_type) {
+            (Some(single), Some(stream)) => Ok((Some(single), Some(stream))),
+            (Some(single), None) => Ok((Some(single), None)),
+            (None, Some(stream)) => Ok((None, Some(stream))),
+            (None, None) => Err(PipelineError::UnsupportedContentType),
+        }
```
kamilkisiela: and then:
```rust
    if single.is_none() && stream.is_none() {
        Err(PipelineError::UnsupportedContentType)
    } else {
        Ok(parsed)
    }
```

kamilkisiela: because all the patterns result in identical output

bin/router/src/pipeline/error.rs
enisdenjo: yes 😬 fixed

bin/router/src/pipeline/normalize.rs
```diff
@@ -25,7 +25,7 @@ pub struct GraphQLNormalizationPayload {
 #[inline]
 pub async fn normalize_request_with_cache(
     supergraph: &SupergraphData,
-    schema_state: &SchemaState,
+    schema_state: &Arc<SchemaState>,
```
kamilkisiela: are those (here and below) left after wrong rebase?

lib/hive-console-sdk/Cargo.toml
```diff
@@ -25,7 +25,7 @@ anyhow = "1"
 tracing = "0.1"
 serde = "1"
 tokio = { version = "1.36.0", features = ["full"] }
-md5 = "0.8.0"
+md5 = "0.7.0"
```
kamilkisiela: same

bin/router/src/pipeline/header.rs
enisdenjo: This all changed when I introduced the headers-accept crate. Please review again.

bin/router/src/pipeline/header.rs
enisdenjo: [Replied on the other comment.](https://github.com/graphql-hive/router/pull/665#discussion_r2686326456)

bin/router/src/pipeline/normalize.rs
enisdenjo: coming from the subscriptions pr, reverted

bin/router/src/pipeline/header.rs
enisdenjo: I love it!

bin/router/src/pipeline/header.rs
enisdenjo: didnt know that syntax, nice!

enisdenjo: I didnt use it though because that introduced a partial move and I had to either drop the `@` or implement the copy trait which would copy those variables leading to memory waste (probably)

bin/router/src/pipeline/parser.rs
```diff
@@ -18,7 +18,7 @@ pub struct GraphQLParserPayload {
 
 #[inline]
 pub async fn parse_operation_with_cache(
-    app_state: &RouterSharedState,
+    app_state: &Arc<RouterSharedState>,
```
ardatan:
```suggestion
    app_state: &RouterSharedState,
```

bin/router/src/pipeline/query_plan.rs
```diff
@@ -12,7 +12,7 @@ use xxhash_rust::xxh3::Xxh3;
 #[inline]
 pub async fn plan_operation_with_cache(
     supergraph: &SupergraphData,
-    schema_state: &SchemaState,
+    schema_state: &Arc<SchemaState>,
```
ardatan:
```suggestion
    schema_state: &SchemaState,
```

bin/router/src/pipeline/validation.rs
```diff
@@ -10,8 +10,8 @@ use tracing::{error, trace};
 #[inline]
 pub async fn validate_operation_with_cache(
     supergraph: &SupergraphData,
-    schema_state: &SchemaState,
-    app_state: &RouterSharedState,
+    schema_state: &Arc<SchemaState>,
```
ardatan:
```suggestion
    schema_state: &SchemaState,
```

bin/router/src/pipeline/validation.rs
```diff
@@ -10,8 +10,8 @@ use tracing::{error, trace};
...
+    app_state: &Arc<RouterSharedState>,
```
ardatan:
```suggestion
    app_state: &RouterSharedState,
```

bin/router/src/pipeline/header.rs
```diff
@@ -1,28 +1,250 @@
...
+        if media_type == &GRAPHQL_RESPONSE_JSON_MEDIA_TYPE {
+            Some(SingleContentType::GraphQLResponseJSON)
+        } else if media_type == &JSON_MEDIA_TYPE {
+            Some(SingleContentType::JSON)
+        } else {
+            None
+        }
+    }
+
+    pub const fn as_str(&self) -> &'static str {
```
ardatan: There is `strum` package we already use that creates static str converters for enums. Might be cleaner???

enisdenjo: The strum pkg has `IntoStaticStr` derive macro which does not support `.as_str()`. It is meant to be used like this:

```rs
let stat: &'static str = something.into();
```

And `into` cannot be used in const methods so we also cant:

```
pub const fn as_str(&self) -> &'static str {
    self.into();
}
```

TL;DR we'd have to use type annotations/casting to use `IntoStaticStr` and writing `&'static str` everywhere seems more work than just implementing `as_str()`. 

What do you think?

bin/router/src/pipeline/header.rs
enisdenjo: Now using headers-accept crate. Updated PR description, please read for context before reviewing.

bin/router/src/pipeline/header.rs
ardatan: Oh not a big deal. I thought adding a derive macro and  instead of implementing whole trait with `match`es and `as_str(` methods might be easier. But I don't have a strong opinion. It was just an idea.
Because I guess it also support from strings not just into.

enisdenjo: Sorted in https://github.com/graphql-hive/router/pull/665/commits/078d5a84afbcb77b7a359a96fbe7f20580362b0d, thanks Arda!

bin/router/src/pipeline/execution.rs
```diff
@@ -35,27 +34,13 @@ pub struct PlannedRequest<'req> {
...
-                "dry-run" => expose_query_plan = ExposeQueryPlanMode::DryRun,
-                _ => {}
-            }
-        }
-    }
-
-    let extensions = if expose_query_plan == ExposeQueryPlanMode::Yes
-        || expose_query_plan == ExposeQueryPlanMode::DryRun
+    let extensions = if matches!(expose_query_plan, ExposeQueryPlanMode::Yes)
+        || matches!(expose_query_plan, ExposeQueryPlanMode::DryRun)
```
kamilkisiela:
```suggestion
    let extensions = if matches!(expose_query_plan, ExposeQueryPlanMode::Yes | ExposeQueryPlanMode::DryRun)
```

enisdenjo: nice

bin/router/src/jwt/mod.rs
```diff
@@ -52,11 +52,11 @@ impl JwtAuthRuntime {
         Ok(instance)
     }
 
-    fn lookup(&self, req: &HttpRequest) -> Result<(Option<String>, String), LookupError> {
+    fn lookup(&self, headers: &HeaderMap) -> Result<(Option<String>, String), LookupError> {
```
dotansimha, line 55: This all makes perfect sense. We didn't have to depends on the full request anyway here.

bin/router/src/pipeline/mod.rs
```diff
@@ -54,103 +48,62 @@ pub mod query_plan;
 pub mod usage_reporting;
 pub mod validation;
 
-static GRAPHIQL_HTML: &str = include_str!("../../static/graphiql.html");
-
 #[inline]
 pub async fn graphql_request_handler(
     req: &HttpRequest,
     body_bytes: Bytes,
+    single_content_type: Option<SingleContentType>,
```
dotansimha: Should we just have a simpler way to manage this variaty in content-type?

I mean, something like this: 

```rs
struct RequestContentType {
  Single(SingleContentType)
  Stream(StreamContentType)
}

impl RequestContentType {
   // utilities here if needed, like `is_stream` or `content_type` for getting it easily
}

// ... 
graphql_request_handler(
    req: &HttpRequest,
    body_bytes: Bytes,
    content_type: Option<RequestContentType>,
```

This way the code inside this fn becomes a bit simpler and doesn't have to be so explicit about the difference between streaming and single. 

With this setup, you can also start by doing a single enum variant in this PR (for single) and no need for the placeholder, as the `enum` can grow later when you add the streaming one.

bin/router/src/pipeline/coerce_variables.rs
```diff
@@ -30,21 +27,10 @@ impl CoerceVariablesPayload {
 
 #[inline]
 pub fn coerce_request_variables(
-    req: &HttpRequest,
     supergraph: &SupergraphData,
     execution_params: &mut ExecutionRequest,
     normalized_operation: &Arc<GraphQLNormalizationPayload>,
 ) -> Result<CoerceVariablesPayload, PipelineError> {
-    if req.method() == Method::GET {
```
dotansimha, line 37: Yeah, it makes perfect sense to moving it away from this function!

bin/router/src/pipeline/header.rs
```diff
@@ -1,28 +1,240 @@
...
+
+    #[inline]
+    fn negotiate(
+        &self,
+    ) -> Result<(Option<SingleContentType>, Option<StreamContentType>), PipelineError> {
+        let content_types = self
+            .headers()
+            .get(ACCEPT)
+            .and_then(|value| value.to_str().ok())
+            .unwrap_or("");
```
dotansimha: does `""` has an actual value over just re-using the `Option` and `None`?

bin/router/src/pipeline/header.rs
```diff
@@ -1,28 +1,240 @@
...
-            .map(|s| s.contains(content_type))
+            .map(|s| s.contains(TEXT_HTML_MIME))
             .unwrap_or(false)
     }
+
+    #[inline]
+    fn negotiate(
+        &self,
+    ) -> Result<(Option<SingleContentType>, Option<StreamContentType>), PipelineError> {
+        let content_types = self
```
dotansimha, line 249: nit: `content_types` while it's a single string? `content_type`?

bin/router/src/pipeline/header.rs
```diff
@@ -1,28 +1,240 @@
...
+
+        let agreed = negotiate_content_type(content_types).map_err(|err| {
+            error!("Failed to parse Accept header: {}", err);
+            PipelineError::InvalidHeaderValue(ACCEPT)
+        })?;
+
+        // at this point we treat no content type as "user explicitly does not support any known types"
+        // this is because only empty accept header or */* is treated as "accept everything" and we check
+        // that above
+        if agreed.0.is_none() && agreed.1.is_none() {
```
dotansimha: while tuples are amazing in Rust, the `.0` and `.1` are not super readable :) 

Please see my comment above about using an `enum` for that purpose, is might also make this piece of code more readble, and the enum can also host the `negotiate` / `negotiate_content_type` functions maybe?

bin/router/src/pipeline/error.rs
```diff
@@ -95,6 +107,8 @@ impl PipelineError {
                 "OPERATION_RESOLUTION_FAILURE"
             }
             Self::JwtError(err) => err.error_code(),
+            Self::SubscriptionsNotSupported => "SUBSCRIPTIONS_NOT_SUPPORTED",
```
dotansimha: nit: all of these could be just `strum` (in a separate PR of course)

bin/router/src/pipeline/execution.rs
```diff
@@ -35,28 +34,15 @@ pub struct PlannedRequest<'req> {
...
-            }
-        }
-    }
-
-    let extensions = if expose_query_plan == ExposeQueryPlanMode::Yes
-        || expose_query_plan == ExposeQueryPlanMode::DryRun
-    {
+    let extensions = if matches!(
+        expose_query_plan,
+        ExposeQueryPlanMode::Yes | ExposeQueryPlanMode::DryRun
```
dotansimha, line 56: I think `DryRun` had a unique meaning where the QP runs and returns the extensions with the QP, but doesn't really execute. I'm not sure when/where it was lost, but it seems like now it will run the query even when `DryRun` is used?

bin/router/src/pipeline/header.rs
```diff
@@ -1,28 +1,240 @@
...
-        "application/graphql-response+json";
-    pub static ref TEXT_HTML_CONTENT_TYPE: &'static str = "text/html";
-    pub static ref APPLICATION_GRAPHQL_RESPONSE_JSON: HeaderValue =
-        HeaderValue::from_static(&APPLICATION_GRAPHQL_RESPONSE_JSON_STR);
+use crate::pipeline::error::PipelineError;
+
+/// Non-GraphQL content type, used to detect if the client can accept GraphiQL responses.
+pub const TEXT_HTML_MIME: &str = "text/html";
+
+const JSON_MEDIA_TYPE: MediaType =
```
dotansimha: `const` over `lazy_static` is great!

bin/router/src/lib.rs
```diff
@@ -20,19 +20,28 @@ use crate::{
...
+    header::{CONTENT_TYPE, RETRY_AFTER},
+    Method,
+};
 use ntex::{
     util::Bytes,
     web::{self, HttpRequest},
 };
 use tracing::{info, warn};
 
+static GRAPHIQL_HTML: &str = include_str!("../static/graphiql.html");
```
dotansimha, line 46: I didn't like `GRAPHIQL_HTML` living in `lib`/`mod` anyway, should it just live in some graphiql-specific file?

bin/router/src/lib.rs
```diff
@@ -51,14 +60,42 @@ async fn graphql_endpoint_handler(
...
 
-        let mut res = graphql_request_handler(
+        // Aggree on the response content type so that errors can be handled
+        // properly outside the request handler.
+        let (single_content_type, stream_content_type) = match request.negotiate() {
+            Ok((single, stream)) => (single, stream),
+            Err(err) => return err.into_response(None),
+        };
+
+        if request.method() == Method::GET
```
dotansimha: could also be simplied if we'll do the enum thing

bin/router/src/pipeline/header.rs
dotansimha: With that being said, I just saw https://github.com/picoHz/mediatype/pull/25 and the discussion there, with the root Crate. 
I didn't see much activity there in the past few days, so if it remains, I prefer using the stable one, with `lazy_static` + `unwrap`. And then if it lands and released, we can always update the dependency and the code here.

bin/router/src/pipeline/header.rs
```diff
@@ -1,28 +1,240 @@
...
+            StreamContentType::ApolloMultipartHTTP
+        }
+    }
+}
+
+/// Reads the `Accept` header contents and returns a tuple of accepted/parsed content types.
+/// It perform negotiation and respects q-weights.
+fn negotiate_content_type(
+    accept_header: &str,
+) -> Result<(Option<SingleContentType>, Option<StreamContentType>), HeaderError> {
```
dotansimha: I assume that you added `header-core` and the import because you inherit the `Header` error from `Accept` crate and it's not publicly exposed?

If so, you can get rid of the dependency and the import, and use `<Accept as FromStr>::Err` ;)

bin/router/Cargo.toml
```diff
@@ -57,6 +58,9 @@ cookie = "0.18.1"
 arc-swap = "1.7.1"
 lasso2 = "0.8.2"
 anyhow = { workspace = true }
+mediatype = "0.21.0"
+headers-core = "0.3.0"
```
dotansimha: See my other comment regarding the need for this dependency.

Cargo.toml
```diff
@@ -66,3 +66,8 @@ ahash = "0.8.12"
 once_cell = "1.21.3"
 anyhow = "1.0.100"
 moka = { version = "0.12.11", features = ["future", "sync"] }
+strum = { version = "0.27.2", features = ["derive"] }
+
+# uses branch until https://github.com/picoHz/mediatype/pull/25 lands
+[patch.crates-io]
+mediatype = { git = "https://github.com/enisdenjo/mediatype", branch = "public-new-unchecked" }
```
dotansimha: see my other comment regarding this dependency

bin/router/src/pipeline/coerce_variables.rs
dotansimha: also removing `req` from here makes a lot of sense. We don't need the full req at all

bin/router/src/pipeline/execution_request.rs
```diff
@@ -83,7 +83,7 @@ impl TryInto<ExecutionRequest> for GETQueryParams {
 }
 
 #[inline]
-pub async fn get_execution_request(
+pub async fn get_execution_request_from_http_request(
```
dotansimha, line 86: nit: this could be more rusty

```
pub async fn get_execution_request_from_http_request(
    req: &HttpRequest,
    body_bytes: Bytes,
) -> Result<ExecutionRequest, PipelineError> {
```

➡️ 

```
impl TryFrom<(&HttpRequest, body_bytes: Bytes)> for ExecutionRequest {
   type Err = PipelineError;

   // ...
}
```

bin/router/src/pipeline/header.rs
```diff
@@ -1,28 +1,240 @@
...
+                Some(StreamContentType::ApolloMultipartHTTP),
+            ),
+            (
+                r#"text/event-stream, application/json"#,
+                Some(SingleContentType::JSON),
+                Some(StreamContentType::SSE),
+            ),
+        ];
+
+        for (accept_header, expected_single, expected_stream) in cases {
```
dotansimha: nit: failing a test inside a loop is not very readable. I prefer the error to point me to the line in code, and not to the line in code + some `[i]` that I don't know of.

bin/router/src/pipeline/mod.rs
```diff
@@ -201,19 +137,92 @@ pub async fn execute_pipeline(
...
+        }
+    }
+
+    Ok(response_builder
+        .header(http::header::CONTENT_TYPE, accepted_content_type)
+        .body(response_bytes))
+}
+
+#[inline]
+#[allow(clippy::await_holding_refcell_ref, clippy::too_many_arguments)]
```
dotansimha: I noticed this `clippy::await_holding_refcell_ref` and I wasn't sure I see anywhere an actual point of async holding a RefCell. Pretty sure it can be removed.

bin/router/src/pipeline/execution_request.rs
ardatan: Oh this looks uglier in my opinion. Do we have a similar usage anywhere else? Not sure.

bin/router/src/pipeline/mod.rs
enisdenjo: I've had this, but removed it in https://github.com/graphql-hive/router/pull/665/changes/463471b7b9b4f6adc82ac8e6928a9b19f6437e0f. The reason being is that negotiation does not choose _only one_ content type. It chooses two, one for single and one for stream responses. Then during operation execution, one or the other is selected. Because of this, I got rid of the `RequestContentType` to avoid the assumption of having _one_ content type.

Keeping this in mind, and after thorough refactoring, the only purpose a `RequestContentType` had is to parse a header - which is now a `impl RequestAccepts for HttpRequest`. The is_stream or is_single can be simple enum matches; yeah the checks would help - but you really always want to match both enums and branch out depending on the selection.

bin/router/src/pipeline/execution_request.rs
dotansimha: It's also harder to implement sometimes, especially if there any many things variables needed (in this case, 2). I agree it's not beautiful, that's why it's `nit` 😆

dotansimha: Also @ardatan , I think we don't have many of these, because every time I tried to use it, i reached some edge cases and had to switch to a raw `fn`. 😆

bin/router/src/pipeline/header.rs
enisdenjo: The `negotiate_content_type` would always require text to initate negotion, that was my resoning. But since it gives defaults for the lack of it, I refactored to accept an option. 👍

bin/router/src/pipeline/header.rs
enisdenjo: I answered in the comment above why the enum didnt make sense to me. `.0` and `.1` are only used here. I wanted to use the destructure thingy `let agreed @ { single, stream } = ...` but that does a partial copy and that was a waste. (There's a conversation with Kamil somewhere here).

bin/router/src/pipeline/mod.rs
dotansimha: Ah, I see. I wonder if Type-State pattern can help here? to distiguish between two difference possible state of the content type? (https://www.youtube.com/watch?v=_ccDqRTx-JU) 
(it uses `enum`s, but allow you better code readability and inference based on the enum variation you hold, without any runtime impact)

bin/router/src/lib.rs
enisdenjo: We dont have one, is it really worth creating one just to re-export the static asset? I placed it here because lib actually does respond with the graphiql html.

bin/router/src/lib.rs
enisdenjo: Answered in the original comment, this is really the only place where it would simplify - but this is an exception for the html response so I didnt include it because html response != graphql response.

bin/router/src/lib.rs
dotansimha: Not really, you are right. But if the logic of "should we render graphiql?" is moved there, it might make more sense. But for sure it's out of scope for your PR

bin/router/src/pipeline/mod.rs
enisdenjo: I thought about that one too, but to me it seemed anologous to the enum approach - one _or_ the other?

Remember, we choose 2 content types at all time. Here are the possible cases:
- Some(single) and None(stream): can only single response, errors out on subscriptions
- None(single) and Some(stream): will always respond, queries get streamed so would subscriptions. errors are also streamed
- Some(single) and Some(stream): will always respond, queries are single responses, subscriptions are streams. errors are single responses (because client accepts both)
- (just for the exhaustive list) None and None: of course fail reqeust immediately

dotansimha: Yeah, it just seems a bit exhaustive 🤔 

So from what you said, I learn that we have `None` (4th) which is easy, but the possible combo is tricky. 

Should we just be more explicit and deal with content type in some area of the code, and then later roll this as something easier for us to work with? 

```
enum ConnectionMode { // or, ResponseMode? 
    SingleOnly,    // Some(single), None(stream)
    StreamOnly,    // None(single), Some(stream)
    Dual,          // Some(single), Some(stream)
}
``` 

(just really trying to avoid that `Option, Option`, i feel like it complicates also other areas of the code)

enisdenjo: Oooh, nice one. I like how that looks! Let me try it out, I'll report back here

enisdenjo: p.s. I dont like the `(Option, Option)` either. I'd like to avoid it too

bin/router/src/pipeline/header.rs
enisdenjo: It _is_ a single string, but that single string can, and usually will, contain multiple content types. 😬

dotansimha: ah ok :)

bin/router/src/pipeline/execution.rs
enisdenjo: Funnily enough, it's broken on `main` too. The check is lacking the same. 😬 

I'll fix it here, good catch.

enisdenjo: https://github.com/graphql-hive/router/pull/665/commits/ed8d6d416f3068d5e09702490f6429a0e0c0ce5f, added tests for all 3 cases (disabled, dry-run, true).

bin/router/src/lib.rs
enisdenjo: Made a task to not forget. https://linear.app/the-guild/issue/GW-559/refactor-graphiql-handling

bin/router/src/pipeline/header.rs
enisdenjo: That's exactly it, fixed! Thanks.

bin/router/src/pipeline/mod.rs
enisdenjo: Inherited from main, removed. 👍

bin/router/src/pipeline/execution.rs
dotansimha: thanks!

bin/router/src/pipeline/mod.rs
enisdenjo: here it is https://github.com/graphql-hive/router/pull/665/commits/94684e1d5e8696db503d9877e02ec50b52109e4d, it's looking very nice. it also covers a few other things:
- `graphql_request_handler` doesnt have `single_content_type: Option<SingleContentType>, stream_content_type: Option<StreamContentType>` arguments anymore and we therefore make sure there's always a valid response mode (no `None, None` which is invalid)
- no more `agreed.0.is_none() && agreed.1.is_none()`
- looks nice

bin/router/src/pipeline/header.rs
enisdenjo: not the case anymore since https://github.com/graphql-hive/router/commit/94684e1d5e8696db503d9877e02ec50b52109e4d.

bin/router/src/pipeline/header.rs
enisdenjo: added the `accept_header` to the message in https://github.com/graphql-hive/router/commit/94684e1d5e8696db503d9877e02ec50b52109e4d. failing tests would then fail with:
```
wrong agreed response mode when negotiating: text/event-stream;q=0.5, multipart/mixed;q=1;subscriptionSpec="1.0"
```
making sure you can match with cases real quick

bin/router/src/pipeline/mod.rs
dotansimha: yeah, that's 10/10 readability and super clean on how the response will be handled

bin/router/src/pipeline/header.rs
dotansimha: nice

bin/router/src/pipeline/header.rs
enisdenjo: here it is https://github.com/graphql-hive/router/pull/665/commits/90c26571051a17d63fe91a0ef3c7f7c7c7d6dd4c

Cargo.toml
enisdenjo: replied in original comment.

bin/router/src/pipeline/execution.rs
```diff
@@ -65,14 +63,35 @@ pub async fn execute_plan(
...
-    };
+    if matches!(expose_query_plan, ExposeQueryPlanMode::DryRun) {
+        let body = project_by_operation(
+            &Value::Null,
+            vec![],
+            &extensions,
+            planned_request.normalized_payload.root_type_name,
+            &[],
+            &None,
+            1000, // TODO: what would the estimate be here?
```
kamilkisiela: No need to have estimation here, it's purely for performance reasons, to preallocate a chunk of memory.
The dry-run is meant to be for local development, so no need to be smart about pre-alloc here, not worth the deal.

enisdenjo: Set to 0

bin/router/src/pipeline/header.rs
```diff
@@ -1,28 +1,314 @@
...
+            Ok(SingleContentType::JSON)
+        } else {
+            Err("Unsupported single content type")
+        }
+    }
+}
+
+impl SingleContentType {
+    // no consts until https://github.com/picoHz/mediatype/pull/25 lands
+    pub fn media_types() -> Vec<MediaType<'static>> {
```
kamilkisiela: We could do something like:
```
static SINGLE_CONTENT_TYPE_MEDIA_TYPES: Lazy<Vec<MediaType<'static>>> = Lazy::new(|| {
    // first collect the string representations to keep them alive
    // in order to parse them into MediaType instances that _borrow_
    // the items from the vec
    let strs: Vec<&'static str> = SingleContentType::iter().map(|ct| ct.into()).collect();
    strs.iter()
        .map(|s| {
            MediaType::parse(s)
                // SAFETY: we control the strings being parsed here. see the enum variants
                .unwrap()
        })
        .collect()
});
```

+

```
impl SingleContentType {
    pub fn media_types() -> &'static Vec<MediaType<'static>> {
        &SINGLE_CONTENT_TYPE_MEDIA_TYPES
    }
}
```

This way we don't iterate over the enum variants and parse them to media types and allocate a temporary vector for every upstream request. It will happen once.

enisdenjo: very VERY nice, https://github.com/graphql-hive/router/pull/665/commits/d9d7ff7ff4184ec8f6dac581be80cd3edbdf7aee

bin/router/src/pipeline/header.rs
```diff
@@ -1,28 +1,323 @@
...
+        if essence_media_type_string == SingleContentType::GraphQLResponseJSON.as_ref() {
+            Ok(SingleContentType::GraphQLResponseJSON)
+        } else if essence_media_type_string == SingleContentType::JSON.as_ref() {
+            Ok(SingleContentType::JSON)
+        } else {
+            Err("Unsupported single content type")
+        }
+    }
+}
+
```
kamilkisiela, line 56: We could avoid allocating the String by laveraging those individual components https://github.com/picoHz/mediatype/blob/8150f29ff62a0214cb9ee534f9a63ce694723987/src/media_type.rs#L34-L41 to match the media type with the enum variant.

That's one allocation that does not dot affect performance too much, so it's fine to keep as is.

kamilkisiela: I did some numbers and it'd be 4x faster, but we're talking 70-113 ns vs 16-36 ns

enisdenjo: I originally used consts everywhere, which is the most optimal way. But I had to fork the crate, because of [this](https://github.com/picoHz/mediatype/pull/25) since the maintainer does not budge.

On @dotansimha's request I removed the fork dependency and refactored all this in
https://github.com/graphql-hive/router/pull/665/commits/90c26571051a17d63fe91a0ef3c7f7c7c7d6dd4c.

enisdenjo: Do you think we should go back to the fork @kamilkisiela?

enisdenjo: Or just avoid the String alloc?

kamilkisiela: It's fine as is :) there's a way to improve it, but we're talking about 4x speed-up on a small number, so it does not affect the overall performance noticeably anyway

kamilkisiela: if we ever see this being a culprit (I don't think we ever will), we will fight it

enisdenjo: let's not leave anything up to chance. I changed it to avoid allocations, wdyt?

https://github.com/graphql-hive/router/pull/665/commits/da3883568df6547fd28c162d5ecc7f10bf031003

kamilkisiela: Perfect!

PR #672, opened 2026-01-14
refactor(router,executor): less lifetimes and drop unsafe for Output Rewrites

kamilkisiela:
We also now store `job.flatten_node_path.to_string()` in a local variable to avoid calling `to_string` for every error.

### DIFF
--- a/lib/executor/src/execution/plan.rs
+++ b/lib/executor/src/execution/plan.rs
@@ -390,7 +390,7 @@ impl<'exec> Executor<'exec> {
         ctx: &mut ExecutionContext<'exec>,
         job: ExecutionJob<'exec>,
     ) -> Result<(), PlanExecutionError> {
-        let _: () = match job {
+        match job {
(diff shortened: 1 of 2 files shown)


lib/executor/src/execution/plan.rs
```diff
@@ -604,7 +604,7 @@ impl<'exec> Executor<'exec> {
 
         match self.execute_fetch_node(fetch_node, representations).await? {
             ExecutionJob::Fetch(job) => Ok(ExecutionJob::FlattenFetch(FlattenFetchJob {
-                flatten_node_path: &node.path,
+                flatten_node_path: node.path.clone(),
```
ardatan: Actually we can make it `&node.path` and change `FlattenFetchJob.flatten_node_path` to `&'exec FlattenNodePath` to avoid cloning. What do you think?

kamilkisiela: true, thanks

ardatan (approved):

PR #705, opened 2026-01-22
fix(router): Render GraphiQL when accepting text/html with highest q-weight

### LINKED ISSUE #704: No graphiql on router 0.0.33
Just tried out the latest router release 0.0.33 and graphiql won't load.

When navigating to my graphql endpoint in the browser I get

```json
{
  "errors": [
    {
      "message": "Failed to deserialize query parameters",
      "extensions": {
        "code": "INVALID_QUERY_PARAMS"
      }
    }
  ]
}
```

Not sure if this is related to the introspection enable/disable work, nothing in the release notes that this has been included in the release.

Things work as expected on 0.0.32

Thanks
Mike

enisdenjo:
Closes #704

Now the `ResponseMode` introduced at #665 will also consider the `GraphiQL` response mode and negotiate accordingly with browsers that accept both JSON and HTML, but prefer HTML as per the q-weight.

### DIFF
--- a/bin/router/src/lib.rs
+++ b/bin/router/src/lib.rs
@@ -23,7 +23,6 @@ use crate::{
     jwt::JwtAuthRuntime,
     logger::configure_logging,
     pipeline::{
-        error::PipelineError,
         graphql_request_handler,
         header::{RequestAccepts, ResponseMode, TEXT_HTML_MIME},
         usage_reporting::init_hive_user_agent,
@@ -65,8 +64,7 @@ async fn graphql_endpoint_handler(
         // agree on the response content type so that errors can be handled
         // properly outside the request handler.
         let response_mode = match request.negotiate() {
-            Ok(Some(response_mode)) => response_mode,
-            Ok(None) => return PipelineError::UnsupportedContentType.into_response(None),
+            Ok(response_mode) => response_mode,
             Err(err) => return err.into_response(None),
         };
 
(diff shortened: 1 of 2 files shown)


bin/router/src/lib.rs
```diff
@@ -68,11 +65,12 @@ async fn graphql_endpoint_handler(
         // agree on the response content type so that errors can be handled
         // properly outside the request handler.
         let response_mode = match request.negotiate() {
-            Ok(response_mode) => response_mode,
+            Ok(Some(response_mode)) => response_mode,
+            Ok(None) => return PipelineError::UnsupportedContentType.into_response(None),
```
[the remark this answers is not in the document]
enisdenjo: Cheers, fixed.

bin/router/src/lib.rs
```diff
@@ -82,14 +78,6 @@ async fn graphql_endpoint_handler(
             }
         }
 
-        // not a graphiql request and no supported content types
-        let response_mode = match response_mode {
-            Some(mode) => mode,
-            None => {
-                return PipelineError::UnsupportedContentType.into_response(response_mode);
-            }
-        };
```
kamilkisiela, line 91: We don't handle it now?

enisdenjo: We do, right during negotiation.

https://github.com/graphql-hive/router/blob/cfa8a75f7e27766c644579fec4ad4fd8e1895e83/bin/router/src/pipeline/header.rs#L264-L285

PR #716, opened 2026-01-26
fix(release): changesets cascading 

dotansimha:
- [x] changesets cascading
- [x] remove styleguide around changesets 
- [ ] ~pin relevant deps~
- [x] fix issues related to https://github.com/graphql-hive/console/issues/7557

### DIFF
--- a/lib/hive-console-sdk/Cargo.toml
+++ b/lib/hive-console-sdk/Cargo.toml
@@ -10,7 +10,7 @@ homepage = "https://github.com/graphql-hive/router"
 authors = ["The Guild"]
 
 [dependencies]
-graphql-tools =  { path = "../graphql-tools", version = "=0.5.0" }
+graphql-tools =  { path = "../graphql-tools", version = "0.5.0" }
 
 async-trait = { workspace = true }
 thiserror = { workspace = true }


lib/hive-console-sdk/Cargo.toml
```diff
@@ -10,7 +10,7 @@ homepage = "https://github.com/graphql-hive/router"
 authors = ["The Guild"]
 
 [dependencies]
-graphql-tools =  { path = "../graphql-tools", version = "0.5.0" }
+graphql-tools =  { path = "../graphql-tools", version = "=0.5.0" }
```
ardatan: Do we still need to pin this?
I think re-export of graphql-tools, then import the stuff from SDK crate in the Apollo Router plugin instead of graphql_tools already solves the issue. So not sure if pinning is still needed.

On the other hand, pinning graphql_tools here might cause multiple versions of graphql_tools. If there is a crate that has graphql_tools dep with a compatible version like 0.5.1, then cargo will install both 0.5.0 and 0.5.1 in this case.

dotansimha: Yeah good point, I'll revert the pinned dep as the re-export solves it better.

PR #718, opened 2026-01-27
feat(router): support multiple Hive CDN endpoints

ardatan:
Implements https://the-guild.dev/graphql/hive/docs/schema-registry/high-availability-resilence

Closes ROUTER-251 

Note;
`SingleOrMultiple<T>` has been implemented with serde and schemars support to represent fields that can have multiple or single values as we have for `endpoint` in our case.

### DIFF
--- a/.changeset/multiple-endpoints.md
+++ b/.changeset/multiple-endpoints.md
@@ -1,0 +1,20 @@
+---
+router: patch
+config: patch
+---
+
+# Support multiple endpoints in Hive CDN Supergraph config
+
+In order to support a Secondary CDN endpoint for better reliability, the Hive CDN Supergraph configuration has been updated to allow specifying either a single endpoint or multiple endpoints.
+This change enables users to provide a list of CDN endpoints, enhancing the robustness of their supergraph setup.
+
+[Learn more about it in the relevant Hive Console documentation here](https://the-guild.dev/graphql/hive/docs/schema-registry/high-availability-resilence).
+
+```diff
+supergraph:
+    source: hive
+-    endpoint: https://cdn-primary.example.com/supergraph
++    endpoint:
++       - https://cdn-primary.example.com/supergraph
++       - https://cdn-secondary.example.com/supergraph
+```


bin/router/src/supergraph/hive.rs
```diff
@@ -85,19 +86,31 @@ impl SupergraphHiveConsoleLoader {
...
+        match endpoint {
+            HiveConsoleCdnEndpoint::Single(url) => {
+                fetcher_builder = fetcher_builder.add_endpoint(url);
+            }
+            HiveConsoleCdnEndpoint::Multiple(urls) => {
+                for url in urls {
+                    fetcher_builder = fetcher_builder.add_endpoint(url);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist it is fine

e2e/src/hive_cdn_supergraph.rs
```diff
@@ -329,4 +329,69 @@ mod hive_cdn_supergraph_e2e_tests {
...
+        let json_body: Value = from_slice(&test::read_body(resp).await).unwrap();
+        let types_arr = json_body
+            .get("data")
+            .unwrap()
+            .get("__schema")
+            .unwrap()
+            .get("types")
+            .unwrap()
+            .as_array()
+            .unwrap();
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist Now I use snapshots.

bin/router/src/supergraph/hive.rs
```diff
@@ -76,7 +77,7 @@ impl SupergraphLoader for SupergraphHiveConsoleLoader {
 
 impl SupergraphHiveConsoleLoader {
     pub fn try_new(
-        endpoint: String,
+        endpoint: HiveConsoleCdnEndpoint,
```
dotansimha: Any reason here to "know" the `HiveConsoleCdnEndpoint` struct? 
I mean, it can easily just be `endpoints: Vec<String>` and then just call:
```
for url in urls {
   fetcher_builder = fetcher_builder.add_endpoint(url);
}
```

lib/router-config/src/supergraph.rs
```diff
@@ -74,6 +74,13 @@ pub enum SupergraphSource {
     },
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
+#[serde(untagged)]
+pub enum HiveConsoleCdnEndpoint {
```
dotansimha: See my other comment, you can have `impl HiveConsoleCdnEndpoint` with `fn get_endpoints(&self): Vec<String>`

bin/router/src/supergraph/hive.rs
ardatan: I implemented `Into<Vec<String>>` which is better, I guess.

lib/router-config/src/supergraph.rs
ardatan: https://github.com/graphql-hive/router/pull/718#discussion_r2731184094

lib/router-config/src/primitives/single_or_multiple.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+impl<T: JsonSchema> JsonSchema for SingleOrMultiple<T> {
+    fn schema_name() -> Cow<'static, str> {
+        format!("SingleOrMultiple<{}>", T::schema_name()).into()
+    }
+
+    fn json_schema(gen: &mut SchemaGenerator) -> Schema {
+        let schema = json_schema!({
```
dotansimha, line 28: @ardatan can you confirm that this one actually work?

Check the output here: https://github.com/graphql-hive/router/pull/718/changes#diff-0b5ca119d2be595aa307d34512d9679e49186307ef94201e4b3dfa079aa89938R1854 
it seems like it generated empty string?

lib/router-config/src/primitives/single_or_multiple.rs
```diff
@@ -0,0 +1,36 @@
+use std::borrow::Cow;
+
+use schemars::{json_schema, JsonSchema, Schema, SchemaGenerator};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum SingleOrMultiple<T> {
```
dotansimha, line 8: Do we have other places where it might make sense to use this today? 
(not for this PR, it's general question, just to know if we need follow up PRs)

lib/router-config/src/supergraph.rs
dotansimha: Great

lib/router-config/src/primitives/single_or_multiple.rs
ardatan: I just made it generic for future cases. For now, endpoints is the only case. Maybe persisted operations or subgraph fallback endpoints...

lib/router-config/src/primitives/single_or_multiple.rs
ardatan: Yeah it works. But jsonschema2mk doesn't support optional `oneOf`s. You can see it generates the schema as expected(which is enough for validation) here;
https://github.com/graphql-hive/router/actions/runs/21392800332/job/61583600889?pr=718#step:8:35

dotansimha: ok great! that was the actual goal here. thanks!

PR #719, opened 2026-01-27
Fix handling of 4xx client errors for failover in Hive SDK

kamilkisiela:
The 4xx client errors are now properly treated as errors and trigger endpoint failover, instead of being returned as successful responses. This ensures the CDN fallback mechanism works correctly when endpoints return client errors like 403 Forbidden or 404 Not Found.

Previously it was treated as Ok 200 and the parser received an empty string, which caused a handled panic (runtime not affected).

Previous behaviour: https://github.com/graphql-hive/router/actions/runs/21391996460/job/61580904276

### DIFF
--- a/e2e/src/supergraph.rs
+++ b/e2e/src/supergraph.rs
@@ -1,7 +1,11 @@
 #[cfg(test)]
 mod supergraph_e2e_tests {
-    use std::{sync::Arc, time::Duration};
+    use std::{
+        sync::Arc,
+        time::{Duration, Instant},
+    };
 
+    use mockito::Mock;
     use ntex::{time, web::test};
     use sonic_rs::{from_slice, to_string_pretty, JsonValueTrait, Value};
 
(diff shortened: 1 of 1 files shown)


lib/hive-console-sdk/src/supergraph_fetcher/async_fetcher.rs
```diff
@@ -26,21 +26,31 @@ impl SupergraphFetcher<SupergraphFetcherAsyncState> {
...
-                            )),
-                        ));
-                    } else {
-                        Ok(ok_res)
-                    }
-                }
-                resp
+                req.send()
+                    .await
+                    .map_err(SupergraphFetcherError::Network)
```
dotansimha, line 31: The error returned from `.send` is defined as:
> This method fails if there was an error while sending request,
>     /// redirect loop was detected or redirect limit was exhausted.

Should we treat it in a different way than the error that might come up later?

kamilkisiela: No idea, I just fixed the 4xx errors here

Issue #727, opened 2026-01-28
Allow to customize max request body size


dotansimha:
On of our customers asked for this as part of their migration to Hive Router. They are currently using it with

Things to cover in the plan:

* Security
  * How should we implement this without creating a security issue? (for example: if the body is 100MB and the limit is 1MB, we should never read the whole 100MB)
  * DDoS protection
  * Fail fast (mentioned above)
  * [https://en.wikipedia.org/wiki/Zip_bomb](<https://en.wikipedia.org/wiki/Zip_bomb>)
* Performance
  * any impact?
* Interaction with other things
  * Compression (do we check the body before or after compression? both?)
  * multipart/sse/chunk?
* Is there a scenario where body size should be dynamic?
  * if so, think about the cost
* Config structure
  * Default value
    * should we even limit?
    * if so, to what?
  * under `http`? `limits`? somewhere else?
* Error Handling
  * What should be the response type, status code and body?
  * How do we log this so the developer know about this?
* How other routers are doing that? what they use as default?

ardatan:
> How should we implement this without creating a security issue? (for example: if the body is 100MB and the limit is 1MB, we should never read the whole 100MB)

As documented here, we can read the stream by ourselves until it exceeds the limit as shown here https://ntex.rs/docs/request
So in that way, it will fail with PipelineError that has its own message, status code and error code.
```rust
#[web::post("/")]
#[inline]
pub async fn read_body_stream(
    mut body_stream: web::types::Payload,
    shared_state: &RouterSharedState,
) -> Result<Bytes, PipelineError> {// payload is a stream of Bytes objects
    let max_size = shared_state
        .router_config
        .limits
        .max_request_body_size;
    let mut body = BytesMut::new();
    while let Some(chunk) = body_stream.try_next().await? {
        // limit max size of in-memory payload
        if (body.len() + chunk.len()) > max_size {
            return Err(PipelineError::PayloadTooLarge);
        }
        body.extend_from_slice(&chunk);
    }
    Ok(body.freeze())
}
```

This will end up with a response like;
```json
{
    "errors": [
      {
        "message": "Request body size exceeds the limit",
        "extensions": {
          "code": "PAYLOAD_TOO_LARGE"
        }
      }
    ]
  }
```
Or we can use ntex's builtin configuration which probably just returns 413 error

```rust
fn main() {
    let app = App::new().service(
        web::resource("/index.html")
          // limit size of the payload
[...]

PR #729, opened 2026-01-29
feat(router): configuration to limit the HTTP request body

ardatan:
Closes https://github.com/graphql-hive/router/issues/727

Instead of getting the collected bytes, we get the body stream then we read it by ourselves in our own request handler. While reading it, if it exceeds the limit, we stop the pipeline with a proper GraphQL response + 413 HTTP status code.

First it checks whether the content-length header exceeds the limit, then if it exceeds;

```json
{
  "errors": [
    {
      "message": "Content-Length exceeds the maximum allowed size",
      "extensions": {
        "code": "PAYLOAD_TOO_LARGE_CONTENT_LENGTH"
      }
    }
  ]
}
```

After that, if the buffered body chunks exceed the limit;

```json
{
  "errors": [
    {
      "message": "Request body exceeds the maximum allowed size while reading the stream",
      "extensions": {
        "code": "PAYLOAD_TOO_LARGE_BODY_STREAM"
      }
    }
  ]
}
```

New configuration flag to limit the incoming HTTP request body size in the router before parsing the request(JSON etc).

```yaml
limits:
  max_request_body_size: 2MB # Human readable size format
```

By default, this limit is set to 2MB.

### DIFF
--- a/lib/router-config/src/limits.rs
+++ b/lib/router-config/src/limits.rs
@@ -31,7 +31,7 @@ pub struct LimitsConfig {
     /// It is used to prevent too many aliases that could lead to overfetching or DOS attacks.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_aliases: Option<MaxAliasesRuleConfig>,
-    
+
     #[serde(default = "default_max_request_body_size")]
     #[schemars(with = "String")]
     pub max_request_body_size: Size,
--- a/docs/README.md
+++ b/docs/README.md
@@ -1665,6 +1665,13 @@ Configuration for checking the limits such as query depth, complexity, etc.
 |**max\_request\_body\_size**|`string`|Default: `"2 MB"`<br/>||
 |[**max\_tokens**](#limitsmax_tokens)|`object`, `null`|Configuration of limiting the number of tokens in the incoming GraphQL operations.<br/>|yes|
 
+**Example**
+
+```yaml
+max_request_body_size: 2 MB
+
+```
+
 <a name="limitsmax_aliases"></a>
 ### limits\.max\_aliases: object,null
 
@@ -1679,12 +1686,6 @@ It is used to prevent too many aliases that could lead to overfetching or DOS at
 |Name|Type|Description|Required|
 |----|----|-----------|--------|
 |**n**|`integer`|Aliases threshold<br/>Format: `"uint"`<br/>Minimum: `0`<br/>|yes|
-**Example**
-
-```yaml
-max_request_body_size: 2 MB
-
-```
 
 <a name="limitsmax_depth"></a>
 ### limits\.max\_depth: object,null


lib/router-config/src/limits.rs
```diff
@@ -58,3 +64,18 @@ pub struct MaxTokensRuleConfig {
...
+fn deserialize_human_size<'de, D>(deserializer: D) -> Result<usize, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    let size: human_size::Size = s
+        .parse()
+        .map_err(|e| serde::de::Error::custom(format!("Failed to parse human size: {}", e)))?;
+    Ok(size.to_bytes() as usize)
+}
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist fixed

e2e/src/body_limit.rs
```diff
@@ -0,0 +1,34 @@
...
+        let req = init_graphql_request("{ __typename }", None);
+        let resp = test::call_service(&app.app, req.to_request()).await;
+        // Check the status code
+        assert_eq!(resp.status(), ntex::http::StatusCode::PAYLOAD_TOO_LARGE);
+
+        let body = test::read_body(resp).await;
+        let json_body: sonic_rs::Value = sonic_rs::from_slice(&body).unwrap();
+
+        insta::assert_snapshot!(sonic_rs::to_string_pretty(&json_body).unwrap(), @r###"
+        {
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist fixed

e2e/src/body_limit.rs
```diff
@@ -0,0 +1,34 @@
...
+        wait_for_readiness(&app.app).await;
+
+        let req = init_graphql_request("{ __typename }", None);
+        let resp = test::call_service(&app.app, req.to_request()).await;
+        // Check the status code
+        assert_eq!(resp.status(), ntex::http::StatusCode::PAYLOAD_TOO_LARGE);
+
+        let body = test::read_body(resp).await;
+        let json_body: sonic_rs::Value = sonic_rs::from_slice(&body).unwrap();
+
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist fixed

lib/router-config/src/limits.rs
```diff
@@ -58,3 +64,18 @@ pub struct MaxTokensRuleConfig {
     /// Tokens threshold
     pub n: usize,
 }
+
+fn deserialize_human_size<'de, D>(deserializer: D) -> Result<usize, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    let size: human_size::Size = s
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist fixed

bin/router/src/pipeline/body_read.rs
```diff
@@ -0,0 +1,24 @@
...
+pub async fn read_body_stream(
+    mut body_stream: web::types::Payload,
+    shared_state: &RouterSharedState,
+) -> Result<Bytes, PipelineError> {
+    let max_size = shared_state.router_config.limits.max_request_body_size;
+    let mut body = BytesMut::new();
+    while let Some(chunk) = body_stream.try_next().await? {
+        // limit max size of in-memory payload
+        if (body.len() + chunk.len()) > max_size {
+            return Err(PipelineError::PayloadTooLarge);
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist fixed

lib/router-config/src/limits.rs
```diff
@@ -58,3 +74,9 @@ pub struct MaxTokensRuleConfig {
     /// Tokens threshold
     pub n: usize,
 }
+
+fn default_max_request_body_size() -> Size {
+    "2MB"
```
kamilkisiela: Why 2MB and not 1MB or 10MB?

bin/router/src/pipeline/body_read.rs
```diff
@@ -0,0 +1,44 @@
...
+pub async fn read_body_stream(
+    req: &HttpRequest,
+    mut body_stream: web::types::Payload,
+    shared_state: &RouterSharedState,
+) -> Result<Bytes, PipelineError> {
+    let max_size = shared_state
+        .router_config
+        .limits
+        .max_request_body_size
+        .to_bytes() as usize;
```
kamilkisiela: Could we just pass the usize directly and not rely on RouterSharedState?

lib/router-config/src/limits.rs
```diff
@@ -58,3 +74,9 @@ pub struct MaxTokensRuleConfig {
...
+        .parse()
+        .expect("Default max request body size should be a valid human size")
```
kamilkisiela: Let's tell the user where the value belongs, something like `limits.max_request_body_size` at least.

bin/router/src/pipeline/error.rs
```diff
@@ -118,6 +118,14 @@ pub enum PipelineError {
     #[error("Subscriptions are not supported over accepted transport(s)")]
     #[strum(serialize = "SUBSCRIPTIONS_TRANSPORT_NOT_SUPPORTED")]
     SubscriptionsTransportNotSupported,
+
+    #[error(transparent)]
+    #[strum(serialize = "PAYLOAD_READ_ERROR")]
+    PayloadReadError(#[from] PayloadError),
```
kamilkisiela: Is it used somewhere?

bin/router/src/pipeline/body_read.rs
```diff
@@ -0,0 +1,44 @@
...
+            .map_err(|_| PipelineError::InvalidHeaderValue(CONTENT_LENGTH))?;
+        let content_length: usize = content_length_str
+            .parse()
+            .map_err(|_| PipelineError::InvalidHeaderValue(CONTENT_LENGTH))?;
+        if content_length > max_size {
+            return Err(PipelineError::PayloadTooLarge);
+        }
+    }
+
+    let mut body = BytesMut::new();
```
kamilkisiela: Maybe based on the content-length we could preallocate?

bin/router/src/pipeline/error.rs
```diff
@@ -118,6 +118,14 @@ pub enum PipelineError {
...
+
+    #[error("Request body size exceeds the limit")]
+    #[strum(serialize = "PAYLOAD_TOO_LARGE")]
+    PayloadTooLarge,
```
kamilkisiela: Maybe we should add a hint whether it's because of Content-Length or streamed body? It makes it easier for people to understand why it even happened.

kamilkisiela: Do we log this error somewhere, and if so, how and with what log level? It would be good for observability, otherwise it may get missed.

bin/router/src/pipeline/error.rs
ardatan: Here https://github.com/graphql-hive/router/blob/199758ea0875bde520a102b30b8f6ec16dbb8958/bin/router/src/pipeline/body_read.rs#L36

lib/router-config/src/limits.rs
ardatan: I checked Apollo, and used their default limit.
https://www.apollographql.com/docs/graphos/routing/security/request-limits

bin/router/src/pipeline/error.rs
ardatan: Like any other PipelineError, here. 
https://github.com/graphql-hive/router/blob/max_body_limit/bin/router/src/lib.rs#L91

ardatan: Did a refactor there. What do you think?

bin/router/src/pipeline/body_read.rs
ardatan: Done 👍

bin/router/src/pipeline/body_read.rs
ardatan: Done 👍

lib/router-config/src/limits.rs
ardatan: Done 👍

lib/router-config/src/limits.rs
dotansimha: 2MB is quite high when it comes to just GraphQL query. I think it can be a good default and it's good to have it as default value to avoid inifinite body being sent to the router.

PR #730, opened 2026-01-29
enhance(router): simplify error handling in the request handler

ardatan:
Using https://ntex.rs/docs/errors#an-example-of-a-custom-error-response

So instead of using `into_response()` etc in the request handler, we just throw the error as is and `web::error::WebResponseError` trait takes care of creating the response.

This will help the code look cleaner so we benefit from `?`, and avoid using `match` for every single error handling.

Realized this while working on this PR; https://github.com/graphql-hive/router/pull/729

Maybe in the next step, we can get rid of extra nested functions for graphql_request_handler.

This is helpful for the plugin system in order to handle all errors in one-place.

### DIFF
--- a/bin/router/src/pipeline/error.rs
+++ b/bin/router/src/pipeline/error.rs
@@ -14,7 +14,7 @@ use ntex::{
     http::ResponseBuilder,
     web::{self, error::QueryPayloadError},
 };
-use sonic_rs::json;
+use serde::Serialize;
 use strum::IntoStaticStr;
 
 use crate::{
@@ -212,6 +212,11 @@ impl From<&PipelineError> for GraphQLError {
     }
 }
 
+#[derive(Serialize)]
+struct FailedExecutionResult {
+    errors: Vec<GraphQLError>,
+}
+
 impl web::error::WebResponseError for PipelineError {
     fn error_response(&self, req: &web::HttpRequest) -> web::HttpResponse {
         // Retrieve the negotiated response mode, defaulting to standard if not set
@@ -247,9 +252,6 @@ impl web::error::WebResponseError for PipelineError {
             }
         };
 
-        res.json(&json!({
-            "errors": errors
-        }))
+        res.json(&FailedExecutionResult { errors })
     }
 }
-


bin/router/src/pipeline/execution.rs
```diff
@@ -118,7 +118,7 @@ pub async fn execute_plan(
         executors: &supergraph.subgraph_executor_map,
         initial_errors: planned_request
             .authorization_errors
-            .into_iter()
+            .iter()
```
dotansimha: nice catch

bin/router/src/pipeline/header.rs
```diff
@@ -278,10 +279,18 @@ impl RequestAccepts for HttpRequest {
...
+            Some(response_mode) => {
+                let response_mode = Arc::new(response_mode);
+                self.extensions_mut().insert(response_mode.clone());
+                Ok(response_mode)
+            }
             None => Err(PipelineError::UnsupportedContentType),
         }
     }
+    #[inline]
+    fn response_mode(&self) -> Option<Arc<ResponseMode>> {
```
dotansimha: can you explain why this is needed?

dotansimha: Is it because we need to be aware of the responsemode when you implement the ntex error trait?

ardatan: Yes 👍

bin/router/src/lib.rs
```diff
@@ -44,72 +45,56 @@ use tracing::{info, warn};
...
+    }
 
-        res
-    } else {
-        warn!("No supergraph available yet, unable to process request");
+    // agree on the response content type so that errors can be handled
+    // properly outside the request handler.
+    let response_mode = request.negotiate()?;
+
+    if response_mode == ResponseMode::GraphiQL.into() {
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist Fixed

bin/router/src/pipeline/header.rs
```diff
@@ -258,12 +268,15 @@ pub trait RequestAccepts {
...
+    fn negotiate(&mut self) -> Result<Arc<ResponseMode>, PipelineError>;
+    /// If the request has already negotiated the response mode, return it.
+    /// Useful in the error to response handler
+    fn response_mode(&self) -> Option<Arc<ResponseMode>>;
 }
 
 impl RequestAccepts for HttpRequest {
     #[inline]
-    fn negotiate(&self) -> Result<ResponseMode, PipelineError> {
+    fn negotiate(&mut self) -> Result<Arc<ResponseMode>, PipelineError> {
```
dotansimha: can we simplify the `RequestAccepts` and avoid making it `mut` here? 

So it will remain `fn negotiate(&self) -> Result<ResponseMode, PipelineError> {` and just return the responsemode thing, then we can `set` it in the calling function and mutate the reqest. 

The current impl feel a bit unexpected as the function negotiates, not negotiates and srores.

dotansimha (changes requested):

bin/router/src/pipeline/header.rs
ardatan: Splitted it! Let me know what you think now?

PR #734, opened 2026-02-02
Tracing pipeline improvements

kamilkisiela:
- Adds a Trace-aware version of `BatchSpanProcessor`.
- Refactors the export pipeline for more reliable delivery
- Normalizes the graphql and http spans naming - the `hive.kind` is now the name of the span.
- Removes the introspection trace filtering (if somebody has introspection disabled in production, it won't be captured anyway and I thought there's no reason to offer such filtering as it downgrades performance and nobody asked for it  - except Dotan :) that I convinced not to have it, as other router does not offer it too).
- Big performance improvements. MEM dropped from 1GB to 90MB while maintaining 8200 rps (with tracing disabled it's 10k rps and 42MB on my machine).
- Added concurrency test for trace ID uniqueness, as it's easy to mess up by creating spans incorrectly.
- I updated `tracing` to latest version, to use `record_all` macro that dispatches record events altogether. Every dispatch has a cost, now we do it in a batch when possible.

### DIFF
--- a/lib/internal/Cargo.toml
+++ b/lib/internal/Cargo.toml
@@ -51,7 +51,6 @@ opentelemetry_sdk = { workspace = true, features = [
     "rt-tokio",
     "logs",
     "spec_unstable_metrics_views",
-    "experimental_async_runtime",
     "experimental_trace_batch_span_processor_with_async_runtime"
 ] }
 opentelemetry-jaeger-propagator = { workspace = true }


lib/internal/src/telemetry/traces/hive_console_exporter.rs
```diff
@@ -374,47 +382,63 @@ impl<E: SpanExporter> HiveConsoleExporter<E> {
...
             };
 
-            trace_id_to_subgraph_names
-                .entry(trace_id)
-                .or_default()
-                .insert(subgraph_name.to_string());
+            let names = subgraph_names_by_trace.entry(trace_id).or_default();
+            if !names.contains(&subgraph_name) {
+                names.push(subgraph_name);
+            }
```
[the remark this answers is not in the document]
kamilkisiela: too big cost of hashing

lib/internal/Cargo.toml
```diff
@@ -49,6 +51,7 @@ opentelemetry_sdk = { workspace = true, features = [
     "rt-tokio",
     "logs",
     "spec_unstable_metrics_views",
+    "experimental_async_runtime",
```
dotansimha: Any reason to use that instead of the other feature that enables just `tokio`? (`rt-tokio`)

lib/router-config/src/telemetry/tracing.rs
```diff
@@ -35,13 +35,6 @@ impl TracingConfig {
 pub struct TracingInstrumentationConfig {
     #[serde(default)]
     pub spans: TracingSpansInstrumentationConfig,
-    /// Whether to create spans for pure introspection queries
-    #[serde(default = "default_introspection_instrumentation")]
-    pub introspection: bool,
```
dotansimha, line 40: good

lib/internal/Cargo.toml
kamilkisiela: https://github.com/open-telemetry/opentelemetry-rust/blob/580633e03a31e73c50b9c9f0d531c2b2ef38dbca/opentelemetry-sdk/src/runtime.rs#L69-L87

we do `rt-tokio` already, I just followed the code above, and thought we need `experimental_async_runtime` as well to enable it fully, but it looks like https://github.com/open-telemetry/opentelemetry-rust/blob/580633e03a31e73c50b9c9f0d531c2b2ef38dbca/opentelemetry-sdk/Cargo.toml#L49C1-L49C9 implicitly enables the async one too

PR #749, opened 2026-02-04
feat: Operation Complexity - Limit Aliases

ardatan:
Accidentially merged, and reverted
https://github.com/graphql-hive/router/pull/746

### DIFF
--- a/bin/router/src/pipeline/validation/max_aliases_rule.rs
+++ b/bin/router/src/pipeline/validation/max_aliases_rule.rs
@@ -1,0 +1,257 @@
+use std::collections::HashMap;
+
+use graphql_tools::{
+    ast::OperationVisitorContext,
+    static_graphql::query::Definition,
+    validation::{
+        rules::ValidationRule,
+        utils::{ValidationError, ValidationErrorContext},
+    },
+};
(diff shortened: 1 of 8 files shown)


bin/router/src/pipeline/validation/max_aliases_rule.rs
```diff
@@ -0,0 +1,257 @@
...
+        let mut alias_count: usize = 0;
+        // Get the alias of the current node if it is a field
+        if let CountableNode::Field(field) = countable_node {
+            if field.alias.is_some() {
+                alias_count = self.check_limit(alias_count + 1)?;
+            }
+        }
+
+        // If it is a node that has selections, iterate over the selection set, and get their number of aliases
+        if let Some(selection_set) = countable_node.selection_set() {
```
dotansimha, line 81: nit: we can consider to have `impl SelectionSet` in this file, and then extend it with `.aliases_count()`.

dotansimha (approved):
LGTM!

bin/router/src/pipeline/validation/max_aliases_rule.rs
ardatan: But we do it nestedly so, I didn't want to get the existing types involved into this specific use case

PR #752, opened 2026-02-06
enhance(executor): refactor plan execution

ardatan:
Closes https://github.com/graphql-hive/router/pull/683
Closes https://github.com/graphql-hive/router/pull/711
Closes https://github.com/graphql-hive/router/pull/681

While working on the plugin system PR, I realized that we should never return an error in the plan execution.
Why? Because in GraphQL, we always return a result. And in case of an error in steps, we add it to `errors` then return those errors in the final result. It doesn't matter where the error happens, it should be always a result with an error.

And I also found out that we handle errors, and put them into `ctx.errors` in different places which is error prone, and we can throw an error which will ignore the entire result, and forwarded directly to the pipeline handler. We don't want that. So I did a refactor that `execute_plan_node` never throws.

And `prepare_job_future` that prepares the job futures to be processed later to return a future of `Result<Job..`. Why? Because the only place we throw an error is the fetch node execution (which is job technically).

While working on `prepare_job_future`, I saw that we have an extra `ConcurrencyScope` struct which is used in parallel node execution. It takes all the parallel jobs, waits for all them to be resolved then processes their results. But there was an issue with that;
image

I did a refactor to use `FuturesUnordered` directly, and process the job results whenever they are resolved without waiting all of them to be resolved. This looks more efficient, no?
However, borrow checker wasn't happy with that. Because `ctx.data` is borrowed in `prepare_job_future` and the borrow stays until the future is resolved. But actually job_future doesn't need that because `ctx.data` is only used to generate `representations` which is done already while job is being resolved. So I did a refactor to let `ctx.data` go in sync while the future is only from the fetch job.

While doing this refactor, I realized that the signature of the future is `Result<Option<Job` but the truth is the result throws only while executing the fetch node, and it is `None` only when it is flatten job, and there is no data except the error possibly thrown by traverse_and_callback's call of project_requires. I checked `project_requires` to see why it can throw. I see that it uses `sonic_rs::to_value` when there is no selection set for a value(probably tha cases when there is a scalar with an object value). I see that we can replace it with the implementation we have in `response` projection (project_without_selection_set), and it doesn't throw. So that, I got rid of `Result<` in traverse_and_callback, project_requires. Also I see that we have an extra `RequestProjectionContext` just for possible_types, I removed that too.

After all those, I changed it to be `Option<BoxFuture<Result<Job` so the future is not generated when there is no job. And there is no error handling during the preparation of the future.

After all those, I did a bit more changes like moving response header handling to the pipeline handler so we don't need to create `HeaderMap` inside plan execution, and apply it in the pipeline handler by directly using the ResponseBuilder. 

Additional notes;
- Now `extensions` is no longer `Option`. So that it would be simpler to handle `extensions` in the plugin system. It should be ok since `extensions` is not serialized if it is empty.
- I did some naming changes like `final_response` to `data` and `initial_errors` to `errors` to make things more clear for the future changes because they refer to `data` and `errors` in the final execution result. `{ data, errors, extensions }`
- I found out that when an error happens in a fetch job of a flatten job, affected path is not correctly reflected. So I added `affected_path` parameter to `prepare_fetch_job` so that we can pass `affected_path` from the flatten job to the fetch job to have the path correctly in the errors. See the diff in `error_handling` e2e test.


lib/executor/src/execution/plan.rs
```diff
@@ -116,422 +122,368 @@ pub async fn execute_query_plan<'exec>(
...
+                                                .or_insert_with(Vec::new);
+                                            error_paths.push(error_path);
+                                        }
+                                        if let Some(entity) = entities.get(*entity_index) {
+                                            // SAFETY: `new_val` is a clone of an entity that lives for `'a`.
+                                            // The transmute is to satisfy the compiler, but the lifetime
+                                            // is valid.
+                                            let new_val: Value<'_> =
+                                                unsafe { std::mem::transmute(entity.clone()) };
+                                            deep_merge(target, new_val);
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist it is ok

lib/router-config/src/lib.rs
```diff
@@ -38,7 +37,7 @@ use crate::{
     traffic_shaping::TrafficShapingConfig,
 };
 
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
+#[derive(Debug, Default, Deserialize, Serialize, JsonSchema)]
```
dotansimha: Any reason `Default` was added here?

e2e/src/error_handling.rs
```diff
@@ -56,7 +56,8 @@ mod error_handling_e2e_tests {
               "message": "Failed to send request to subgraph \"http://0.0.0.0:9876/products\": client error (Connect)",
               "extensions": {
                 "code": "SUBGRAPH_REQUEST_FAILURE",
-                "serviceName": "products"
+                "serviceName": "products",
+                "affectedPath": "me.reviews.@.product"
```
dotansimha: Is this an intended change? I wasn't sure if I can correlate this to a piece of code that was changed?

lib/executor/src/headers/mod.rs
```diff
@@ -353,8 +367,12 @@ mod tests {
         )
         .unwrap();
 
-        let mut final_headers = HeaderMap::new();
-        modify_client_response_headers(accumulator, &mut final_headers).unwrap();
+        let mut response_builder = ntex::http::Response::Ok();
+        accumulator
+            .modify_client_response_headers(&mut response_builder)
+            .unwrap();
```
dotansimha: why the `unwrap` here? I saw you added an dedicated error for ErrorPropagation, isn't that a case for it as well?

lib/executor/src/headers/mod.rs
```diff
@@ -417,8 +435,12 @@ mod tests {
...
```
dotansimha: same regarding `unwrap`

lib/executor/src/execution/plan.rs
```diff
@@ -50,78 +41,78 @@ use crate::{
...
-        ctx.headers_plan,
-        ctx.jwt_auth_forwarding,
-        // Deduplicate subgraph requests only if the operation type is a query
-        ctx.operation_type_name == "Query",
-    );
-
-    if let Some(node) = &ctx.query_plan.node {
-        executor.execute_plan_node(&mut exec_ctx, node).await?;
-    }
+    let errors = opts.errors;
```
dotansimha: nit: i think the name `initial_errors` made some sense here, as you use it as starting point for the errors...

lib/executor/src/execution/plan.rs
```diff
@@ -50,78 +41,78 @@ use crate::{
...
-    if error_count > 0 {
-        ctx.span.record_error_count(error_count);
-        ctx.span
-            .record_errors(|| exec_ctx.errors.iter().map(|e| e.into()).collect());
+    let dedupe_subgraph_requests = opts.operation_type_name == "Query";
+
+    let mut exec_ctx = ExecutionContext::new(query_plan, data, errors);
+    // No need for `new`, it has too many parameters
+    // We can directly create `Executor` instance here
+    let executor = Executor {
```
dotansimha: nit: since `Executor` is mostly being created from `opts`, we can implement `Into` for it, or atleast a creator function like: `Executor::from_opts(opts, dedupe_subgraph_requests)`

lib/executor/src/headers/plan.rs
```diff
@@ -130,3 +130,13 @@ type AggregatedHeader = (HeaderAggregationStrategy, Vec<HeaderValue>);
 pub struct ResponseHeaderAggregator {
     pub entries: HashMap<HeaderName, AggregatedHeader>,
 }
+
+impl ResponseHeaderAggregator {
+    pub fn none_if_empty(self) -> Option<Self> {
```
dotansimha: nit: `none_if_empty` feels too verbose, and could be called `as_option`, or even use `Into<Option>`

e2e/src/error_handling.rs
ardatan: Yes definitely. It is a fix in the error handling. Previously fetch node handler wasn't accepting affected_path and this leads affected path is missing in the fetch node errors coming from a flatten node actually.
https://github.com/graphql-hive/router/blob/refactor-plan-execution/lib/executor/src/execution/plan.rs#L316

lib/executor/src/headers/mod.rs
ardatan: Oh it is in the test code actually. It was already `unwrap()` before;
https://github.com/graphql-hive/router/pull/752/changes/BASE..1292bbb26409d88ccc1176ace9f40cde2ca4abdc#diff-4e5ea7dc48c4151b3747873b39de672d76622e21e94d1fc862dc91f4f53c3b07L357

lib/executor/src/headers/mod.rs
ardatan: Same, it is a test. The previous code also had `unwrap()`.

lib/executor/src/headers/plan.rs
ardatan: `as_option` can be confusing because this really returns `None` if it is `empty`. `as_option` feels like wrapping it with `Some()` without considering its emptiness, no?

lib/executor/src/headers/mod.rs
ardatan: Same, it is a test. The previous code also had unwrap().

lib/router-config/src/lib.rs
ardatan: I wanted to simplify the tests like here instead of adding all the fields whenever the configuration has changes;
https://github.com/graphql-hive/router/pull/752/changes#diff-33c9e06e738dde0f63ef177987a5821d328c78989ca484ed2654a29c3a4a293fR729

PR #768, opened 2026-02-12
Add unique leader fingerprint for inflight spans

kamilkisiela:
Makes `hive.inflight.key` span attribute unique, for better identification of the leader and joiners in a distributed system.


e2e/src/telemetry/hive.rs
```diff
@@ -222,14 +222,17 @@ async fn test_hive_http_export() {
     "
     );
 
+    insta::with_settings!({filters => vec![
```
dotansimha, line 225: i love it

PR #770, opened 2026-02-12
feat(router): metrics 

kamilkisiela:
This PR introduces a full metrics pipeline based on OpenTelemetry. It adds support for emitting histograms, counters and gauges for HTTP and GraphQL operations, exports them through OTLP or Prometheus, and allows fine‑grained control over metrics.

- [Documentation PR](https://github.com/graphql-hive/console/pull/7696)
- [Documentation preview](https://pr-7696.hive-landing-page.pages.dev/docs/router/observability/metrics)

**New dependencies**
The router now depends on `opentelemetry‑prometheus` and `prometheus` crates, and `humantime` for duration parsing.

**Performance considerations**
Observed metrics are collected asynchronously and should not block request processing.

**`Metrics` struct**
aggregates metrics for five domains - HTTP server, HTTP client, GraphQL pipeline, the supergraph loader and the internal caches. Each sub‑component exposes counters and histograms and can be disabled via configuration.

**Meter provider setup**
The Telemetry struct now holds separate `traces_provider`, `metrics_provider` and an optional `PrometheusRuntime`. During initialisation the router builds an OpenTelemetry meter provider based on `telemetry.metrics.*` configuration and sets it as global via `opentelemetry::global::set_meter_provider`. If a Prometheus exporter is configured, a `PrometheusRuntime` is created, either attached to the existing HTTP server (reusing the router port) or detached on a dedicated port.

**Prometheus runtime**
The PrometheusRuntime enum encapsulates two modes:
- `Attached` - metrics are served from the same HTTP server on a configurable path (default `/metrics`).
- `Detached` - if a distinct port is configured, the runtime spawns a dedicated `ntex::HttpServer` with its own lifecycle. I limited the server to a **single worker**.

**Endpoint conflict detection**
Because metrics may live under `/metrics` or another path, the router now encapsulates paths into a `RouterPaths` struct. On start‑up it verifies that GraphQL, health, readiness and Prometheus endpoints do not collide and returns a `RouterInitError::EndpointConflict` if they do.

**Request context storage**
A new `pipeline::request_extensions` module defines small structs stored in `HttpRequest` extensions. These hold the raw body size, the GraphQL operation name and type, and the final response status. It also exposes helper functions that are used by different pipeline stages to read and write data.

## Configuration

Metrics are disabled by default. They are enabled when at least one exporter is present and marked as enabled.
Each exporter entry has a `kind` field (e.g. `otlp` or `prometheus`) and optional settings such as `interval`, `temporality` and `protocol`. A Prometheus exporter can specify a `path` and `port` ("empty" means reuse the router port).

The `telemetry.metrics.instrumentation` section allows tuning histogram aggregation.

A common histogram configuration defines explicit buckets for `bytes` (request/response body sizes) and `seconds` (durations). The default buckets are covered in the docs. We do two buckets, one per unit, as their values and their ranges are completely different.

A per-instrument override lets you disable individual metrics or strip certain attributes.

## Noteworthy changes

- Caches are now stored in `CacheState`. This centralisation makes it easier to manage cache invalidation on supergraph reloads and to register metrics observers.
- Cache reads and writes are now standardised. I used `.entry().or_try_insert_with()` API of moka everywhere, to guarantee that concurrent calls on the same not-existing entry are coalesced into one evaluation. We used to do `get().await` and then `insert().await` in some cases, but now we have minimum cache writes and maximum reusability of just inserted entries.
- I also created `EntryResultHitMissExt` trait that extends Moka's Entry API, to give us insights into cache hits, misses and errors. Useful for metrics and spans attempting to record cache hits/misses.

## Example config

```yaml
telemetry:
  metrics:
    exporters:
      - kind: prometheus       # enable the Prometheus exporter
        enabled: true
        path: /metrics         # override the default path
        port: 6969             # spins up a dedicated http server, running
      - kind: otlp
        endpoint: http://otel‑collector:4318
        protocol: http    # or grpc
        interval: 30s     # push every 30s
        temporality: cumulative
    instrumentation:
      common:
        histogram:
          aggregation: explicit # or exponential
          bytes:
            buckets: [128, 512, 1024, 2048, 4096, 8192, 16384, 32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 3145728, 4194304, 5242880]
            record_min_max: false
          seconds:
            buckets: [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10]
            record_min_max: false
      instruments:
        http_server_request_duration_seconds:
          attributes:
            graphql.operation.name: false    # drop operation name label, as it has high cardinality
```

### DIFF
--- a/audits/package.json
+++ b/audits/package.json
@@ -11,7 +11,7 @@
     "ci:test:graphql-over-http": "start-server-and-test start:test-router http://localhost:4000 test-junit:graphql-over-http"
   },
   "devDependencies": {
-    "graphql-federation-gateway-audit": "graphql-hive/federation-gateway-audit#07d1fb9",
+    "graphql-federation-gateway-audit": "graphql-hive/federation-gateway-audit#bf0aa76",
     "graphql-http": "1.22.4",
     "typescript": "5.9.3",
     "start-server-and-test": "2.1.5"
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
     "audits": {
       "name": "router-audits-checks",
       "devDependencies": {
-        "graphql-federation-gateway-audit": "graphql-hive/federation-gateway-audit#07d1fb9",
+        "graphql-federation-gateway-audit": "graphql-hive/federation-gateway-audit#bf0aa76",
         "graphql-http": "1.22.4",
         "start-server-and-test": "2.1.5",
         "typescript": "5.9.3"
@@ -806,7 +806,9 @@
       }
     },
     "node_modules/@jest/diff-sequences": {
-      "version": "30.0.1",
+      "version": "30.3.0",
+      "resolved": "https://registry.npmjs.org/@jest/diff-sequences/-/diff-sequences-30.3.0.tgz",
+      "integrity": "sha512-cG51MVnLq1ecVUaQ3fr6YuuAOitHK1S4WUJHnsPFE/quQr33ADUx1FfrTCpMCRxvy0Yr9BThKpDjSlcTi91tMA==",
       "dev": true,
       "license": "MIT",
       "engines": {
@@ -823,6 +825,8 @@
     },
     "node_modules/@jest/schemas": {
       "version": "30.0.5",
+      "resolved": "https://registry.npmjs.org/@jest/schemas/-/schemas-30.0.5.tgz",
+      "integrity": "sha512-DmdYgtezMkh3cpU8/1uyXakv3tJRcmcXxBOcO0tbaozPwpmh4YMsnWrQm9ZmZMfa5ocbxzbFk6O4bDPEc/iAnA==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -2157,7 +2161,7 @@
     },
     "node_modules/graphql-federation-gateway-audit": {
       "version": "1.0.0",
-      "resolved": "git+ssh://git@github.com/graphql-hive/federation-gateway-audit.git#07d1fb9aac38af416c9143295a8dc059d7e9220f",
+      "resolved": "git+ssh://git@github.com/graphql-hive/federation-gateway-audit.git#bf0aa76a8a4947159d1616fa6e4d67de86ef28b0",
       "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
@@ -2171,7 +2175,7 @@
         "get-port": "7.1.0",
         "graphql": "16.13.1",
         "graphql-yoga": "5.18.1",
-        "jest-diff": "30.2.0",
+        "jest-diff": "30.3.0",
         "kill-port-process": "4.0.2",
         "wait-on": "9.0.4",
         "yargs": "18.0.0"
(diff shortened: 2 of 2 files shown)


lib/internal/src/http.rs
```diff
@@ -0,0 +1,72 @@
...
+                    // that is not yet stable
+                    "QUERY"
+                } else {
+                    // For telemetry purposes, we log the method as "_OTHER"
+                    "_OTHER"
+                }
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
kamilkisiela: It did compile, so you're wrong

lib/internal/src/telemetry/metrics/http_server_metrics.rs
```diff
@@ -0,0 +1,226 @@
...
+            KeyValue::new(
+                labels::GRAPHQL_RESPONSE_STATUS,
+                graphql_response_status.as_str(),
+            ),
+            KeyValue::new(
+                labels::GRAPHQL_OPERATION_NAME,
+                graphql_operation_name
+                    .map(str::to_string)
+                    .unwrap_or_else(|| values::UNKNOWN.to_string()),
+            ),
```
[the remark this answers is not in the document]
kamilkisiela: Exactly, that's why I covered it in docs as "high cardinality" warning

bin/router/src/pipeline/normalize.rs
```diff
@@ -106,15 +101,21 @@ pub async fn normalize_request_with_cache(
...
-                    .insert(cache_key, payload_arc.clone())
-                    .await;
 
-                Ok(payload_arc)
-            }
-        }
+                Ok(Arc::new(payload))
+            })
+            .await
+            .map_err(PipelineError::from)
```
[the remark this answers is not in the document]
kamilkisiela: Not really, compiles with no problems

lib/router-config/Cargo.toml
```diff
@@ -29,6 +29,7 @@ once_cell = "1.21.3"
 
 schemars = "1.0.4"
 humantime-serde = "1.1.1"
+humantime = "2.3.0"
```
dotansimha: move to root Cargo

dotansimha: also used in interna crate

lib/router-config/src/primitives/toggle.rs
```diff
@@ -0,0 +1,90 @@
...
+                }
+            }
+        }
+    }
+}
+
+#[derive(JsonSchema)]
+#[serde(untagged)]
+#[allow(dead_code)]
+enum ToggleWithSchema<T> {
```
dotansimha, line 70: dead code?

lib/internal/src/background_tasks/mod.rs
```diff
@@ -48,12 +47,10 @@ impl BackgroundTasksManager {
         self.arbiter.spawn(f);
     }
 
-    pub fn shutdown(&mut self) {
+    pub fn shutdown(&self) {
```
dotansimha, line 50: good catch

lib/router-config/Cargo.toml
kamilkisiela: done

PR #772, opened 2026-02-12
enhance(router): hash the schema, validation rules and the operation itself for the validation cache key

ardatan:
This is needed for the plugin system; so that when the validation rules or the public schema is changed, the cache key will be stable.

If a plugin changes the schema or validation rules, the cache key should be generated based on those instead of the existing schema or existing validation rules otherwise it won't be accurate.


bin/router/src/pipeline/validation/mod.rs
```diff
@@ -22,38 +24,42 @@ pub async fn validate_operation_with_cache(
     let validate_span = GraphQLValidateSpan::new();
 
     async {
-        let consumer_schema_ast = &supergraph.planner.consumer_schema.document;
+        let cache_key = {
+            let mut hasher = Xxh3::new();
+            supergraph.planner.consumer_schema.sdl.hash(&mut hasher);
```
dotansimha: I don't think it's a good idea to hash the schema for every operation during validation. This schema could be pretty big, and hashing it in every request could be too heavy. 
`Xxh3` excells in small data inputs (https://github.com/Cyan4973/xxHash#small-data), and adding the full consumer facing, and the validation input codes into the hash and re-do this for every request, seems like too heavy to me. 

Why not calculate a hash for the validation plan when it changes, and calculate hash for consumer schema when it changes? this way it's done only once, and then you hash-the-hash here along with the operation.

ardatan: Did a refactor;
https://github.com/graphql-hive/router/pull/772/commits/df0bc4b10ceb77902dd1751617fe01889076a0a0

bin/router/src/shared_state.rs
```diff
@@ -63,6 +64,7 @@ impl Expiry<String, Arc<JwtTokenPayload>> for JwtClaimsExpiry {
 pub struct RouterSharedState {
     pub validation_plan: ValidationPlan,
     pub parse_cache: Cache<u64, ParseCacheEntry>,
+    pub validate_cache: Cache<u64, Arc<Vec<ValidationError>>>,
```
dotansimha: Any reason to move it here? The reason it was defined in `schema_state` is because we consider validation associated with the schema...

ardatan: Cache key has the schema now, no? If the cache key has the schema then why do we need to invalidate the cache storage per schema?

dotansimha: But if the schema changes, is there a reason to keep that in memory and not clear the cache?

lib/graphql-tools/src/parser/schema/ast.rs
```diff
@@ -414,6 +663,12 @@ pub enum DirectiveLocation {
     VariableDefinition,
 }
 
+impl Hash for DirectiveLocation {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        format!("DirectiveLocation::{}", self.as_str()).hash(state);
```
kamilkisiela: Let's avoid allocation here?

ardatan: Better now?
https://github.com/graphql-hive/router/pull/772/changes/ca45e427a3b14ccee48d1211f9eda50b87f47332

lib/graphql-tools/src/parser/schema/ast.rs
```diff
@@ -519,3 +786,416 @@ impl FromStr for DirectiveLocation {
         Ok(val)
     }
 }
+
+#[cfg(test)]
+mod tests {
```
kamilkisiela: Because of the way xor is used in `hash_list_unordered`, these schemas will have the same hash:

```graphql
directive @tag repeatable on OBJECT
type Query @tag @tag {
  f: String
}
```

and 

```graphql
directive @tag repeatable on OBJECT
type Query {
  f: String
}
```

ardatan: What do you think?
https://github.com/graphql-hive/router/pull/772/commits/4bd6b137342df0ac5438269f084ce6de69ac677f

kamilkisiela: it still does not solve the problem.
`A xor A` = 0

```graphql
directive @a repeatable on OBJECT
directive @b repeatable on OBJECT
directive @c repeatable on OBJECT
directive @d repeatable on OBJECT

type Query @a @a @b @c {
  f: String
}
```

will collide with:

```graphql
directive @a repeatable on OBJECT
directive @b repeatable on OBJECT
directive @c repeatable on OBJECT
directive @d repeatable on OBJECT

type Query @b @c @d @d {
    f: String
}
```

`a ^ a ^ b ^ c = b ^ c` == `b ^ c ^ d ^ d`

kamilkisiela: You can do:
```rust
let mut xor = 0u64;
let mut sum = 0u64;
let mut count = 0u64; // or precalcule it if it's `ExactSizeIterator`

for item in items {
  let mut hasher = Xxh3::new();
  item.hash(&mut hasher);
  let value = hasher.finish();

  xor ^= value;
  sum = sum.wrapping_add(value);
  count = count.wrapping_add(1);
}

let mut out = Xxh3::new();
count.hash(&mut out);
xor.hash(&mut out);
sum.hash(&mut out);
out.finish()
```


or something like that

kamilkisiela: it will be still faster than allocating and sorting

ardatan: Do you think this is ok?
https://github.com/graphql-hive/router/pull/772/commits/3d233ee0d0405a44ca0cc459a263dda4cdd87fd7

kamilkisiela: yes, this will do it

bin/router/src/shared_state.rs
ardatan: Reverted

PR #798, opened 2026-02-26
fix(router): add missing `record_response` in error response handler

ardatan:
Call `record_response` if the response is built by the error handlere

### DIFF
--- a/bin/router/src/lib.rs
+++ b/bin/router/src/lib.rs
@@ -148,7 +148,7 @@ async fn graphql_endpoint_handler(
 
         Ok(res)
     }
-    .instrument(root_http_request_span.clone())
+    .instrument(root_http_request_span.clone().span)
     .await
 }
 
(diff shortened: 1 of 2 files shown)


bin/router/src/lib.rs
```diff
@@ -110,6 +110,9 @@ async fn graphql_endpoint_handler(
         .telemetry_context
         .extract_context(&HeaderExtractor(request.headers()));
     let root_http_request_span = HttpServerRequestSpan::from_request(&request);
+    request
+        .extensions_mut()
+        .insert(root_http_request_span.clone());
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist fixed

bin/router/src/lib.rs
```diff
@@ -144,7 +147,7 @@ async fn graphql_endpoint_handler(
 
         Ok(res)
     }
-    .instrument(root_http_request_span.clone())
+    .instrument(root_http_request_span.clone().span)
```
dotansimha, line 151: I think there's an implementation for `Borrow` there so it should be ok to pass it like that, and the `Borrow` impl is turning it into the nested `span` field?

ardatan: `.instrument` doesn't use `Borrow`. It uses the owned value.

dotansimha (approved):

PR #799, opened 2026-02-26
fix(plugin_system): export `configure_global_allocator` and update examples

dotansimha:


bin/router/src/jwt/jwks_manager.rs
```diff
@@ -24,6 +24,7 @@ impl JwksManager {
     }
 
     pub fn all(&self) -> Vec<Arc<JwkSet>> {
+      let v = Duration
```
ardatan: Is this extra?

dotansimha: oops, cleared!

ardatan (approved):

PR #800, opened 2026-02-26
chore(router): inline documentation for hook payloads

ardatan:
Ref ROUTER-269

I disabled doc testing because it requires to have the entire code in the `///` docs, then it becomes harder to follow.

### DIFF
--- a/lib/executor/src/plugins/hooks/on_execute.rs
+++ b/lib/executor/src/plugins/hooks/on_execute.rs
@@ -27,7 +27,7 @@ pub struct OnExecuteStartHookPayload<'exec> {
     /// The context object that can be used to share data across different plugin hooks for the same request.
     /// It is unique per request and is dropped after the response is sent.
     ///
-    /// [Learn more about the context data sharing in the docs](https://graphql-hive.com/docs/router/extensibility/plugin_system#context-data-sharing)
+    /// [Learn more about the context data sharing in the docs](https://the-guild.dev/graphql/hive/docs/router/extensibility/plugin_system#context-data-sharing)
     pub context: &'exec PluginContext,
     /// The query plan generated for the incoming GraphQL request.
     /// It includes the details of how the router plans to execute the request across the subgraphs.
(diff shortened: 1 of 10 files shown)


lib/executor/src/plugins/hooks/on_execute.rs
```diff
@@ -14,36 +14,100 @@ use crate::response::graphql_error::GraphQLError;
...
+    /// ```
+    ///  let my_header = payload.router_http_request.headers.get("my-header");
+    ///  // do something with the header...
+    ///  payload.proceed()
+    /// ```
     pub router_http_request: &'exec RouterHttpRequest<'exec>,
+    /// The context object that can be used to share data across different plugin hooks for the same request.
+    /// It is unique per request and is dropped after the response is sent.
+    ///
+    /// [Learn more about the context data sharing in the docs](https://graphql-hive.com/docs/router/extensibility/plugin_system#context-data-sharing)
```
dotansimha: I think the link should not be `graphql-hive.com`, but `the-guild.dev/graphql/hive`?

dotansimha: also please confirm on other links

dotansimha (changes requested):

lib/executor/src/plugins/hooks/on_graphql_params.rs
```diff
@@ -16,12 +16,24 @@ use crate::plugin_trait::StartHookResult;
 use ntex::http::Response;
 
 #[derive(Debug, Default, Serialize)]
+/// The GraphQL parameters parsed from the HTTP request body by the router.
+/// This includes the `query`, `operationName`, `variables`, and `extensions`
+/// [Learn more about GraphQL-over-HTTP params](https://graphql.org/learn/serving-over-http/#request-format)
 pub struct GraphQLParams {
     #[serde(skip_serializing_if = "Option::is_none")]
+    /// The GraphQL query string parsed from the HTTP request body by the router
+    /// This contains the source test of a GraphQL query, mutation, or subscription sent by the client in the request body.
```
dotansimha: `test` -> `text`

ardatan: Fixed

lib/executor/src/plugins/hooks/on_execute.rs
ardatan: Fixed

dotansimha (approved):

PR #809, opened 2026-03-02
feat(router, query-planner, executor): add entity alias batching

kamilkisiela:
Introduces batched entity fetches in the query plan. It allows multiple compatible entity fetches (`Flatten(Fetch)`) to be executed through a single `BatchFetch` node, while still keeping incompatible fetches separate (I will explain in detail). 
Compatibility is decided by service name, variable definitions, input and output selection sets, and rewrites, so batching happens only when the resulting plan is correct.

---

**Notes for reviewers**

To understand the behavior, I recommend to look at the tests in `lib/query-planner/src/planner/query_plan/optimize.rs` file.
I tried to cover a lot of cases there, in an easy to read format, for you to review quicker and understand it better, but also for "future me" :)

I also tried to make the diff minimal and avoid touching functions that existed (especially around plan execution :wink:).

Two big files to look closely:
- `lib/query-planner/src/planner/query_plan/optimize.rs` - where the QueryPlan is optimized and Flatten(Fetch) turned into BatchNode)
- `lib/executor/src/execution/plan.rs` - where the execution happens.

I added top-level descriptions to files to explain the algorithm (what and why), and I also added comments with visualisations and examples to make "future me (and you)" happier.

---

For example, if the plan has two compatible entity fetches to `inventory` service, one for `products.@` and one for `topProducts.@`, the planner now produces one BatchFetch alias that serves both paths. In simple terms, the router sends one _entities request and then applies the returned entities to both places in the response. If the two fetches are not compatible, for example because they select different entity fields, the planner still creates one `BatchFetch` node but splits them into different aliases so behavior remains correct.

Here's a visual:

```graphql
Parallel {
  Flatten(path: "products.@") {
    Fetch(service: "inventory") {
      {
        ... on Product {
          upc
        }
      } =>
      {
        ... on Product {
          shippingEstimate
        }
      }
    }
  }
  Flatten(path: "topProducts.@") {
    Fetch(service: "inventory") {
      {
        ... on Product {
          upc
        }
      } =>
      {
        ... on Product {
          shippingEstimate
        }
      }
    }
  }
}
```

is optimized into:

```graphql
BatchFetch(service: "inventory") {
  {
    _e0 {
      paths: [
        "products.@"
        "topProducts.@"
      ]
      {
        ... on Product {
          upc
        }
      }
    }
  }
  {
    _e0: _entities(representations: $__batch_reps_0) {
      ... on Product {
        shippingEstimate
      }
    }
  }
}
```

**Aliasing**

When two entity fetches go to the same subgraph but request different output fields, they are batched into one BatchFetch node with two aliases, not one alias.

```graphql
BatchFetch(service: "inventory") {
  {
    _e0 {
      paths: [
        "products.@"
      ]
      {
        ... on Product {
          upc
        }
      }
    }
    _e1 {
      paths: [
        "topProducts.@"
      ]
      {
        ... on Product {
          upc
        }
      }
    }
  }
  {
    _e0: _entities(representations: $__batch_reps_0) {
      ... on Product {
        shippingEstimate
      }
    }
    _e1: _entities(representations: $__batch_reps_1) {
      ... on Product {
        inStock
      }
    }
  }
}
```

In this example, both aliases share the same requires shape (upc), but the output selection is different (shippingEstimate vs inStock), so the planner creates two aliases (_e0, _e1) inside one batch fetch.

**Errors**

When a batched subgraph response contains errors, the router first routes each error by alias.
In batch operations, subgraph errors are usually scoped to alias paths like `_e0` or `_e1`. The router maps those aliases back to the corresponding batch alias group and rewrites the first path segment to `_entities`, because entity error normalization expects paths in that format. 
After that, each alias-scoped error is expanded to all original response paths that shared the same deduplicated representation hash, so duplicated entities receive consistent error paths. Errors that cannot be matched to any known alias are kept as unmatched and attached at the top level, so they are not lost.

**Minification of the BatchFetch query**

The `BatchFetch` operation document is minified. This keeps the generated operation compact, reduces payload size, and makes hashing/deduplication stable for equivalent operations. They are produced identically to regular `Fetch` nodes.

**Network**

This change aim to reduce number of http requests. In our benchmarked scenario it reduced downstream subgraph requests from `13` to `7` for the same operation. The number of execution waves stays the same, so query depth does not change.
HTTP overhead is much lower, as you can see. This should also reduce pressure on subgraphs, because entities are resolved in one batched subgraph call instead of being resolved across multiple incoming GraphQL requests, where the lack of DataLoader or another caching layer could otherwise cause duplicate resolution work.

### DIFF
--- a/e2e/src/entity_batching.rs
+++ b/e2e/src/entity_batching.rs
@@ -1,0 +1,63 @@
+#[cfg(test)]
+mod entity_batching_e2e_tests {
+    use std::collections::BTreeMap;
+
+    use sonic_rs::JsonValueTrait;
+
+    use crate::testkit::{ClientResponseExt, TestRouter, TestSubgraphs};
+
+    #[ntex::test]
+    async fn measure_bench_operation_subgraph_calls() {
+        let subgraphs = TestSubgraphs::builder().build().start().await;
+
+        let router = TestRouter::builder()
+            .with_subgraphs(&subgraphs)
+            .inline_config(
+                r#"
+                  supergraph:
+                    source: file
+                    path: supergraph.graphql
+                  traffic_shaping:
+                    all:
+                      dedupe_enabled: false
+                  "#,
+            )
+            .build()
+            .start()
+            .await;
+
+        let res = router
+            .send_graphql_request(include_str!("../../bench/operation.graphql"), None, None)
+            .await;
+
+        assert!(res.status().is_success(), "Expected 200 OK");
+
+        let json_body = res.json_body().await;
+
+        assert!(json_body["data"].is_object());
+        assert!(
+            json_body["errors"].is_null(),
+            "expected no GraphQL errors in bench operation response"
+        );
+
+        let mut subgraph_request_counts: BTreeMap<&str, usize> = BTreeMap::new();
+        for subgraph in ["accounts", "inventory", "products", "reviews"] {
+            let request_count = subgraphs
+                .get_requests_log(subgraph)
+                .map_or(0, |requests| requests.len());
+            subgraph_request_counts.insert(subgraph, request_count);
+        }
+
+        insta::assert_snapshot!(
+            sonic_rs::to_string_pretty(&subgraph_request_counts).unwrap(),
+            @r#"
+        {
+          "accounts": 2,
+          "inventory": 2,
+          "products": 2,
+          "reviews": 1
+        }
+        "#
+        );
+    }
+}
--- a/lib/query-planner/src/planner/plan_nodes.rs
+++ b/lib/query-planner/src/planner/plan_nodes.rs
@@ -29,65 +29,76 @@ pub struct QueryPlan {
     pub node: Option<PlanNode>,
 }
 
-impl QueryPlan {
-    pub fn fetch_nodes(&self) -> Vec<&FetchNode> {
-        match self.node.as_ref() {
-            Some(node) => {
-                let mut list = vec![];
-                Self::fetch_nodes_from_node(node, &mut list);
-                list
-            }
-            None => vec![],
-        }
-    }
-
-    fn fetch_nodes_from_node<'a>(node: &'a PlanNode, list: &mut Vec<&'a FetchNode>) {
-        match node {
-            PlanNode::Condition(node) => {
-                if let Some(node) = node.else_clause.as_ref() {
-                    Self::fetch_nodes_from_node(node.as_ref(), list);
-                }
-                if let Some(node) = node.if_clause.as_ref() {
-                    Self::fetch_nodes_from_node(node.as_ref(), list);
-                }
-            }
-            PlanNode::Fetch(node) => {
-                list.push(node);
-            }
-            PlanNode::Sequence(node) => {
-                for child in &node.nodes {
-                    Self::fetch_nodes_from_node(child, list);
-                }
-            }
-            PlanNode::Parallel(node) => {
-                for child in &node.nodes {
-                    Self::fetch_nodes_from_node(child, list);
-                }
-            }
-            PlanNode::Flatten(node) => {
-                Self::fetch_nodes_from_node(&node.node, list);
-            }
-            PlanNode::Subscription(node) => {
-                Self::fetch_nodes_from_node(node.primary.as_ref(), list);
-            }
-            PlanNode::Defer(_) => {
-                unreachable!("DeferNode is not supported yet");
-            }
-        }
-    }
-}
-
 #[allow(clippy::large_enum_variant)]
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(tag = "kind")]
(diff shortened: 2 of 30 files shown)


lib/query-planner/src/planner/plan_nodes.rs
```diff
@@ -90,6 +43,64 @@ pub enum PlanNode {
     Defer(DeferNode),
 }
 
+impl PlanNode {
+    pub fn as_fetch(&self) -> Option<&FetchNode> {
```
dotansimha, line 47: nit: there's a crate for that `as_` functions: https://docs.rs/enum-as-inner/latest/enum_as_inner/

e2e/src/entity_batching.rs
```diff
@@ -0,0 +1,66 @@
...
+                  traffic_shaping:
+                    all:
+                      dedupe_enabled: false
+                  "#,
+            )
+            .build()
+            .start()
+            .await;
+
+        router.wait_for_ready(Some(Duration::from_secs(30))).await;
```
enisdenjo: Testkit router already waits for ready ([unless explicitly disabled](https://github.com/graphql-hive/router/blob/5f17d4b137abdb9c14e00be63fbc670030ba0ec5/e2e/src/testkit/mod.rs#L456-L464))

https://github.com/graphql-hive/router/blob/5f17d4b137abdb9c14e00be63fbc670030ba0ec5/e2e/src/testkit/mod.rs#L692-L700

```suggestion
```

kamilkisiela: thanks, removed the extra wait

lib/query-planner/src/planner/query_plan/optimize.rs
```diff
@@ -0,0 +1,2364 @@
...
+struct PlanOptimizer<'a> {
+    supergraph: &'a SupergraphState,
+}
+
+impl PlanOptimizer<'_> {
+    fn optimize_node(&self, node: PlanNode) -> Result<PlanNode, QueryPlanError> {
+        match node {
+            PlanNode::Fetch(_) | PlanNode::BatchFetch(_) => Ok(node),
+            PlanNode::Flatten(flatten_node) => {
+                assert!(
```
dotansimha: nit: `assert!(` will panic in prod. We do have `QueryPlanError` here is might worth using it.

lib/query-planner/src/planner/query_plan/optimize.rs
```diff
@@ -0,0 +1,2364 @@
...
+        // Otherwise keep the original node
+        optimized_nodes.push(node);
+    }
+
+    Ok(optimized_nodes)
+}
+
+fn partition_by_subgraph(
+    nodes: &[PlanNode],
+) -> Result<HashMap<String, Vec<EntityFetch>>, QueryPlanError> {
```
dotansimha, line 600: nit (or not): this function eventually returns a `HashMap` that points to subgraph and fetches, and since it's using `HashMap`, then the order of the keys might change. 
In terms of functionality, it doesn't really matter, but I think it could theoretically produce 2 query plans that functions the same, but have different order of steps?

dotansimha (approved):

lib/query-planner/src/planner/plan_nodes.rs
kamilkisiela: Oh I even did a macro, but then I thought it's just to reduce amount of code that is written once.

lib/query-planner/src/planner/query_plan/optimize.rs
kamilkisiela: we should at some point just not do node: PlanNode but be stricter.

I will do the Err tomorrow

lib/query-planner/src/planner/query_plan/optimize.rs
kamilkisiela: I think it does not matter, as we replace nodes based on index, from 
```
fn optimize_parallel_node(
    nodes: Vec<PlanNode>, <--
```

lib/query-planner/src/planner/query_plan/optimize.rs
kamilkisiela: @copilot can you create a PR to my PR and use QueryPlanError instead of assertion?

lib/query-planner/src/planner/plan_nodes.rs
dotansimha: yeah not a big deal at all

lib/query-planner/src/planner/query_plan/optimize.rs
dotansimha: nice

lib/query-planner/src/planner/query_plan/optimize.rs
dotansimha: oh, ok, missed that! thanks

PR #831, opened 2026-03-06
feat: HTTP Callback Protocol

enisdenjo:
Ref ROUTER-259

This is a stacked PR based on #620.

Adding support for the [HTTP Callback Protocol](https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/callback-protocol) for GraphQL subscriptions. This enables subgraphs to push subscription data to the router via HTTP POST callbacks instead of maintaining long-lived streaming connections, which scales better for routers with many simultaneous open subscriptions.

# Configuration

```yaml
subscriptions:
  enabled: true
  callback:
    public_url: https://router.example.com:4000/callback
    path: /callback # Optional, defaults to /callback. Must start with /.
    heartbeat_interval: 5s # Optional, defaults to 5s. Set to 0s to disable heartbeats.
    subgraphs: # List of subgraph names that use this protocol
      - reviews
      - notifications
```

- `public_url`: The URL subgraphs will use to reach the router's callback endpoint. Must be reachable by the subgraphs.
- `path`: The route registered on the router's HTTP server for receiving callbacks. Defaults to `/callback`. Must start with `/` (not currently validated - ntex routing will silently fail if it doesn't).
- `heartbeat_interval`: How often subgraphs must send `check` messages to keep the subscription alive. `0s` disables heartbeats.
- `subgraphs`: Which subgraphs use the HTTP Callback protocol. Subgraphs not listed here fall through to WebSocket or HTTP streaming based on other config.

Protocol selection priority: `callback.subgraphs` > `websocket` > HTTP streaming (default).

# Implementation

## lib/executor/src/executors/http_callback.rs

The `HttpCallbackSubgraphExecutor` implementing `SubgraphExecutor`:

- `execute()` returns an error - HTTP Callback is subscription-only, queries/mutations use the regular HTTP executor
- `subscribe()` generates a UUID subscription ID and verifier, sends the subscription request to the subgraph with the callback extension, and returns a stream
- The channel receiver becomes the subscription stream; the callback handler (on the router's HTTP server) sends messages into the channel
- Subscription cleanup happens on stream drop - the entry is removed from `ActiveSubscriptionsMap`, causing future callbacks to get 404

### `ActiveSubscriptionsMap`

A `DashMap<SubscriptionId, ActiveSubscription>` shared between the executor and the callback handler. Each entry holds the verifier (for authentication) and the channel sender.

The map is shared across supergraph reloads - it lives in `SchemaState` and is passed through to each new `SubgraphExecutorMap` on rebuild. This means active subscriptions survive schema changes.

In the future, this map will be used for all subscription protocols (not just HTTP Callback), enabling subscription termination and deduplication.

## bin/router/src/pipeline/http_callback.rs

Exports `handler` which is the HTTP handler for `POST /callback/{subscription_id}`:

- Validates the `subscription-protocol: callback/1.0` header
- Deserializes the callback payload and validates `kind`, `id`, and `verifier`
- Routes by `action`:
  - `check`: Returns 204 with protocol header (no channel message needed)
  - `next`: Extracts the `payload` field, sends it through the channel as bytes for downstream deserialization
    - Gets re-searialized, is a hot path. Figure out a fix.
  - `complete`: Sends through the channel with optional errors, removes the subscription from the active map
- Returns 404 for unknown subscription IDs (per spec, signals termination to the emitter)

## lib/router-config/src/subscriptions.rs

- New `CallbackConfig` struct with `public_url`, `path`, `heartbeat_interval`, and `subgraphs`
- New `SubscriptionProtocol::HTTPCallback` variant
- `get_protocol_for_subgraph` checks callback subgraphs first, then websocket, then falls back to HTTP streaming

## lib/executor/src/executors/map.rs

- `SubgraphExecutorMap` now carries `active_callback_subscriptions: ActiveSubscriptionsMap`
- `register_executor` handles `SubscriptionProtocol::HTTPCallback` by creating an `HttpCallbackSubgraphExecutor`
- `from_http_endpoint_map` accepts and uses the shared `ActiveSubscriptionsMap`

# To subgraph response and error

Improved error handling and subgraph response `From` trait things will be improved upoon in #620.

# TODOs

- [x] router-side heartbeat timeout enforcement (background sweeper task)
- [ ] ~~body size limit on callback endpoint~~
  do we really need it though? we're receiving stuff from subgraphs that should be trusted sources
- [x] avoid triple-pass deserialization of `next` payloads (deserialize → serialize → deserialize)
- [x] e2e test for heartbeat behavior
- [x] e2e test for client disconnection propagating 404 to emitter
- [x] e2e test for invalid verifier rejection
- [x] validate that `callback.path` starts with `/` at config load time

_*Also inherits those from #620._

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1925,6 +1925,7 @@ Configuration for subgraphs using the HTTP Callback protocol.
 |Name|Type|Description|Required|
 |----|----|-----------|--------|
 |**heartbeat\_interval**|`string`|The interval at which the subgraph must send heartbeat messages.<br/>If set to 0, heartbeats are disabled. Defaults to 5 seconds.<br/>Default: `"5s"`<br/>|no|
+|**listen**|`string`, `null`|The IP address and port the router will listen on for subscription callbacks.<br/>When set, the router will start a dedicated HTTP server bound to this address<br/>for receiving callback messages from subgraphs, separate from the main GraphQL server.<br/>When not set, the callback handler is registered on the main server.<br/><br/>Example: `0.0.0.0:4001`<br/>|no|
 |**path**|`string`|The path of the router's callback endpoint.<br/>Must be an absolute path starting with `/`. Defaults to `/callback`.<br/>Default: `"/callback"`<br/>Pattern: `^/`<br/>|no|
 |**public\_url**|`string`|The public URL that subgraphs will use to send callback messages to this router.<br/><br/>Your public_url must match the server address combined with the router's path.<br/>Meaning, if your server is `http://localhost:4000` and the path is `/callback`,<br/>your `public_url` should be `http://localhost:4000/callback`.<br/><br/>Example: `https://example.com:4000/callback`<br/>Format: `"uri"`<br/>|yes|
 |[**subgraphs**](#subscriptionscallbacksubgraphs)|`string[]`|The list of subgraph names that use the HTTP callback protocol.<br/>Default: <br/>|no|


bin/router/src/lib.rs
```diff
@@ -229,10 +242,18 @@ pub fn configure_ntex_app(
...
 ) {
     if let Some(websocket_path) = websocket_path {
         cfg.route(websocket_path, web::get().to(ws_index));
     }
+    if let Some(callback_path) = callback_path {
+        let callback_route = format!(
+            "{}/{{subscription_id}}",
+            callback_path.trim_end_matches('/'),
+        );
+        cfg.route(&callback_route, web::post().to(callback_handler));
```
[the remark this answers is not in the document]
enisdenjo: its safe, we're dealing with subgraphs that are a part of an internal network

bench/subgraphs/graphql_with_subscriptions.rs
```diff
@@ -88,6 +125,88 @@ where
...
+                let req = match GraphQLRequest::<GraphQLRejection>::from_request(req, &()).await {
+                    Ok(req) => req,
+                    Err(err) => return Ok(err.into_response()),
+                };
+
+                let gql_request = req.into_inner();
+
+                let sub_ext = match gql_request.extensions.get("subscription") {
+                    Some(val) => {
+                        let json = serde_json::to_value(val).unwrap();
```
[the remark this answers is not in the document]
enisdenjo: tests

lib/executor/src/executors/http_callback.rs
```diff
@@ -0,0 +1,339 @@
...
+        execution_request: &SubgraphExecutionRequest<'_>,
+        subscription_id: &str,
+        verifier: &str,
+    ) -> Result<Vec<u8>, SubgraphExecutorError> {
+        let mut body = Vec::with_capacity(4096);
+        body.put(FIRST_QUOTE_STR);
+        write_and_escape_string(&mut body, execution_request.query);
+
+        let mut first_variable = true;
+        if let Some(variables) = &execution_request.variables {
```
ardatan, line 122: I'm curious if we can prevent repetitive code here in different places, because we touch here a lot as in the new batching PR(the way we build stringified variables). So in that kind of changes, we will have to change every single executor if we don't have a single place that has this stringifying logic.

enisdenjo: right, I'll check it out

enisdenjo:
> Any reason not to allow multiple transports? does it always have to be one?

Subgraphs usually always choose one transport for subscriptions.

> If that's the case and we can only have one, I'm wondering how that will play when you configure both? how do we validate and not just have priority for the selection?

We could validate the config, good point. I dont think that's pressing at the moment though, but I created a todo.

bin/router/src/schema_state.rs
```diff
@@ -284,3 +313,53 @@ impl BackgroundTask for SupergraphBackgroundLoaderTask {
...
+                _ = token.cancelled() => {
+                    debug!("heartbeat enforcer cancelled, stopping");
+                    return;
+                }
+                _ = ntex::time::sleep(self.heartbeat_interval) => {}
+            }
+
+            let mut timed_out = Vec::new();
+            for entry in self.active_subscriptions.iter() {
+                let last = *entry.value().last_heartbeat.lock().unwrap();
```
kamilkisiela, line 342: What if the latency grows a bit, due to many factors and the heartbeat runs after 200ms instead of 150ms? Maybe we should have a grace period?

lib/router-config/src/lib.rs
```diff
@@ -179,6 +179,13 @@ impl HiveRouterConfig {
                 .unwrap_or_else(|| self.graphql_path())
         })
     }
+
+    pub fn callback_path(&self) -> Option<&str> {
```
kamilkisiela: I think it would be a good idea to run the callback endpoint on a different port than the main http server component.

For the same reasons as in case of prometheus 

cc @dotansimha

dotansimha: Yes, I fully agree here, either run on a different port, or allow users to configure the http listener for that. Since this is considred "internal" (but public), then users might not want to expose it by default along side the main (real public) HTTP service.

lib/router-config/src/subscriptions.rs
```diff
@@ -10,17 +11,76 @@ pub struct SubscriptionsConfig {
...
+    pub callback: Option<CallbackConfig>,
     /// Configuration for subgraphs using WebSocket protocol.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub websocket: Option<WebSocketConfig>,
 }
 
+/// Configuration for the HTTP Callback subscription mode.
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, Default)]
+#[serde(deny_unknown_fields)]
+pub struct CallbackConfig {
```
dotansimha, line 29: Thinking a bit on security aspect:
1- @kamilkisiela mentioned the need to allowing running the listener on another port, to allow the service to become internal and not fully exposed to the public.

2- Apollo's implementation allows to pass `identity` value, which is just an arbitrary string ("key"?) that's required to be passed with the callback payloads, so you can ensure only specific actors that has that value are able to send the callbacks. Any thoughts on that? Should we have a similar setup?

lib/router-config/src/subscriptions.rs
```diff
@@ -10,17 +11,76 @@ pub struct SubscriptionsConfig {
...
+    /// Meaning, if your server is `http://localhost:4000` and the path is `/callback`,
+    /// your `public_url` should be `http://localhost:4000/callback`.
+    ///
+    /// Example: `https://example.com:4000/callback`
+    pub public_url: String,
+    /// The path of the router's callback endpoint.
+    /// Must be an absolute path starting with `/`. Defaults to `/callback`.
+    #[serde(
+        default = "default_callback_path",
+        deserialize_with = "deserialize_absolute_path"
```
dotansimha: nit: we already saw some cases where we need custom deserializators, so we implemented this as primitices that implementes `JsonSchema` and also custom deserializer.  This way we can fail during config parsing, provide a meaningful error message, and also have full YAML validations (with regex). 

I think it makes sense to create a `AbsolutePath` primitice. Check `HttpHeaderName` one for a quick example.

lib/router-config/src/subscriptions.rs
```diff
@@ -10,17 +11,76 @@ pub struct SubscriptionsConfig {
...
+#[serde(deny_unknown_fields)]
+pub struct CallbackConfig {
+    /// The public URL that subgraphs will use to send callback messages to this router.
+    ///
+    /// Your public_url must match the server address combined with the router's path.
+    /// Meaning, if your server is `http://localhost:4000` and the path is `/callback`,
+    /// your `public_url` should be `http://localhost:4000/callback`.
+    ///
+    /// Example: `https://example.com:4000/callback`
+    pub public_url: String,
```
dotansimha: This could be `Url` primitive of `schemars` so it will be fully validated as URL. 

See `url` crate and the `url` feature of https://docs.rs/schemars/latest/schemars/#feature-flags

docs/README.md
```diff
@@ -18,7 +18,7 @@
 |[**override\_subgraph\_urls**](#override_subgraph_urls)|`object`|Configuration for overriding subgraph URLs.<br/>Default: `{}`<br/>||
 |[**plugins**](#plugins)|`object`|Configuration for custom plugins<br/>||
 |[**query\_planner**](#query_planner)|`object`|Query planning configuration.<br/>Default: `{"allow_expose":false,"timeout":"10s"}`<br/>||
-|[**subscriptions**](#subscriptions)|`object`|Configuration for subscriptions.<br/>Default: `{"enabled":false}`<br/>||
+|[**subscriptions**](#subscriptions)|`object`|Configuration for subscriptions.<br/>Default: `{"callback":null,"enabled":false}`<br/>||
```
dotansimha: if null, it can be omitted for serializing with `skip_serialization_if` thing

lib/router-config/src/lib.rs
enisdenjo: But this should be an extra option, not a builtin. Having the callback on a different port incurs a charge on the infrastucture, the users would need to expose multiple ports for example - changing their container, their firewall, their urls (append the port), etc. etc.

TBH, at this moment I wouldnt do this.

lib/router-config/src/subscriptions.rs
enisdenjo: 1. Answered that on the other comment.
2. The HTTP callback protocol has a `verifier` value (which is the agreed identity key). We do the same yes.

lib/router-config/src/lib.rs
dotansimha: I think it does make sense, so let's discuss this a bit more.

I have this scenario in mind:  if you are using K8s, then you could have:
```
Pod:router -> exposes 4000 for /graphql and 4001 for /callback 
Service:router-server -> exposes 4000 and 4001 internally 
Ingress:router -> exposes router-server:4000 only to the public 

Pod:subgraph -> uses `router-server:4001/callback` for HTTP callbacks 
```

This is a super popular setup where you use K8s internals for expose ports (or not) through the `Service` internally, and then through `Ingress` externally. But using this setup, you can have some ports as "internal" and other containers in your cluster can communicate with the internal ports as well.

For the scenario of the callbacks, the Router and the subgraphs can run on the same cluster, and the subgraph can safely notify the router, without exposing the full router path to the public. 

While some setups allow to filter ingress by path (so you can expose only `/graphql` and not `/callback`, even if they are on the same port), not all setups have that, like in case of Docker. 
In a Docker setup, a user would expose only 1 port (the public one) and leverage `network:` of Docker to allow subgraph->router calls without exposing the callback endpoints publicly.

lib/router-config/src/subscriptions.rs
dotansimha: Ah, I might have missed something here then, where do I specify that value? I saw it's in the payload/message, but i guess it needs to be known by the router as well and configured somehow?

bin/router/src/schema_state.rs
enisdenjo: added a grace period of 500ms

docs/README.md
enisdenjo: done

lib/router-config/src/subscriptions.rs
enisdenjo: looking good https://github.com/graphql-hive/router/pull/831/commits/21836c2d0d1384cc3fe5a98d6c3282e9eb6396d0

lib/router-config/src/subscriptions.rs
enisdenjo: https://github.com/graphql-hive/router/pull/831/commits/b7067dadd9ff65d469606f587e59ef0aee32b27c

lib/router-config/src/subscriptions.rs
enisdenjo: It's generated for each subscription: https://github.com/graphql-hive/router/blob/e48437e24debac102adf6e698dd2ea9d9b9f281b/lib/executor/src/executors/http_callback.rs#L213

dotansimha: ah ok, i got it now! thanks!

lib/router-config/src/subscriptions.rs
dotansimha: perfect

lib/router-config/src/subscriptions.rs
dotansimha: NICE

lib/router-config/src/lib.rs
enisdenjo: done in https://github.com/graphql-hive/router/pull/831/commits/c4d09da2e59f863958bae518bd2a22b51062bd88

bin/router/src/pipeline/http_callback.rs
```diff
@@ -0,0 +1,194 @@
...
+};
+use hive_router_plan_executor::response::graphql_error::GraphQLError;
+use ntex::util::Bytes;
+use ntex::web::{self, types::Path, HttpRequest, HttpResponse};
+use serde::Deserialize;
+use tracing::{debug, trace, warn};
+
+#[derive(Debug, Deserialize)]
+struct CallbackPayload<'a> {
+    kind: String,
```
dotansimha: I saw we validate this to always be `"subscription"`, which makes sense. If you wish to offload the validation from the code, you can use `serde_constant` to declare a constant value, and then serde will do this validation while deserializing.

bin/router/src/pipeline/http_callback.rs
```diff
@@ -0,0 +1,194 @@
...
+    action: String,
```
dotansimha: `action` is a predefined set, right? of next/complete/.. 

should this be `enum` + `strum`?

bin/router/src/pipeline/http_callback.rs
```diff
@@ -0,0 +1,194 @@
...
+            SUBSCRIPTION_PROTOCOL_HEADER, CALLBACK_PROTOCOL_VERSION
+        );
+        return Err(bad_request());
+    }
+
+    Ok(())
+}
+
+fn parse_payload(body: &Bytes) -> Result<CallbackPayload<'_>, HttpResponse> {
+    serde_json::from_slice(body).map_err(|e| {
```
dotansimha: Any reason to use `serde_json` + `Deserlize` from serde on the `CallbackPayload`? 

We use `sonic_rs` in most places where it comes to JSON, and I wonder if we should align this as well?

bin/router/src/pipeline/http_callback.rs
```diff
@@ -0,0 +1,194 @@
...
+        warn!("Failed to parse callback payload: {}", e);
```
dotansimha: Since the sender needs to be internal and we assume the actor/sender are: 1- valid, 2-known, 3-aware of the spec, should this be printed as `error!` to it will pop up that something is off?

bin/router/src/pipeline/http_callback.rs
```diff
@@ -0,0 +1,194 @@
...
+
+pub async fn handler(
+    req: HttpRequest,
+    path: Path<String>,
+    body: Bytes,
+    active_subscriptions: web::types::State<ActiveSubscriptionsMap>,
+) -> HttpResponse {
+    let subscription_id_from_path = path.into_inner();
+
+    if let Err(response) = validate_protocol(&req) {
```
dotansimha: (very) nit-picking (so you can ignore this if you want): 
you can use `?` to process the entire pipeline in a `{ ... }` scope that returns `Result<...>` and then do `if let Error(response)` only once. 

Again, very nitpicking, very preference-based 😆

bin/router/src/pipeline/http_callback.rs
```diff
@@ -0,0 +1,194 @@
...
+    };
+
+    if let Err(response) = validate_payload(&payload, &subscription_id_from_path) {
+        return response;
+    }
+
+    let subscription = match active_subscriptions.get(&payload.id) {
+        Some(sub) => sub,
+        None => {
+            debug!(
```
dotansimha: I feel like this might need to be elevated from debug to warn? thoughts?

bin/router/src/pipeline/http_callback.rs
enisdenjo: because I used serde_json::value::RawValue to avoid re-serialization and pass the bytes through, but I just found out that sonic_rs::value::LazyValue is the direct analog to it...

fixed in https://github.com/graphql-hive/router/pull/831/commits/3ec1827ddf1c922303d441bcb9f2ab385c22acdc

bin/router/src/schema_state.rs
```diff
@@ -42,6 +48,7 @@ pub struct SchemaState {
     pub plan_cache: Cache<u64, Arc<QueryPlan>>,
     pub validate_cache: Cache<u64, Arc<Vec<ValidationError>>>,
     pub normalize_cache: Cache<u64, Arc<GraphQLNormalizationPayload>>,
+    pub active_callback_subscriptions: ActiveSubscriptionsMap,
```
dotansimha, line 51: just confirming: `active_callback_subscriptions` is part of `SchemaState` because we want to clear/reset it when the schema changes?

enisdenjo: no, because we use it both in the callback subgraph executor and in the callback handler in the router. but in the future, yes, it will be used for both deduplication and schema changes

lib/router-config/src/lib.rs
dotansimha: thanks! 🙇

bin/router/src/pipeline/http_callback.rs
enisdenjo: https://github.com/graphql-hive/router/pull/831/commits/b922b8d6d8eb86fa7c05a3aea0fa1d683b83650d

bin/router/src/schema_state.rs
dotansimha: Gotcha, I was asking because we split out the stuff from the global state into the `SchemaState` so we can have full control over the things that depends on the schema, and manage the lifecycle of the schema in one place. 

Since we are going to use it for that purpose if the future, it's ok to put it here now. Anything the current impl do when the schema changes? or we'll take care of it all after?

bin/router/src/pipeline/http_callback.rs
dotansimha: nice, thanks

bin/router/src/pipeline/http_callback.rs
enisdenjo: https://github.com/graphql-hive/router/pull/831/commits/8c9924af397ab69e4fed7c6587c6993957a274e7

bin/router/src/pipeline/http_callback.rs
enisdenjo: https://github.com/graphql-hive/router/pull/831/commits/8c9924af397ab69e4fed7c6587c6993957a274e7

p.s. couldnt use serde-constant because it doesn't support string const generics

bin/router/src/pipeline/http_callback.rs
enisdenjo: you're right, I keep thinking (wrongly) public subgraphs... https://github.com/graphql-hive/router/pull/831/commits/6f5999d3d0d97f6e982eb286f95cf8c2d1c1821e

bin/router/src/pipeline/http_callback.rs
ardatan: I think we should always avoid error handling in different places. It is easier to have the error handling in a single place on top.

dotansimha: Yeah, I fully agree with that take! 
It's just that all of these functions are returning `Result<String, Response>` and it's tempting to just align them all into a single scope where you do `Result<TheFinalThingIWant, Response>` 😆

enisdenjo: resolved in the dms, current is simpler

ardatan: My point is a bit more than that actually. Pipeline-related things that are converted to HttpResponses should always be `Result<..., PipelineError>` then `handle_pipeline_error` should be only responsible function to convert them into HttpResponse.
Because error handling is constantly growing with logging,metrics, tracing, plugin hooks invocation etc. So if we have all those in one-place, we won't have to look at every single place one -by-one to see where all those come from.

bin/router/src/schema_state.rs
enisdenjo: atm does nothing, none of the transports do... I have a todo in the subs pr though. I'll try doing it, if it becomes too much, lets leave it for an upcoming PR

bin/router/src/pipeline/http_callback.rs
enisdenjo: this is not a part of the pipeline, so no pipeline error, the handler is just (none of the rest have pipeline errors either)

```rs
pub async fn handler(
    req: HttpRequest,
    path: Path<String>,
    body: Bytes,
    active_subscriptions: web::types::State<ActiveSubscriptionsMap>,
) -> HttpResponse {
    let subscription_id_from_path = path.into_inner();

    if let Err(response) = validate_protocol(&req) {
        return response;
    }

    let payload = match parse_payload(&body) {
        Ok(p) => p,
        Err(response) => return response,
    };

    if let Err(response) = validate_payload(&payload, &subscription_id_from_path) {
        return response;
    }

    let subscription = match active_subscriptions.get(&payload.id) {
        Some(sub) => sub,
        None => {
            warn!(
                subscription_id = %payload.id,
                "Subscription not found, may have been terminated"
            );
            return not_found();
        }
    };

    if subscription.verifier != payload.verifier {
        warn!(subscription_id = %payload.id, "Invalid verifier for subscription");
        return bad_request();
    }

    match payload.action {
        CallbackAction::Check => handle_check(&payload.id, &subscription),
        CallbackAction::Next => {
            handle_next(&payload.id, &payload, subscription, &active_subscriptions)
        }
        CallbackAction::Complete => {
            handle_complete(&payload.id, &payload, subscription, &active_subscriptions)
        }
    }
}
```

ardatan: I meant something like this https://github.com/graphql-hive/router/compare/denis-http-callbacks...denis-http-callbacks-arda-error?expand=1
Instead of `if Err(response) return response` blocks.

bin/router/src/pipeline/http_callback.rs
```diff
@@ -0,0 +1,201 @@
...
+fn validate_payload(
+    payload: &CallbackPayload<'_>,
+    subscription_id_from_path: &str,
+) -> Result<(), HttpResponse> {
+    if payload.id != subscription_id_from_path {
+        warn!(
+            "Subscription ID mismatch: path='{}', body='{}'",
+            subscription_id_from_path, payload.id
+        );
+        return Err(bad_request());
```
dotansimha: I think Apollo's spec mentions that this needs to be 404 and not 400? 

> 
id | The identifier for the message's associated subscription, generated by Router during protocol initialization.Router should respond with a 404 status code if this value does not match the ID of an active subscription.

bin/router/src/schema_state.rs
```diff
@@ -284,3 +313,57 @@ impl BackgroundTask for SupergraphBackgroundLoaderTask {
...
+                    return;
+                }
+                _ = ntex::time::sleep(self.heartbeat_interval) => {}
+            }
+
+            let mut timed_out = Vec::new();
+            for entry in self.active_subscriptions.iter() {
+                let last = *entry.value().last_heartbeat.lock().unwrap();
+                if Instant::now().duration_since(last)
+                    > self.heartbeat_interval +
```
dotansimha, line 344: `last` is `last_heartbeat`, right? when you initialize it, you already set the value to now +  `heartbeat_interval`. then here you add it when calculating, so you add `heartbeat_interval` to the last actual time?

bin/router/src/pipeline/http_callback.rs
enisdenjo: good catch https://github.com/graphql-hive/router/pull/831/commits/2f35ac5be4c22d0bee075f1b7ebac7d4708ad32f

bin/router/src/schema_state.rs
```diff
@@ -284,3 +313,57 @@ impl BackgroundTask for SupergraphBackgroundLoaderTask {
...
+            // separate iter so that we dont mess up the slice while looping
+            for id in timed_out {
+                debug!(
+                    subscription_id = %id,
+                    "terminating subscription due to missed heartbeat"
+                );
+                if let Some((_, sub)) = self.active_subscriptions.remove(&id) {
+                    let _ = sub.sender.send(CallbackMessage::Complete {
+                        errors: Some(vec![GraphQLError::from_message_and_extensions(
+                            "Subgraph gone due heartbeat timeout".to_string(),
```
dotansimha: `due to`

bin/router/src/schema_state.rs
enisdenjo: the first now + interval is to not kick off the client on the first check (establishing connection debounces the interval)

the second is grace period, avoid kicking off clients in case of increased latency

tl;dr both are necessary

dotansimha: ah ok!

bin/router/src/schema_state.rs
enisdenjo: https://github.com/graphql-hive/router/pull/831/commits/9c843d8a3589f5bab28917c0a90147e07ac55dd5

PR #844, opened 2026-03-12
Cleanup error handling

ardatan:
- Remove `ParserCacheError` and use `PipelineError`
- Remove `From<Arc<XError>>` implementations in favor of the below
- So that I could remove the extra casts and mappings
- Introduce `PipelineErrorArc` to support `Arc<PipelineError>` result from `or_try_insert_with` calls from parsing, planning, validation and normalization
- Remove unused `ParseCacheEntry.try_new`

### DIFF
--- a/bin/router/src/pipeline/error.rs
+++ b/bin/router/src/pipeline/error.rs
@@ -1,6 +1,6 @@
 use std::{sync::Arc, vec};
 
-use graphql_tools::validation::utils::ValidationError;
+use graphql_tools::{parser::query::MinifyError, validation::utils::ValidationError};
 use hive_router_plan_executor::{
     execution::{error::PlanExecutionError, jwt_forward::JwtForwardingError},
     headers::errors::HeaderRuleRuntimeError,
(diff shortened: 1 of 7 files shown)


bin/router/src/pipeline/error.rs
```diff
@@ -143,30 +143,9 @@ pub enum PipelineError {
...
-            ParserCacheError::MinifyError(err) => {
-                PipelineError::FailedToMinifyParsedOperation(err.clone())
-            }
-            ParserCacheError::ValidationErrors(errs) => {
-                PipelineError::ValidationErrors(errs.clone())
-            }
-        }
-    }
+    #[error(transparent)]
+    PipelineErrorArc(#[from] Arc<PipelineError>),
```
dotansimha, line 148: can you please explain why we need `PipelineErrorArc`? why not put the Arc on higher level?

ardatan: What do you mean?

dotansimha: What part in the code makes us needing this kind or error? 
It's a bit weird to be that we have `PipelineError` and then inside we have a variant named `PipelineErrorArc` that holds `Arc<PipelineError>`, why not just use `Arc<PipelineError>` in the relevant places instead?

ardatan: Inside the pipeline, for example the parser part of the pipeline impl, we have or_try_insert_with that is called if cache miss. But the return value is always Arc even if Error. So when it fails, it returns Arc<PipelineError> in this case. But the actual error type is PipelineError in the flow. So I added this variant to catch Arc ones too.

dotansimha: I wonder if that's really needed in this case. When you have `Arc<Error>`, you can still do `?` on it, so what the reason to have special handling for the `Arc<Error>`? 🤔

ardatan: Arc<Error> is different than Error unfortunately :/ So it is considered as a different kind. That's why i put a variant for it with "from"

PR #888, opened 2026-04-01
Rename GraphiQL to Laboratory everywhere

ardatan:


lib/router-config/src/env_overrides.rs
```diff
@@ -14,9 +14,9 @@ pub struct EnvVarOverrides {
     #[envconfig(from = "LOG_FILTER")]
     pub log_filter: Option<String>,
 
-    // GraphiQL overrides
-    #[envconfig(from = "GRAPHIQL_ENABLED")]
-    pub graphiql_enabled: Option<bool>,
+    // Laboratory overrides
+    #[envconfig(from = "LABORATORY_ENABLED")]
```
kamilkisiela, line 18: @mskorokhodov @ardatan 

Let’s make sure we document that as breaking in the changeset

lib/router-config/src/lib.rs
```diff
@@ -51,9 +51,9 @@ pub struct HiveRouterConfig {
     #[serde(default)]
     pub log: LoggingConfig,
 
-    /// Configuration for the GraphiQL interface.
+    /// Configuration for the Hive Laboratory interface.
     #[serde(default)]
-    pub graphiql: GraphiQLConfig,
+    pub laboratory: LaboratoryConfig,
```
kamilkisiela, line 56: @mskorokhodov @ardatan let’s make sure we document the change in changeset

PR #900, opened 2026-04-08
Circuit Breakers on Subgraph Execution

### LINKED ISSUE #489: Add Circuit Breaker per Subgraph/Endpoint
We should introduce a circuit breaker mechanism per subgraph endpoint to protect the router from cascading failures when a subgraph service starts misbehaving (timeouts, 5xx errors, etc).

Basically, when a subgraph keeps failing or timing out above a certain threshold, the router should trip the circuit for that target - fast-failing requests for a short cooldown period instead of queuing or retrying forever.

Then after cooldown, the circuit goes half-open, sends a few probe requests, and either closes (if healthy) or opens again.

This should be configurable per subgraph and integrated into the router metrics (prom metrics to allow users to observe it) and config system.

Benefits:
- Prevent a single bad subgraph from tanking overall request latency
- Protect connection pools from overload

ardatan:
Closes https://github.com/graphql-hive/router/issues/489
Ref ROUTER-158
Documentation https://github.com/graphql-hive/docs/pull/86
Preview of the feature doc https://b0d94292-hive-platform-docs.theguild.workers.dev/graphql/hive/docs/router/guides/performance-tuning#circuit-breaker
Preview of the API doc https://b0d94292-hive-platform-docs.theguild.workers.dev/graphql/hive/docs/router/configuration/traffic_shaping#circuit_breaker

## Circuit Breaker implementation on Subgraph Execution

Anything is thrown as SubgraphExecutorError is considered as an error by the circuit breaker. It uses the same logic that we have in usage reporting in Hive Console SDK with `recloser` library.

Circuit breaker configuration is available for all subgraphs as a global configuration and subgraph-level configuration.
Besides the implementation, the metrics are also added which is mentioned in the proposal inside the GitHub issue linked above.

What are considered as errors by Circuit Breaker?
- Everything that throws in the subgraph execution today incl. invalid JSON payloads, empty response bodies, connection errors etc.
- In addition, all [500. 502, 503, 504] responses are considered as errors too by default. These status codes can be changed by the user. In that case, the original response is held in case of the circuit breaker throws. Then it is reused to keep the behavior of today. So those responses are not considered as an error by the execution engine itself but only for the circuit breaker.


lib/executor/src/executors/map.rs
```diff
@@ -475,6 +501,57 @@ impl SubgraphExecutorMap {
...
+                }
+            }
+            
+            let circuit_breaker = circuit_breaker.build_async().map_err(|e| {
+                SubgraphExecutorError::CircuitBreakerCreationError(e, subgraph_name.to_string())
+            })?;
+
+            self.circuit_breakers_by_subgraph
+                .insert(subgraph_name.to_string(), circuit_breaker);
+        }
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist Propose a suggestion so I can apply.

lib/router-config/src/traffic_shaping.rs
```diff
@@ -246,3 +258,36 @@ impl Default for TrafficShapingRouterConfig {
...
+    /// Percentage after what the circuit breaker should kick in.
+    /// Default: .5
+    #[serde(default)]
+    pub error_threshold: Option<f32>,
+    /// Count of requests before starting evaluating.
+    /// Default: 5
+    #[serde(default)]
+    pub volume_threshold: Option<usize>,
+    /// After what time the circuit breaker is attempting to retry sending requests in milliseconds.
+    /// Default: 30s
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist Propose a suggestion so I can apply.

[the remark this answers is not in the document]
ardatan: @gemini-code-assist anyway addressed it

lib/executor/src/executors/error.rs
```diff
@@ -53,6 +54,11 @@ pub enum SubgraphExecutorError {
     #[error("Failed to initialize or load native TLS root certificates: {0}")]
     #[strum(serialize = "SUBGRAPH_HTTPS_CERTS_FAILURE")]
     NativeTlsCertificatesError(std::io::Error),
+    #[error("Unable to create circuit breaker: {0} for subgraph \"{1}\"")]
+    CircuitBreakerCreationError(CircuitBreakerError, String),
+    #[error("Rejected by the circuit breaker")]
+    #[strum(serialize = "SUBGRAPH_CIRCUIT_BREAKER_REJECTED")]
+    CircuitBreakerRejected,
```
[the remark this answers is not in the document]
ardatan: @copilot addressed

lib/internal/src/telemetry/metrics/catalog.rs
```diff
@@ -107,6 +107,8 @@ pub mod names {
     pub const PLAN_CACHE_REQUESTS_TOTAL: &str = "hive.router.plan_cache.requests_total";
     pub const PLAN_CACHE_DURATION: &str = "hive.router.plan_cache.duration";
     pub const PLAN_CACHE_SIZE: &str = "hive.router.plan_cache.size";
+    pub const CIRCUIT_BREAKER_REJECTED_REQUESTS: &str =
+        "hive.router.circuit_breaker.rejected_requests";
```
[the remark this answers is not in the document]
ardatan: @copilot addressed

lib/internal/src/telemetry/metrics/circuit_breaker_metrics.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub fn record_rejected_request(&self, subgraph_name: &str) {
+        if let Some(counter) = &self.rejected_requests {
+            counter.add(
+                1,
+                &[KeyValue::new(
+                    labels::SUBGRAPH_NAME,
+                    subgraph_name.to_string(),
+                )],
+            );
+        }
```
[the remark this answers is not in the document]
ardatan: @copilot addressed

docs/README.md
```diff
@@ -3023,9 +3046,28 @@ Optional per-subgraph configurations that will override the default configuratio
...
+The circuit breaker will be triggered based on the error rate of requests to the subgraph, and will attempt to reset after a certain timeout.
+
+
+**Properties**
+
+|Name|Type|Description|Required|
+|----|----|-----------|--------|
+|**enabled**|`boolean`|Enable or disable the circuit breaker for the subgraph.<br/>Default: false (circuit breaker is disabled)<br/>Default: `false`<br/>||
+|**error\_threshold**|`number`, `null`|Percentage after what the circuit breaker should kick in.<br/>Default: .5<br/>Format: `"float"`<br/>||
+|**reset\_timeout**|`string`|After what time the circuit breaker is attempting to retry sending requests in a certain duration.<br/>Default: 30s<br/>||
```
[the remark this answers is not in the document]
ardatan: @copilot addressed

lib/executor/src/executors/map.rs
```diff
@@ -200,9 +207,28 @@ impl SubgraphExecutorMap {
...
+                                recloser::Error::Rejected => {
+                                    // Record circuit breaker rejection in metrics
+                                    self.telemetry_context
+                                        .metrics
+                                        .circuit_breaker
+                                        .record_rejected_request(subgraph_name);
+                                    SubgraphExecutorError::CircuitBreakerRejected
+                                }
+                            })
+                            .await?
```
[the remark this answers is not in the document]
ardatan: @copilot recloser already uses Arc so it is not needed.

ardatan:
@copilot resolve the merge conflicts in this pull request

ardatan:
@copilot resolve the merge conflicts in this pull request

lib/router-config/src/traffic_shaping.rs
```diff
@@ -289,3 +301,38 @@ impl Default for TrafficShapingRouterConfig {
...
+    None
+}
+
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct TrafficShapingSubgraphCircuitBreakerConfig {
+    /// Enable or disable the circuit breaker for the subgraph.
+    /// Default: false (circuit breaker is disabled)
+    #[serde(default = "default_circuit_breaker_enabled")]
+    pub enabled: bool,
```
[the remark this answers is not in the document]
ardatan: @copilot address this

lib/executor/src/executors/map.rs
```diff
@@ -207,9 +214,28 @@ impl SubgraphExecutorMap {
         let mut execution_result = match execution_result {
```
dotansimha, line 252: Please apply the fixes based on our discussion, in regards on how HTTP error status codes are handled.

lib/executor/src/executors/map.rs
```diff
@@ -205,9 +212,28 @@ impl SubgraphExecutorMap {
...
             Some(execution_result) => execution_result,
             None => {
-                executor
-                    .execute(execution_request, timeout, plugin_req_state)
-                    .await?
+                let exec_fut = executor.execute(execution_request, timeout, plugin_req_state);
+                let circuit_breaker = self.circuit_breakers_by_subgraph.get(subgraph_name);
```
dotansimha: If I understand the flow correctly, then `circuit_breakers_by_subgraph` is a `DashMap` and the `get` will return a `Ref`. The description mentions that if you hold that ref for a long time, you might hit a deadlock. 
Here, you are using the value and holding it in the closure while you `await` the `exec_fut` so the circuit breaker is locked while this is running. 

Is that intentional?

lib/executor/src/executors/map.rs
dotansimha, line 352: If I understand correctly, the `subscribe` handler does not have a circuit breaker. Should it have one? I assume it does? 

cc @enisdenjo

dotansimha (changes requested):

lib/executor/src/executors/map.rs
ardatan: Which discussion? What do we agreed on exactly?

lib/router-config/src/traffic_shaping.rs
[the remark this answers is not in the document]
ardatan: @copilot @devin-ai-integration addressed? Confirm please?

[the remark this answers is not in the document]
dotansimha: @ardatan please also add a test to confirm this before

lib/executor/src/executors/map.rs
ardatan: Thanks Copilot but I want to give my answer :D and somehow it pushed the fix before me. So my comment is;
I didn't think it can be an issue because we never need a write lock (concurrently). But you are right about this anti-pattern. So I was planning to fix it by cloning the cheap "Recloser" which uses `Arc` internally. So the fix from Copilot looks good.

lib/executor/src/executors/map.rs
ardatan: It was a question for me Copilot :D Anyway @dotansimha , let's wait for @enisdenjo 's answer. because I am not sure how it should be implemented.

lib/router-config/src/traffic_shaping.rs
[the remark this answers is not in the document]
ardatan: Added it -> https://github.com/graphql-hive/router/commit/74899195771627b7f4d7752bfe1c3048739d599b

lib/executor/src/executors/common.rs
```diff
@@ -19,7 +19,7 @@ pub trait SubgraphExecutor {
         execution_request: SubgraphExecutionRequest<'a>,
         timeout: Option<Duration>,
         plugin_req_state: Option<&'a PluginRequestState<'a>>,
-    ) -> Result<SubgraphResponse<'a>, SubgraphExecutorError>;
+    ) -> Result<SubgraphResponse<'static>, SubgraphExecutorError>;
```
ardatan, line 23: This was already static after the changes in subscriptions PR(see subscribe fn), I made it explicit in order to wrap it in InternalServerError error, so that we can reuse the response below.

lib/executor/src/executors/map.rs
ardatan: As discussed on Slack, I added an option called "error_status_codes" which will be the status codes that are considered as an error by the circuit breaker. And the default status code is 503 unless that option is set by the user

e2e/configs/circuit_breaker_global.router.yaml
```diff
@@ -0,0 +1,13 @@
+# yaml-language-server: $schema=../../router-config.schema.json
+supergraph:
```
kamilkisiela: Can't we have inlined confings instead?

kamilkisiela: I think those are not even used

ardatan: You're right. They are not used, removed!

lib/router-config/src/traffic_shaping.rs
```diff
@@ -329,6 +350,98 @@ pub struct ServerTLSConfig {
...
+    /// Enable or disable the circuit breaker for the subgraph.
+    /// Default: false (circuit breaker is disabled)
+    ///
+    /// When unset on a subgraph-level configuration, the value falls back
+    /// to the value defined in the global (`all`) circuit breaker
+    /// configuration.
+    #[serde(default)]
+    pub enabled: Option<bool>,
+    /// Percentage after what the circuit breaker should kick in.
+    /// Default: 50%
```
kamilkisiela, line 362: Are you sure it's 50% by default? Is it specified by us somewhere or by the library we're using? If it's the library then the defaults may change over time, so imo we should make it explicit.

ardatan: The defaults are set in the `CircuitBreakerBuilder` we use. I am not sure if I can make "default" as part of `serde` annotations because let's say if I add `default` to subgraph config, it will unexpectedly override the one defined in `all`. I can add a test that checks `50%` percentage explicitly.

kamilkisiela:
@ardatan With subgraph request deduplication enabled, one `503` can be cloned to multiple callers, and each caller reports a failure to the same circuit breaker. I'm thinking whether it's fine to do it this way or not... This can open the breaker based on client request concurrency rather than actual servise failure.

lib/executor/src/executors/map.rs
enisdenjo: sorry for the late reply here. the only sensible way to implement the circuit breaker for subscribe is _only_ on the first `Result`. the errors within the stream are already after the router has successfully established a subscription with the subgraph

lib/executor/src/executors/map.rs
```diff
@@ -213,9 +236,61 @@ impl SubgraphExecutorMap {
...
+                    .get(subgraph_name)
+                    .map(|r| r.value().clone());
+                match circuit_breaker {
+                    Some(circuit_breaker) => {
+                        let SubgraphCircuitBreaker {
+                            recloser,
+                            error_status_codes,
+                        } = circuit_breaker;
+                        // Treat configured status codes as errors so the
+                        // circuit breaker can track them. Default: only 503.
```
ardatan: Change the default and support a "joker" syntax
from @dotansimha: default -> [500, 502, 503, 504]
from @kamilkisiela -> 5xx (x as a wildcard)
from @enisdenjo -> [50] == 50x == [500,502, 503, 504, 505]

lib/executor/src/executors/map.rs
```diff
@@ -213,9 +236,61 @@ impl SubgraphExecutorMap {
...
+                                    }
+                                    other_err => Err(other_err),
+                                },
+                                Err(recloser::Error::Rejected) => {
+                                    // Record circuit breaker rejection in metrics
+                                    self.telemetry_context
+                                        .metrics
+                                        .circuit_breaker
+                                        .record_rejected_request(subgraph_name);
+                                    Err(SubgraphExecutorError::CircuitBreakerRejected)
```
ardatan, line 307: from @dotansimha - check how it is shown in the OTEL traces

lib/executor/src/executors/map.rs
```diff
@@ -213,9 +236,61 @@ impl SubgraphExecutorMap {
         let mut execution_result = match execution_result {
             Some(execution_result) => execution_result,
             None => {
-                executor
-                    .execute(execution_request, timeout, plugin_req_state)
-                    .await?
+                let exec_fut = executor.execute(execution_request, timeout, plugin_req_state);
```
ardatan, line 255: from @dotansimha - ensure the cancelled request isn't considered as an error by the circuit breaker)
- First we need to make sure that the cancelled request is not throwing an `Err`

lib/executor/src/executors/map.rs
ardatan: Done!

lib/executor/src/executors/map.rs
ardatan: Added E2Es @dotansimha

lib/executor/src/executors/map.rs
ardatan: Added an E2E @dotansimha

lib/executor/src/executors/map.rs
ardatan: @dotansimha implemented as we decided on!

lib/executor/src/executors/error.rs
```diff
@@ -57,6 +60,12 @@ pub enum SubgraphExecutorError {
     #[error(transparent)]
     #[strum(serialize = "SUBGRAPH_HTTPS_CERTS_FAILURE")]
     TlsCertificatesError(#[from] TlsCertificatesError),
+    #[error("Unable to create circuit breaker: {0} for subgraph \"{1}\"")]
+    #[strum(serialize = "SUBGRAPH_CIRCUIT_BREAKER_CREATION_FAILURE")]
+    CircuitBreakerCreationError(CircuitBreakerError, String),
+    #[error("Rejected by the circuit breaker")]
+    #[strum(serialize = "SUBGRAPH_CIRCUIT_BREAKER_REJECTED")]
+    CircuitBreakerRejected,
```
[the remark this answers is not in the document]
ardatan: @copilot it is ok to add new errors

lib/executor/src/executors/error.rs
```diff
@@ -93,6 +102,9 @@ pub enum SubgraphExecutorError {
     #[error("HTTP Callback protocol configured for subgraph but no callback configuration provided for router")]
     #[strum(serialize = "SUBGRAPH_HTTP_CALLBACK_NOT_CONFIGURED")]
     HttpCallbackNotConfigured,
+    #[error("Subgraph internal server error")]
+    #[strum(serialize = "SUBGRAPH_INTERNAL_SERVER_ERROR")]
+    InternalServerError(Box<SubgraphResponse<'static>>),
```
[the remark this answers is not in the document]
ardatan: @copilot it is ok to add new errors

lib/executor/src/execution/error.rs
```diff
@@ -84,18 +84,18 @@ impl PlanExecutionError {
...
         // Then we move owned Strings directly into builder methods.
         // This way we avoid cloning through Into<String> in those methods.
 
-        if let Some(subgraph_name) = val.context.subgraph_name {
+        if let Some(subgraph_name) = &val.context.subgraph_name {
             error = error.add_subgraph_name(subgraph_name);
         }
-        if let Some(affected_path) = val.context.affected_path {
+        if let Some(affected_path) = &val.context.affected_path {
             error = error.add_affected_path(affected_path);
```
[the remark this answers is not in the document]
ardatan: @copilot update the comment above then

lib/hive-console-sdk/src/circuit_breaker.rs
```diff
@@ -44,6 +52,18 @@ impl CircuitBreakerBuilder {
         self.reset_timeout = timeout;
         self
```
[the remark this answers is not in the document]
ardatan: @copilot update the comment above then

dotansimha (approved):

PR #926, opened 2026-04-16
enhance(plugins): introduce `EarlyHTTPResponse` for a better API to return responses with headers

ardatan:
It is not a breaking change as `PlanExecutionOutput` is still the main "response" type of `on_query_plan` and `on_execute`.

- `end_with_response` now accepts any type that can be converted to `PlanExecutionOutput`, so `EarlyHTTPResponse` is one of them.
- `EarlyHTTPResponse` takes `headers` directly instead of aggregator, so plugins can send headers as `HeaderMap` directly

```rust
payload.end_with_response(
    EarlyHTTPResponse {
        body,
        headers,
        status_code,
    }
);
```

### DIFF
--- a/.changeset/expose-early-http-response.md
+++ b/.changeset/expose-early-http-response.md
@@ -1,0 +1,15 @@
+---
+hive-router-plan-executor: patch
+---
+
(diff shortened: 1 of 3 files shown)


kamilkisiela:
lol, I didn't ask for a review :D I just clicked rebase

PR #927, opened 2026-04-16
Coprocessors

kamilkisiela:
Implements Coprocessor feature to allow external service to intercept and modify GraphQL requests/responses at different stages of the request pipeline.

Design doc: https://guild-oss.slack.com/docs/TAYJ1FSUA/F0ASCC492KS

- [x] context https://github.com/graphql-hive/router/pull/937
- [x] public schema sdl
- [ ] execution stage (future - when demand)
- [ ] subgraph stage (future - when demand)
- [ ] tls (for http/2) (future - when demand)
- [ ] different endpoints per stage (future - when demand)
- [x] benchmark (~11k drops to ~9.2k rps)

---

This PR introduces Coprocessor API. The goal is to let users extend Router in language agnostic way.

## Building blocks
- `Stage` trait that represents a step in the request's pipeline. Every `Stage` implementation defines how to build the http request to coprocessor service, how to parse the response and how to apply mutations. Every stage can `"continue"` or `"break": <status_code>` the flow early.
- `CoprocessorRuntime` wires these stages into one API that is called by Hive Router.
- `RequestContext` - introduced in #937

Every stage has configurable `include` object, so users can opt into the pieces of data they need.
The `include`'s set of fields is different for every stage.

```yaml
include:
  body: true
  headers: true
```

The `context` (that represents `RequestContext`) in all of stages allows users to select what should be passed to the coprocessor service.

```yaml
include:
  context: true # all
  context: ["hive::operation::name"] # only operation's name
```

Currently, we support those stages (they run in this order):
- `router.request`
- `graphql.request`
- `graphql.analysis`
- `graphql.response`
- `router.response`

Those stages are covered in the [design doc](https://guild-oss.slack.com/docs/TAYJ1FSUA/F0ASCC492KS).

The design is forward‑compatible, adding new stages (like execution or subgraph stages) can be added later by implementing the Stage trait.

## Coprocessor service

A coprocessor service can be accessed over http or unix domain socket with http/1, http/2 or h2c protocol.
When using UDS with Router, it's in user's hand to create or truncate the socket file. Hive Router only "talks" to it.

## Observability

I covered metrics, spans and logs around coprocessor runtime.

Metrics:
- `hive.router.coprocessor.requests_total{"coprocessor.stage"}` - Total number of coprocessor requests
- `hive.router.coprocessor.duration{"coprocessor.stage"}` - Duration of coprocessor requests
- `hive.router.coprocessor.errors_total{"coprocessor.stage"}` - Total number of coprocessor errors
- http/uds client of coprocessor api, supports also the `http.client` metrics so users will know about http status codes, responses etc.

Spans:
- `coprocessor` span with `coprocessor.stage` and `coprocessor.id` attributes wraps all stage calls
- `http.client` span wraps http requests to coprocessor service

Logs:
- Every coprocessor response, error, request action has corresponding log line (usually with `debug` level, except `error` for errors) with attributes allowing to correlate coprocessor logs with http logs and graphql logs. There's a link :)


kamilkisiela:
gemini, hold on, it's a draft...

lib/internal/src/expressions/lib.rs
```diff
@@ -105,13 +106,11 @@ impl ExecutableProgram for VrlProgram {
...
 }
 
-/// Generic enum for a value that can be either static or computed via VRL expression
-#[derive(Clone)]
 pub enum ValueOrProgram<T> {
     /// A statically-known value
     Value(T),
     /// A VRL program that computes the value at runtime
-    Program(Box<VrlProgram>),
+    Program(Box<VrlProgram>, ProgramHints),
```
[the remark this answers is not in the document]
kamilkisiela: It will not

bin/router/src/plugins/plugins_service.rs
```diff
@@ -42,9 +62,41 @@ where
...
-            .app_state::<Arc<RouterSharedState>>()
-            .and_then(|shared_state| shared_state.plugins.clone());
+        let shared_state = req.app_state::<Arc<RouterSharedState>>().cloned();
+
+        // Determine if the request should be handled by plugins.
+        // The exceptions are:
+        // - health endpoint
+        // - readiness endpoint
+        // - prometheus endpoint (if it's on the same port)
+        let should_run = {
```
ardatan, line 72: This means plugins' on_http_request hook will never get non GraphQL requests. So we won't be able to cover the cases when the user wants to handle extra endpoints in the plugin system.

plugin_examples/progressive_override_launchdarkly/src/plugin.rs
```diff
@@ -0,0 +1,118 @@
...
+            }
+        }
+
+        if let Ok(mut write) = start_payload.request_context.write() {
+            write
+                .progressive_override()
+                .set_labels_to_override(Some(labels_to_override));
+        }
+
+        start_payload.proceed()
```
ardatan, line 91: Lovely <3 We should definitely have this in the documentation

bin/router/src/plugins/plugins_service.rs
kamilkisiela: it will, it just skips /health /readiness and /metrics

ardatan: Ok ok i

ardatan (approved):

PR #933, opened 2026-04-17
Test HTTP/2 Support and add a flag for H2C to subgraphs

### LINKED ISSUE #340: HTTP versions support (v2)
Currently, we only do HTTP1.

ardatan:
Closes https://github.com/graphql-hive/router/issues/340
Ref ROUTER-118

- Add E2E tests HTTP/2 support for Subgraphs <-> Router and Router <-> Client
- Implement `http2_only` flag to enable H2C for Subgraphs <-> Router


dotansimha (approved):

e2e/src/http2.rs
```diff
@@ -0,0 +1,450 @@
...
+            traffic_shaping:
+                router:
+                    tls:
+                        key_file: "{}"
+                        cert_file: "{}"
+                "#,
+                generated_key_pair.key_file_path, generated_key_pair.cert_file_path
+            ))
+            .build()
+            .start_without_healthcheck()
```
kamilkisiela, line 102: Why?

e2e/src/http2.rs
```diff
@@ -0,0 +1,450 @@
...
+            .expect("Failed to build reqwest client");
+
+        let resp = client
+            .post(&graphql_endpoint)
+            .json(&json!({
+                "query": "{ me { name } }"
+            }))
+            .send()
+            .await
+            .expect("Failed to send request");
```
kamilkisiela, line 242: Why not `client.send` - the one we have for graphql requests?

lib/router-config/src/traffic_shaping.rs
```diff
@@ -93,6 +93,14 @@ pub struct TrafficShapingExecutorSubgraphConfig {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub tls: Option<ClientTLSConfig>,
+
+    /// Enables HTTP/2 cleartext (h2c) for requests to this subgraph.
+    ///
+    /// When enabled, the router will use HTTP/2 prior knowledge to communicate
+    /// with the subgraph over plain HTTP (without TLS).
+    /// This is useful when the subgraph supports HTTP/2 but TLS is not required
+    /// (e.g., in a service mesh or internal network).
```
kamilkisiela: I mean, those are for subgraphs so I don't think we need to mention that. Subgraphs are internal and service mesh == federation.

e2e/src/http2.rs
ardatan: The same reason in TLS tests. When healthcheck is enabled, it uses ntex client without certificates and healthchecks fail.

e2e/src/http2.rs
ardatan: Same reason in TLS tests. I configure a seperate client with certificates and different configuration.

kamilkisiela:
I think there's one issue with how `http2_only` is applied right now.
The implementation it is set on the whole client. That means it also affects `https://` subgraphs.
When `http2_only` is `true`, the client won’t fall back to `HTTP/1.1`. So if any HTTPS subgraph only supports `HTTP/1.1`, requests to that subgraph will fail.

Enabling this for `h2c` can accidentally break `https://` subgraphs.

ardatan:
@kamilkisiela `http2_only` can be enabled for a specific subgraph, too. In that case, it will force HTTP2 even if there is no `https`. Enabling `http2_only` won't accidentally break https subgraphs because http2_only means HTTPS only should be through HTTP2.
https://github.com/graphql-hive/router/pull/933/changes#diff-46985f11da43244857f5ec28715f2dc80397d2bcafc5d77a52cb29e5c6ec4295R361

lib/router-config/src/traffic_shaping.rs
ardatan: Removed that last sentence

kamilkisiela:
> @kamilkisiela `http2_only` can be enabled for a specific subgraph, too. In that case, it will force HTTP2 even if there is no `https`. Enabling `http2_only` won't accidentally break https subgraphs because http2_only means HTTPS only should be through HTTP2. https://github.com/graphql-hive/router/pull/933/changes#diff-46985f11da43244857f5ec28715f2dc80397d2bcafc5d77a52cb29e5c6ec4295R361

My concern is mainly about the global setting and the current wording. Right now the docs frame it as an h2c feature (`http://`), but the implementation also forces `HTTP/2` for `https://` when enabled globally.

What I mean is that it's now either `http/1 | http/2` for `https://` (decided on handshake) and `http/1` for `http://` (always) by default - and when you want do `http2_only: true` to force `http/2` for `http://` it also sets `http/2` only for `https://`.


We could change:
> Enables HTTP/2 cleartext (h2c) for requests to subgraphs. When enabled, the router will use HTTP/2 prior knowledge to communicate with subgraphs over plain HTTP (without TLS). This is useful when subgraphs support HTTP/2 but TLS is not required

To:
> Forces HTTP/2 for subgraph requests. For plain HTTP, this uses h2c. For HTTPS, this requires HTTP/2 and does not fall back to HTTP/1.1.

ardatan:
@kamilkisiela Updated the config description. Let me know if it looks good now.

dotansimha (approved):
LGTM. Let's wait for Kamil's approval and then we can proceed. Thanks!

PR #948, opened 2026-04-28
fix(query-planner): avoid propagating @include/@skip conditions to unconditional fetches

### LINKED ISSUE #949: @Includes and @Skip directives cause unconditional fields to be wrapped around conditional
## Problem

The query planner could incorrectly propagate `@include`/`@skip` conditions across merged fetch steps.
In mixed conditional and unconditional selections, unconditional fields could be wrapped by condition nodes and skipped at runtime.

### Example:

Given an operation like:
```graphql
query ($bool: Boolean) {
  product { # Resolved from Graph "A"
    price # Resolved from Graph "A"
    name @include(if: $bool) # Resolved in Graph "B"
    isCheap # Resolved in Graph "B"
  }
}
```

The Query Planner Generates:
```
Sequence
  Fetch(service: "a") { ... }
  Include(if: $bool) {  - - - - -> Wraps the whole fetch to service B around Include directive
    Flatten(path: "product") {
      Fetch(service: "b") {
        ... on Product { isCheap name }
      }
    }
  }
```

**Problem**: if $bool = false, the entire Include block is skipped, so `isCheap` which is unconditional is incorrectly skipped.

gonzo-32:
# Query Planner Fix: Include/Skip Condition Propagation

Closes: https://github.com/graphql-hive/router/issues/949

## Problem

The query planner could incorrectly propagate `@include`/`@skip` conditions across merged fetch steps.
In mixed conditional and unconditional selections, unconditional fields could be wrapped by condition nodes and skipped at runtime.

### Example:

Given an operation like:
```graphql
query ($bool: Boolean) {
  product { # Resolved from Graph "A"
    price # Resolved from Graph "A"
    name @include(if: $bool) # Resolved in Graph "B"
    isCheap # Resolved in Graph "B"
  }
}
```

The Query Planner Generates:
```
Sequence
  Fetch(service: "a") { ... }
  Include(if: $bool) {  - - - - -> Wraps the whole fetch to service B around Include directive
    Flatten(path: "product") {
      Fetch(service: "b") {
        ... on Product { isCheap name }
      }
    }
  }
```

**Problem**: if $bool = false, the entire Include block is skipped, so `isCheap` which is unconditional is incorrectly skipped.
 
## Root Cause

During fetch step merge, condition propagation was preserved at the step level in cases where type scoping was unclear.
That allowed a condition from a conditional source selection to gate unconditional target selections.

## Fix

Updated query planner condition propagation logic so unconditional fetches are not wrapped in conditional blocks when merging steps.
Conditions are scoped to the relevant source output selections instead of being applied to the whole merged fetch step.
This keeps fields without directives from being incorrectly gated by conditions from other steps.

After the fix the query planner generates:
```
    QueryPlan {
      Sequence {
        Fetch(service: "a") {
          query ($bool:Boolean) {
            product {
              __typename
              price
              id
              ... on Product @include(if: $bool) {
                price
                __typename
                id
              }
            }
          }
        },
        Flatten(path: "product") {
          Fetch(service: "b") {
            {
              ... on Product {
                __typename
                price
                id
              }
            } =>
            ($bool:Boolean) {
              ... on Product {
                isCheap
                ... on Product @include(if: $bool) {
                  name
                }
              }
            }
          },
        },
      },
    },
```

### DIFF
--- a/.changeset/avoid-include-skip-condition-propagation.md
+++ b/.changeset/avoid-include-skip-condition-propagation.md
@@ -1,0 +1,12 @@
+---
+hive-router-query-planner: patch
+hive-router-plan-executor: patch
+hive-router: patch
+---
+
+# Avoid propagating @include/@skip conditions to unconditional fetches
+
+Fixed query planner condition propagation logic to avoid wrapping unconditional fetches
+in conditional blocks when merging steps. This ensures that fields without directives are
+not incorrectly gated by conditions from other steps, allowing for correct execution of
+queries with mixed conditional and unconditional selections.
--- a/lib/query-planner/fixture/tests/simple-include-skip.supergraph.graphql
+++ b/lib/query-planner/fixture/tests/simple-include-skip.supergraph.graphql
@@ -78,6 +78,8 @@ type Product
   skip: Boolean! @join__field(graph: C, requires: "isExpensive")
   neverCalledInclude: Boolean! @join__field(graph: C, requires: "isExpensive")
   neverCalledSkip: Boolean! @join__field(graph: C, requires: "isExpensive")
+  name: String! @join__field(graph: B, requires: "price")
+  isCheap: Boolean! @join__field(graph: B, requires: "price")
 }
 
 type Query @join__type(graph: A) @join__type(graph: B) @join__type(graph: C) {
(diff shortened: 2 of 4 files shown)


kamilkisiela:
Hi @gonzo-32 thanks for your changes, I copied some to this PR https://github.com/graphql-hive/router/pull/953 and fixed additional issues (covered in tests) as I couldn't push to your branch.

gonzo-32:
Thank you @kamilkisiela

PR #951, opened 2026-04-30
fix(query-planner): added missing `isRepeatable` on `type __Directive`

dotansimha:
The router's introspection schema was resolving `isRepeatable`, but it did not appear in the public (consumer) schema, leading to validation errors when introspection schema was executed through Laboratory. 

This change adds the missing `isRepeatable: Boolean!` to `type __Directive`, according to the [GraphQL introspection spec](https://github.com/graphql/graphql-spec/blob/main/spec/Section%204%20--%20Introspection.md).


.changeset/added_missing_isrepeatable_on_type_directive.md
```diff
@@ -0,0 +1,12 @@
+---
+hive-router-query-planner: patch
+hive-router-plan-executor: patch
+hive-router: patch
+node-addon: patch
+---
+
+# Added missing `isRepeatable` on `type __Directive`
```
ardatan, line 8: Should we add tests for that?

dotansimha: sure, done!

ardatan: Do you think we should have tests that is `isRepeatable: true`? Or is it too much? What do you think?

ardatan: @dotansimha ^^ Oh I think you missed this one

dotansimha: yeah good catch, done!

ardatan (approved):

PR #953, opened 2026-05-04
avoid propagating @include/@skip conditions to unconditional fetches

### LINKED ISSUE #949: @Includes and @Skip directives cause unconditional fields to be wrapped around conditional

kamilkisiela:
Continue #948 here (as I have no push rights there) and fix issues with the merging logic.

Closes: #949 and #948

### DIFF
--- a/.changeset/avoid-include-skip-condition-propagation.md
+++ b/.changeset/avoid-include-skip-condition-propagation.md
@@ -1,0 +1,12 @@
+---
+hive-router-query-planner: patch
+hive-router-plan-executor: patch
+hive-router: patch
+---
+
+# Avoid propagating @include/@skip conditions to unconditional fetches
+
+Fixed query planner condition propagation logic to avoid wrapping unconditional fetches
+in conditional blocks when merging steps. This ensures that fields without directives are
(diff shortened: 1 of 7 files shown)


kamilkisiela:
@gemini-code-assist review

PR #954, opened 2026-05-04
Replace Docker-based Redis with mock rust server

kamilkisiela:
Removes `bollard` dependency and e2e Docker test helper. Implements an in-process redis server for `response_cache` plugin tests and switch tests to inline config.

---

We have a test failure from time to time and we rely on Docker yada yada, so I created a tiny redis server for mocking as we really only use it for a single plugin test.


plugin_examples/response_cache/router.config.yaml
```diff
@@ -1,10 +0,0 @@
-# yaml-language-server: $schema=../../router-config.schema.json
-supergraph:
```
ardatan, line 2: I deliberately kept the configuration files under examples, not inline them into the tests.

kamilkisiela: done

ardatan (approved):
I should've done this in this way already. Thanks @kamilkisiela! This makes the relevant E2E way cleaner.

PR #963, opened 2026-05-07
Introduce GraphQL differential testing for monolith and federated benchmarks

kamilkisiela:
Adds a way to perform differential testing (a binary we run) that generates queries from a schema and compares responses from two endpoints.

We'll start with subgraphs available in the repository (`bench/subgraphs`) and a monolithic API (unfortunately in JavaScript - as I had issues with async_graphql). The data is exactly the same, the business logic too.

The tool compares `response.data`, not exact errors - it checks whether errors exists or not.
Mismatches are stored under `failed-tests/case-*` with the generated query, variables, and both endpoint responses.

```mermaid
flowchart LR
  A["Schema"] -->|"parse"| B["QueryGenerator"]
  B -->|"generate randomized queries"| C["Query Cases"]
  C -->|"execute"| D["Baseline Endpoint"]
  C -->|"execute"| E["Candidate Endpoint"]
  D -->|"compare data"| F["Differential Tester"]
  E -->|"compare data"| F
  F -->|"mismatch"| G["failed-tests/case-*"]
  F -->|"match"| H["✅ Pass"]
```

---

Once https://github.com/async-graphql/async-graphql/pull/1827 goes in or the issue is fixed in some other way, we can use rust monolith for testing.


bin/differential/src/main.rs
```diff
@@ -0,0 +1,816 @@
...
+        num_queries = val.parse().unwrap_or(10);
+    }
+
+    println!("Running {} differential queries against:", num_queries);
+    println!("  baseline: {}", baseline_endpoint);
+    println!("  candidate: {}", candidate_endpoint);
+    println!("--------------------------------------------------");
+
+    for i in 0..num_queries {
+        let seed = i as u64 + 42;
```
dotansimha, line 728: If we make the seed more static here, and we use the generator, then some aspects of the generated queries should be deterministic, right? in that case, do you think it worth adding this runner to ci in order to detect issues in PRs?

kamilkisiela: we can, but why not just use it for development, see how it works, improve it a bit and only then go ci?

dotansimha: yeah sure, i'm good with that

dotansimha (approved):

PR #968, opened 2026-05-08
fix #966 without degrading performance

### LINKED ISSUE #966: Valid JSON subgraph response fails to deserialize when object key contains escaped character
## Summary

Hive Router fails to deserialise a valid JSON response from a subgraph when an object key contains an escaped character, for example `\t`, `\n`, `\"`, or similar.

The failure appears to come from deserialising JSON object keys as borrowed `&str`. That works for ordinary keys, but escaped JSON keys may need to be unescaped into an owned string before they can be represented. In that case, requiring a borrowed string rejects valid JSON.

### Example response

```json
{
  "data": {
    "labels": {
      "generic.learnMore.button\t": "Learn more"
    }
  }
}
```

### Outcome

The router fails while deserialising the subgraph response and the error surfaces internally:
```
invalid type: string "generic.learnMore.button\t", expected a borrowed string
```

> Playground

```json
{
  "data": {
    "labels": null
  },
  "errors": [
    {
      "message": "Failed to deserialize subgraph response: invalid type: string \"generic.learnMore.button\\t\", expected a borrowed string at line 1 column 111748\n\n\tutton\\t\":\"\",\"gen\n\t........^.......\n",
      "extensions": {
        "code": "SUBGRAPH_RESPONSE_DESERIALIZATION_FAILURE",
        "serviceName": "labels"
      }
    }
  ]
}
```

### Why this matters

Subgraphs can legally return JSON object keys containing escaped characters. The router should not reject those responses during deserialisation.

kamilkisiela:
Some custom scalars are used as JSON passthrough fields.

For these fields, the router should not try to understand the JSON returned by a subgraph. It should not rebuild it or change it. It should keep the JSON fragment and write it back as JSON.

That is why custom scalar fields marked by the planner are stored as:
```rust
Value::RawJson(Cow<'a, str>)
```

They are not parsed into normal `Value::Object`, `Value::Array`, or scalar variants.

---

**Why this instead of just changing `Object` variant to use 🐮 ?**

One possible fix was to change object keys from `&str` to `Cow<str>` in `Value::Object` enum variant. That fixed this bug, but it made normal response handling more expensive. It was ~8-10% performance degradation.

Most GraphQL response data is structured data. The router understands it and works with it. For that common case, this representation is intentionally cheap:

```rust
Value::Object(Vec<(&str, Value)>)
```

I did not want to make all normal objects slower just to fix a smaller custom scalar case. So I added a narrow escape hatch `Value::RawJson`.

Every custom scalar in in the schema is treated as raw JSON.

---

**How it all wokrs?**

First, the planner records custom scalar paths. When the query plan is built, the planner walks the fetch output selections. It records response paths for custom scalar fields that should be treated as raw JSON.

The planner does this once when it builds the query plan. This means the executor does not need to find these paths again on every request.

Next, the executor uses these paths while parsing subgraph responses. Normal fields use the normal structured `Value` path, custom scalar fields use `Value::RawJson`.

Finally, serialization phase writes raw JSON back. When the executor serializes `Value::RawJson`, it writes the raw JSON bytes directly. It is not written as a quoted JSON string.

`Value::RawJson` is preserved in two places in the final response projection and request projection.

### DIFF
--- a/.changeset/raw_json_for_custom_scalars.md
+++ b/.changeset/raw_json_for_custom_scalars.md
@@ -1,0 +1,10 @@
+---
+hive-router-plan-executor: patch
+hive-router-query-planner: patch
+hive-router: patch
+node-addon: patch
+---
+
+# Preserve custom scalars as raw JSON
+
+Custom scalar fields marked by the query planner are now preserved as raw JSON instead of being parsed and rebuilt as structured response values. This improves correctness for JSON passthrough custom scalars while avoiding performance regressions for normal response handling.
--- a/e2e/src/issues/mod.rs
+++ b/e2e/src/issues/mod.rs
@@ -1,6 +1,6 @@
 #[cfg(test)]
 mod issues_e2e_tests {
-    use crate::testkit::{ClientResponseExt, TestRouter};
+    use crate::testkit::{ClientResponseExt, Started, TestRouter};
 
     #[ntex::test]
     /// https://github.com/graphql-hive/router/issues/880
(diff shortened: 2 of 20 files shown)


ardatan (approved):
~~I think we need a changeset~~

kamilkisiela:
> ~I think we need a changeset~

I added cuz we touch QP, your point is valid even if crossed out :))

PR #971, opened 2026-05-11
feat(router): allow overriding number of HTTP server workers

Meemaw:
Add an `http.workers` config option (and `WORKERS` env var) so the number of HTTP server worker threads can be set explicitly. By default ntex spawns one worker per physical CPU core, which oversubscribes workers in containerized environments (e.g. Kubernetes) where the container's CPU limit is lower than the host's physical core count.

### DIFF
--- a/lib/router-config/src/env_overrides.rs
+++ b/lib/router-config/src/env_overrides.rs
@@ -32,7 +32,7 @@ pub struct EnvVarOverrides {
     #[envconfig(from = "HOST")]
     pub http_host: Option<String>,
     #[envconfig(from = "WORKERS")]
-    pub http_workers: Option<u64>,
+    pub http_workers: Option<usize>,
 
     // Supergraph overrides
     #[envconfig(from = "SUPERGRAPH_FILE_PATH")]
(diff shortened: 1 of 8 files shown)


lib/router-config/src/env_overrides.rs
```diff
@@ -31,6 +31,8 @@ pub struct EnvVarOverrides {
     pub http_port: Option<u64>,
     #[envconfig(from = "HOST")]
     pub http_host: Option<String>,
+    #[envconfig(from = "WORKERS")]
```
ardatan: I am just not sure about `WORKERS` name as the env var's name. @dotansimha @kamilkisiela what do you think? Other than that, the PR looks good to me.

dotansimha: Yeah, I feel like `WORKERS` could be named differently. What about `ROUTER_HTTP_WORKERS`?

dotansimha (approved):
Thanks! 
I just pushed a tiny change to fix the versioning, as we are still in v0 stage so we stick to `patch` at the moment.

Meemaw:
> Thanks! 
> I just pushed a tiny change to fix the versioning, as we are still in v0 stage so we stick to `patch` at the moment. 
> 

Any blockers to get this merged?

dotansimha:
> > Thanks!
> > I just pushed a tiny change to fix the versioning, as we are still in v0 stage so we stick to `patch` at the moment.
> 
> Any blockers to get this merged?

Nothing really. Just rebased and trying to figure out why CI fails

dotansimha (approved):

dotansimha:
Thanks! this is now available in `hive-router 0.0.57`

PR #996, opened 2026-05-18
fix(router): warn about debug binary in production use

dotansimha:
Added a warning message to the router binary to warn users about using a debug binary in production use.

This is intended to prevent cases where users accidentally use a debug binary in production, which can lead to performance issues.

### DIFF
--- a/.changeset/warn_about_debug_binary.md
+++ b/.changeset/warn_about_debug_binary.md
@@ -1,0 +1,9 @@
+---
+hive-router: patch
+---
+
+# Warn about debug binary in production use 
+
+Added a warning message to the router binary to warn users about using a debug binary in production use.
+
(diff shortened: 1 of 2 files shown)


bin/router/src/main.rs
```diff
@@ -7,6 +9,14 @@ configure_global_allocator!();
 
 #[hive_router::main]
 async fn main() -> Result<(), RouterInitError> {
+    if cfg!(debug_assertions) && env::var("CARGO").is_err() {
```
kamilkisiela: Why not make it part of `router_entrypoint`? If people customize Router, they probably write their own `main.rs` files.

dotansimha: Yeah 100%

kamilkisiela (approved):

PR #1049, opened 2026-05-21
Forward operation names to subgraphs

Meemaw:
## Summary

Adds opt-in forwarding of planner-assigned `operationName` values to subgraph request bodies.

When `traffic_shaping.all.forward_operation_name` or `traffic_shaping.subgraphs.<name>.forward_operation_name` is enabled, the planner generates subgraph operation names as `<ClientOperationName>_<FetchStepData.id>` and the executor forwards that value in the GraphQL-over-HTTP JSON `operationName` field.

Forwarding remains disabled by default, preserving the existing behavior of omitting `operationName` from subgraph calls. No forwarded name is generated when the client operation name is empty or absent.

The forwarding decision is applied before concrete subgraph executors receive the `SubgraphExecutionRequest`, so it works consistently for fetches and subscriptions, including HTTP callback subscription requests.

## Configuration

Global opt-in:

```yaml
traffic_shaping:
  all:
    forward_operation_name: true
```

Per-subgraph opt-in:

```yaml
traffic_shaping:
  subgraphs:
    products:
      forward_operation_name: true
```

## Validation

- `cargo fmt --check`
- `cargo check --locked -p hive-router`
- `cargo check --locked -p hive-router-query-planner -p hive-router-config -p hive-router-plan-executor`
- `cargo check --locked -p hive-router-query-planner`
- `cargo test --locked -p hive-router-config forward_operation_name`
- `cargo test --locked -p hive-router-plan-executor operation_name`
- `cargo test --locked -p hive-router-plan-executor query_document_is_json_escaped`
- `cargo test --locked -p hive-router-query-planner root_types`
- `cargo test --locked -p hive-router-query-planner operation_name_is_not_generated_for_empty_client_operation_name`
- `cargo test --locked -p hive-router-query-planner forwarded_operation_names_include_fetch_step_id_for_subscriptions`
- `git diff --check`

### DIFF
--- a/.changeset/forward-subgraph-operation-name.md
+++ b/.changeset/forward-subgraph-operation-name.md
@@ -6,21 +6,21 @@ hive-router-plan-executor: patch
 
 # Forward operationName to subgraphs
 
-Subgraph HTTP and HTTP callback requests now include the planner-assigned `operationName` in the JSON request body by default. Added `traffic_shaping.all.strip_operation_name` and per-subgraph `traffic_shaping.subgraphs.<name>.strip_operation_name` options for deployments that need to preserve the previous omission.
+Subgraph HTTP and HTTP callback requests can now include a planner-assigned `operationName` in the JSON request body. Added the `traffic_shaping.all.forward_operation_name` and per-subgraph `traffic_shaping.subgraphs.<name>.forward_operation_name` options. The option defaults to `false`, so the previous omission remains the default behavior.
 
-Global opt-out:
+Global opt-in:
 
 ```yaml
 traffic_shaping:
   all:
-    strip_operation_name: true
+    forward_operation_name: true
 ```
 
-Per-subgraph opt-out:
+Per-subgraph opt-in:
 
 ```yaml
 traffic_shaping:
   subgraphs:
     products:
-      strip_operation_name: true
+      forward_operation_name: true
 ```
--- a/.changeset/forward-subgraph-operation-name.md
+++ b/.changeset/forward-subgraph-operation-name.md
@@ -1,0 +1,9 @@
+---
+hive-router: patch
+hive-router-config: patch
+hive-router-plan-executor: patch
+---
+
+# Forward operationName to subgraphs
+
+Subgraph HTTP and HTTP callback requests now include the planner-assigned `operationName` in the JSON request body by default. Added `traffic_shaping.all.strip_operation_name` and per-subgraph `traffic_shaping.subgraphs.<name>.strip_operation_name` options for deployments that need to preserve the previous omission.
--- a/.changeset/forward-subgraph-operation-name.md
+++ b/.changeset/forward-subgraph-operation-name.md
@@ -7,3 +7,20 @@ hive-router-plan-executor: patch
 # Forward operationName to subgraphs
 
 Subgraph HTTP and HTTP callback requests now include the planner-assigned `operationName` in the JSON request body by default. Added `traffic_shaping.all.strip_operation_name` and per-subgraph `traffic_shaping.subgraphs.<name>.strip_operation_name` options for deployments that need to preserve the previous omission.
+
+Global opt-out:
+
+```yaml
+traffic_shaping:
+  all:
+    strip_operation_name: true
+```
+
+Per-subgraph opt-out:
+
+```yaml
+traffic_shaping:
+  subgraphs:
+    products:
+      strip_operation_name: true
+```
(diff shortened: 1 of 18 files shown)


lib/executor/src/executors/http.rs
```diff
@@ -52,13 +52,15 @@ pub struct HTTPSubgraphExecutor {
     pub header_map: HeaderMap,
     pub semaphore: Arc<Semaphore>,
     pub dedupe_enabled: bool,
+    pub strip_operation_name: bool,
```
dotansimha: I find it a bit odd to have that at the specific executor level (`http`). 

In addition to that, having it here means that you get `SubgraphExecutionRequest` in `execute` function and you must declare is as `mut` in order to modify the value. 

Any reason not to put this one level up, let's say here: https://github.com/graphql-hive/router/blob/336101c76b14853969e960430cea03ffbb914659/lib/executor/src/execution/plan.rs#L1342 ? 
Or even higher, before calling `prepare_execution_job`. 

My point is that if we put this flag higher, it will propagate to both execute/subsribe, eliminate the need for `mut`, and will have that option for all executors; not just http.

.changeset/forward-subgraph-operation-name.md
```diff
@@ -0,0 +1,26 @@
+---
+hive-router: patch
+hive-router-config: patch
+hive-router-plan-executor: patch
+---
+
+# Forward operationName to subgraphs
+
+Subgraph HTTP and HTTP callback requests now include the planner-assigned `operationName` in the JSON request body by default. Added `traffic_shaping.all.strip_operation_name` and per-subgraph `traffic_shaping.subgraphs.<name>.strip_operation_name` options for deployments that need to preserve the previous omission.
```
dotansimha: We are discussion of whether this configuration should be the default or not. Any reason on why this new flag should change the default behavior we have now? 
The main reasoning we had for striping it was to save the extra bytes over the network, as we didn't see any value of passing it with every subgraph call.

lib/executor/src/executors/http.rs
```diff
@@ -663,3 +677,69 @@ impl SubgraphHttpResponse {
...
+    fn body_omits_operation_name_when_absent() {
+        let request = make_request("query { __typename }", None);
+        let body = build_request_body(&request).unwrap();
+        let body_str = std::str::from_utf8(&body).unwrap();
+        assert_eq!(body_str, r#"{"query":"query { __typename }"}"#);
+    }
+
+    #[test]
+    fn operation_name_is_json_escaped() {
+        let request = make_request("query { __typename }", Some("With\"Quote"));
```
dotansimha: I think adding `"` is not a legal value in GraphQL, that would fail parsing earlier.

dotansimha (changes requested):

.changeset/forward-subgraph-operation-name.md
dotansimha: Ok that's ok to add the name by default ✅

lib/executor/src/executors/http.rs
Meemaw: Addressed in `f56d0218` and follow-up `d66abcda`.

I moved the operation-name decision out of the concrete HTTP executors and into the shared request preparation path, so `execute`/`subscribe` both receive an already-prepared `SubgraphExecutionRequest` and the HTTP/callback executors no longer mutate it.

I also changed the config to opt-in (`traffic_shaping.*.forward_operation_name`, default `false`) and generate forwarded names as `<ClientOperationName>_<FetchStepData.id>` only when the client operation name is non-empty. Added coverage for regular fetches, empty client operation names, and subscriptions.

kamilkisiela:
Hey! I looked into the approach here and I think there is one important architectural concern around query plan caching.

The forwarded name, for example `"<ClientOp>_<FetchStepId>"`, is generated during query planning inside `build_query_plan_from_fetch_graph` and stored directly on `FetchNode.operation_name`.

That means the cached query plan contains a client-operation-specific value. This can become problematic if the query plan cache is ever shared across instances (Redis). One instance could cache a plan containing `GetProduct_0`, and another request with the same plan shape but a different client operation name could reuse that cached plan and forward the wrong operation name.

The approach I took in #1075 , avoids this by keeping the query plan independent of the client operation name. The plan remains cacheable based on its actual shape. The forwarded operation name is generated later, at execution time, inside `OperationNameFactory::generate()`, using the `name_write_position` metadata already stored in `SubgraphFetchOperation`.

I think this separation is important:

- the query plan stays reusable and anonymous
- the client operation name does not become part of the cached plan
- we avoid leaking one request’s operation name into another request
- we keep the option to share the plan cache across instances later
- we can evolve the generated format without invalidating the plan cache

I also added the subgraph name to the operation name.

dotansimha:
Hi @Meemaw , thanks for the effort. We've decided to take another path in order to implement this feature, please see https://github.com/graphql-hive/router/pull/1075 
This should cover your needs from this PR, it was used as inspiration. 

I'm going to cut a release soon :)

PR #1051, opened 2026-05-22
Add support for end_with_graphql_errors

Meemaw:
## What changed

- Added `end_with_graphql_errors(...)` to start and end hook payloads for response types that can represent GraphQL error lists.
- Added `FromGraphQLErrorsToResponse` plus shared JSON serialization for `{ "errors": [...] }`.
- Wired multi-error conversion through HTTP, subgraph HTTP, subgraph execution, and plan execution response types.
- Added unit coverage and a plugin example test for returning two GraphQL errors in one response.

## Why

`end_with_graphql_error(...)` only accepted one `GraphQLError`, forcing plugins to drop additional violations or manually build raw responses. GraphQL responses support an `errors` array, so the plugin API should expose that directly while preserving the existing single-error helper.

## Validation

- `cargo fmt --check`
- `cargo check -p hive-router-plan-executor --offline`
- `cargo check -p hive-router --offline`
- `cargo test -p hive-router-plan-executor end_with_graphql_errors_returns_all_errors --offline -- --nocapture`
- `cargo test -p hive-router-plan-executor from_graphql_errors_to_bytes_serializes_all_errors --offline -- --nocapture`

Note: attempted `cargo test -p forbid-anonymous-operations-plugin-example should_return_multiple_graphql_errors -- --nocapture`, but Cargo could not fetch uncached `ascii v1.1.0` because the local socket firewall rejected the TLS certificate for `static.crates.io`.

### DIFF
--- a/.changeset/end_with_graphql_errors
+++ b/.changeset/end_with_graphql_errors
@@ -1,0 +1,8 @@
+---
+hive-router-plan-executor: minor
+hive-router: patch
+---
+
+## Add `end_with_graphql_errors(...)` to the plugin hook API.
+
+Plugin authors can now terminate execution with multiple GraphQL errors in a single response. This avoids forcing plugins to either return only one error or manually build a raw GraphQL error response when several errors should be reported together.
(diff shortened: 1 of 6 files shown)


dotansimha:
Thanks @Meemaw , i will be available in a minute in `hive-router 0.0.64`

Issue #922, opened 2026-04-15
Subscriptions plugin system/hooks


enisdenjo:
Improvement to https://github.com/graphql-hive/router/pull/620.

PR #1072, opened 2026-05-29
feat(executor): invoke on_subgraph_execute on subscribe path

AtIasShrugged:
## Summary

Closes part of #922.

`on_subgraph_execute` was only invoked on the execute path; subscription registration in `executors::map::SubgraphExecutorMap::subscribe` bypassed all plugin hooks. Plugins that mutate `execution_request.headers` (for example to inject auth or tenancy context) had no way of affecting the subscription registration request, so subgraphs received an unmodified payload.

This PR invokes the hook on the subscribe path using the same loop semantics as the execute path — same payload, same control-flow types. It plugs the gap for the most common plugin use case (request mutation) while leaving the harder lifetime work (`'exec`→`'static` materialization for `EndWithResponse`, end-of-stream semantics for `OnEnd`) for a follow-up.

## Control-flow handling on the subscribe path

| Variant | Behavior |
|---|---|
| `Proceed` | Fully supported — identical to the execute path. |
| `EndWithResponse` | Returns a dedicated `SubgraphExecutorError::SubscribePluginEndWithResponseUnsupported` (code `SUBGRAPH_SUBSCRIBE_PLUGIN_END_WITH_RESPONSE_UNSUPPORTED`). Materializing a `SubgraphResponse<'exec>` into the `'static` `BoxStream` returned by `subscribe` needs a deliberate solution (clone-into-owned for `Value<'a>`) and is intentionally deferred. |
| `OnEnd` | Logs a `tracing::warn!` and drops the callback. There is no symmetric end-of-stream invocation point yet; the rest of the plugin's request mutations still take effect. |

Both deferred variants surface explicitly to the plugin author rather than being silently dropped, so the limitation is discoverable.

## Changes

- `lib/executor/src/executors/map.rs` — `subscribe()` accepts `plugin_req_state: Option<&'exec PluginRequestState<'exec>>` and runs the hook loop before delegating to the underlying subgraph executor. Mirrors `execute()` line-for-line where the semantics line up.
- `lib/executor/src/execution/plan.rs` — single subscribe call site now passes `opts.plugin_req_state.as_ref()`.
- `lib/executor/src/executors/error.rs` — new `SubscribePluginEndWithResponseUnsupported` variant.
- `e2e/src/subscription_plugin_hooks.rs` — two e2e tests:
  - `on_subgraph_execute_invoked_on_subscribe_sse` — registers a plugin that injects a marker header, runs an SSE subscription, asserts the subgraph received the header on the registration request.
  - `on_subgraph_execute_end_with_response_unsupported_on_subscribe` — confirms the dedicated error code reaches the client when a plugin returns `EndWithResponse`.

## Out of scope

The per-event re-execution `// TODO: plugins for subscriptions are not yet supported` in `execution/plan.rs:305` is unchanged — it requires `PluginRequestState<'exec>` to be reshaped (or `Arc`-ified) so it can live across `'static` stream boundaries. Splitting that into a separate PR keeps this one reviewable.

## Test plan

- [x] `cargo check --workspace`
- [x] `cargo clippy --workspace --no-deps -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test -p hive-router-plan-executor --lib` — 92 passed
- [x] `cargo test -p e2e subscription_plugin_hooks` — 2 new tests pass
- [x] `cargo test -p e2e subscriptions` — all 24 existing subscription tests pass (no regression)
- [x] `cargo test -p e2e http_callback` — 5/5 pass
- [x] `cargo test -p e2e websocket_` — 10/10 pass
- [x] `cargo test -p e2e coprocessor` — 31/31 pass

### DIFF
--- a/.changeset/subscription_plugin_hooks
+++ b/.changeset/subscription_plugin_hooks
@@ -1,0 +1,10 @@
+---
+hive-router-plan-executor: minor
+hive-router: patch
+---
+
+## Invoke `on_subgraph_execute` plugin hook on the subscribe path.
+
+Previously, the `on_subgraph_execute` hook was only invoked for queries and mutations; subscription registration bypassed all plugin hooks. Plugins that mutate `execution_request.headers` (for example to inject auth context) had no way of affecting the subscription registration request, so subgraphs saw an unmodified registration payload.
+
+The hook is now invoked on the subscribe path with the same loop semantics as the execute path. Only `proceed` is fully supported for now; `end_with_response` surfaces a dedicated `SUBGRAPH_SUBSCRIBE_PLUGIN_END_WITH_RESPONSE_UNSUPPORTED` error (mapping a single response into a `'static` stream still needs to be worked out), and `on_end` callbacks are dropped with a warning. Tracked upstream in [#922](https://github.com/graphql-hive/router/issues/922).
--- a/e2e/src/lib.rs
+++ b/e2e/src/lib.rs
@@ -55,6 +55,8 @@ mod router_timeout;
 #[cfg(test)]
 mod storage;
 #[cfg(test)]
+mod subscription_plugin_hooks;
+#[cfg(test)]
 mod subscriptions;
 #[cfg(test)]
 mod supergraph;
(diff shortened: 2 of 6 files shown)


AtIasShrugged:
Closing — opened prematurely without local validation against real subgraphs. Will reopen after verification.

AtIasShrugged:
Reopening after addressing prior auto-review feedback and validating end-to-end against a stage cluster. See the updated description for details.

PR #1075, opened 2026-05-29
Forward client operation name to subgraphs

kamilkisiela:
Adds opt-in forwarding of the client operation name to subgraph requests. The operation name is injected into both the query document and the "operationName" JSON field, formatted as `<client_name>__<fetch_id>`.

Config:
```yaml
traffic_shaping:
  all:
    forward_operation_name: true
  subgraphs:
    products:
      forward_operation_name: true
```

Defaults to `false`.

---

The forwarded **name is generated at execution time rather than during query planning**. This keeps the query plan cache clean, so plans remain anonymous and can be shared across instances (important if the cache moves to Redis at some point). If the name were baked into the plan, the cache key would need to include the client operation name, reducing hit rates for identical-shape operations with different names, and a shared cache would leak names between instances.

The name write position is pre-computed from the anonymous document in `SubgraphFetchOperation::from_anonymous_operation()`, so no coupling between the client operation name and the plan.


lib/executor/src/execution/operation_name.rs
```diff
@@ -0,0 +1,161 @@
...
+        let mut fetch_step_id_buf = itoa::Buffer::new();
+        let fetch_step_id = fetch_step_id_buf.format(fetch_step_id);
+        let mut operation_name = String::with_capacity(
+            /* 2 "_" separators */
+            2 + client_operation_name.len() + subgraph_name.len() + fetch_step_id.len(),
+        );
+
+        // Operation name does not need to be sanitized, so we use it raw
+        operation_name.push_str(client_operation_name);
+        operation_name.push('_');
```
dotansimha: I wonder if this delimiter needs to be `__`, like other routers are doing? I assume it's to allow users to distinguish and split based on something that has less chances to be used by the operation name? (an incoming operation name with `query my_query` will end up as `my_query_{name}_{id}` which makes the result of `split` less consistent.

lib/executor/src/execution/operation_name.rs
```diff
@@ -0,0 +1,161 @@
...
+        for char in subgraph_name.chars() {
+            operation_name.push(if char.is_ascii_alphanumeric() {
+                char
+            } else {
+                '_'
+            });
+        }
+
+        // Finally the fetch step id
+        operation_name.push('_');
```
dotansimha: Same as the one above regarding the delimiter

lib/executor/src/execution/operation_name.rs
```diff
@@ -0,0 +1,161 @@
...
+            2 + client_operation_name.len() + subgraph_name.len() + fetch_step_id.len(),
+        );
+
+        // Operation name does not need to be sanitized, so we use it raw
+        operation_name.push_str(client_operation_name);
+        operation_name.push('_');
+
+        // Sanitize the subgraph name as it may contain characters
+        // that will make the operation name invalid
+        for char in subgraph_name.chars() {
```
dotansimha: what's the value in having the subgraph name in the name? I mean, isn't that obvious where the operation goes to? (in logs, outputs, and well, in the subgraph service itself?)

dotansimha (approved):
Approved, but added 2 questions that I would like to think about before merging?

PR #1085, opened 2026-06-02
fix(router): serve GraphQL on browser GET when Laboratory is disabled

Meemaw:
## Problem

Hitting the router with a `GET` from a browser returns **404** when Laboratory is disabled.

A browser sends an `Accept` header that prefers `text/html`:

```
text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7
```

`negotiate()` checks `text/html` first on `GET` and short-circuits to `ResponseMode::Laboratory` before ever considering the `*/*;q=0.8` fallback that `application/json` would match. When Laboratory is disabled, `pipeline/mod.rs` then returned `HttpResponse::NotFound()` instead of handling the request as normal GraphQL.

## Fix

`negotiate()` now takes a `laboratory_enabled: bool`. When `false`, it skips the `text/html` GET branch entirely and re-negotiates the regular GraphQL content types, so the broad `*/*` browsers send resolves to `Dual(JSON, IncrementalDelivery)` — a valid GraphQL response.

Because `negotiate()` only returns `Laboratory` when it is enabled, the `else { NotFound }` branch in `pipeline/mod.rs` is now dead and has been removed.

This keeps the full `Accept` header in play (preserving streaming support) rather than hard-coding a single JSON response.

## Tests

In `pipeline::header::tests`:
- `negotiate_content_types` gains a `laboratory_enabled` field and new cases, including the real browser header with Laboratory disabled → `Dual(JSON, IncrementalDelivery)`.
- `browser_get_with_only_html_and_laboratory_disabled_is_unsupported` — a pathological `Accept: text/html`-only GET (no `*/*`) returns an explicit `UnsupportedContentType` error rather than a misleading 404.
- `html_only_get_renders_laboratory_when_enabled` — the same request still serves Laboratory when enabled.

### DIFF
--- a/.changeset/fix_get_request_detection_when_laboratory_disabled.md
+++ b/.changeset/fix_get_request_detection_when_laboratory_disabled.md
@@ -1,0 +1,9 @@
+---
+hive-router: patch
+---
+
+# Fix GET request detection when Laboratory disabled
+
+Hitting the router with a `GET` request from a browser was returning 404 when Laboratory is disabled. 
+
+This change adds a check to only negotiate for Laboratory when Laboratory is enabled, and fallbacks to GET request handling when Laboratory is disabled.
--- a/bin/router/src/pipeline/header.rs
+++ b/bin/router/src/pipeline/header.rs
@@ -223,15 +223,20 @@ impl ResponseMode {
 pub trait RequestAccepts {
     /// Reads the request's `Accept` header and returns the agreed response mode.
     ///
+    /// When `laboratory_enabled` is `true`, a GET request that prefers `text/html`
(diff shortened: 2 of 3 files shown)


enisdenjo (approved):
this makes sense to me, thanks for the contribution. can you also make a patch changeset so that we can queue this pr for release? and of course run cargo fmt

dotansimha (approved):

dotansimha:
> Hitting the router with a `GET` from a browser returns **404** when Laboratory is disabled.

Did both and rebased, will merge when CI clears. @Meemaw Thanks!

dotansimha:
Fix is available in `hive-router 0.0.65`

Issue #1076, opened 2026-05-31
query-planner: avoid `panic!` when parsing Federation directives


dotansimha:
Right now, we `panic!` in multiple places. This should be avoided and we should print a better error that explains the rejection of the supergraph due to invalid values.

PR #1089, opened 2026-06-03
fix: do not panic on bad schema unwrap

Ty3uK:
In my company we have custom k8s operator that joins every deployed subgraph to supergraph. In some cases it may generate corrupted schema.

Right now `parse_schema` panics on corrupted schema which leads to destroying `mpsc::rx` so new poll of proper schema will raise an error `Failed to send new supergraph SDL: receiver dropped.`

First approach to handle this was like that:
```rust
#[inline]
pub fn parse_schema(
    sdl: &str,
) -> Result<
    graphql_tools::parser::schema::Document<'static, String>,
    graphql_tools::parser::schema::ParseError,
> {
    graphql_tools::parser::parse_schema(sdl).map(|document| document.into_static())
}
```
With that we can propagate parsing error to `SchemaState` and log it properly. But there are to many tests that relies on `parse_schema` returning `Document` and not `Result` so I decided to show both versions for further discussion.

UPD: never mind, I didn't notice `safe_parse_schema`, thanks gemini.

### DIFF
--- a/.changeset/graceful_handling_of_invalid_supergraph.md
+++ b/.changeset/graceful_handling_of_invalid_supergraph.md
@@ -1,0 +1,11 @@
+---
+hive-router: patch
+---
+
+# Graceful handling of invalid Supergraph while polling
+
+As described in [issue #1089](https://github.com/graphql-hive/router/issues/1089), when the Supergraph fails to parse, the internal mpsc channel should not panic or collapse. 
+
+This fix prevents the Router from crashing when the Supergraph fails to parse, and keeps the channel alive for future updates. 
+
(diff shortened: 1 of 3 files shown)


dotansimha:
Thanks @Ty3uK ! This is super helpful. 

I've added a failing test (that fails on `main` and passes on your PR), so the fix is totally valid.
Also, added a changeset for this, and improved how the error is being logged. 

Worth mentioning that we also have an open issue to get rid of other panic points in the router (https://github.com/graphql-hive/router/issues/1076) 

Waiting for CI to clear and then will merge and cut a release with another batch of improvements and bug fixes :)

Ty3uK:
> Worth mentioning that we also have an open issue to get rid of other panic points in the router (https://github.com/graphql-hive/router/issues/1076) 

Oh, I've missed it, thanks 🙏

dotansimha:
Fix is available in `hive-router 0.0.65`

PR #1100, opened 2026-06-06
fix(query-planner): fix union list FieldMove creation

Ty3uK:
My debug journey started from the end of the chain: I noticed that for my query in which I have two same shaped fragments hive-router resolves data only for one of them.

For query like that:
```graphql
query {
  orders {
    ... on OrdersPageConnections {
      items {
        id
        items {
          ... on PaperBook {
            product {
              id
              name
              slug
            }
          }
          ... on DigitalBook {
            product {
              id
              name
              slug
            }
          }
        }
      }
    }
  }
}
```
my "orders" subgraph resolved IDs for both kinds of items, like this:
```json
[
  {
    "__typename":"PaperBook",
    "product": {
      "__typename": "Product",
      "id": "123"
    }
  },
  {
    "__typename":"DigitalBook",
    "product": {
      "__typename": "Product",
      "id": "456"
    }
  }
]
```
But on the subgraph side I was able to see incoming query only for ID "456".

I quickly found `batch_multi_type` optimization and started thinking that the error is somewhere there because I thought that optimizer cut `@` list definition from `response_path`. But after some time I realized that the issue is somewhere earlier in the planning chain and I found that only in one case we didn't pass `field_definition.field_type.is_list()` to `is_list` parameter of `create_field_move`.

### DIFF
--- a/.changeset/fix_union_list_fieldmove_creation.md
+++ b/.changeset/fix_union_list_fieldmove_creation.md
@@ -1,0 +1,7 @@
+---
+hive-router-query-planner: patch
+---
+
+# Fix union list FieldMove creation
+
+In some cases union list was treated as single union field in graph. 
--- a/.changeset/fix_union_list_fieldmove_creation.md
+++ b/.changeset/fix_union_list_fieldmove_creation.md
@@ -1,5 +1,8 @@
 ---
 hive-router-query-planner: patch
+hive-router-plan-executor: patch
+hive-router: patch
+node-addon: patch
 ---
 
 # Fix union list FieldMove creation
--- a/lib/query-planner/src/tests/provides.rs
+++ b/lib/query-planner/src/tests/provides.rs
@@ -245,7 +245,7 @@ fn provides_on_union() -> Result<(), Box<dyn Error>> {
             }
           }
         },
-        Flatten(path: "media|[Movie]") {
+        Flatten(path: "media.@|[Movie]") {
           Fetch(service: "c") {
             {
               ... on Movie {
@@ -281,6 +281,7 @@ fn provides_on_union() -> Result<(), Box<dyn Error>> {
               {
                 "Field": "media"
               },
+              "@",
               {
                 "TypeCondition": [
                   "Movie"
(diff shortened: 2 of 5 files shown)


lib/query-planner/src/tests/union.rs
```diff
@@ -503,3 +503,93 @@ fn union_overfetching_test() -> Result<(), Box<dyn Error>> {
...
+                }
+              }
+            }
+          }
+          fragment a on Product {
+            __typename
+            id
+          }
+        },
+        Flatten(path: "orders|[OrdersPageConnections].items.@.items.@.product") {
```
Ty3uK, line 574: Without fix it was `orders|[OrdersPageConnections].items.@.items.product` (last `@` missing) which was leading to fetching data only for `DigitalBook`.

lib/query-planner/src/tests/provides.rs
```diff
@@ -245,7 +245,7 @@ fn provides_on_union() -> Result<(), Box<dyn Error>> {
             }
           }
         },
-        Flatten(path: "media|[Movie]") {
+        Flatten(path: "media.@|[Movie]") {
```
Ty3uK, line 248: @dotansimha, I have missed the fact that there is already a failing test for this issue. Should I keep my freshly written test or the old one is good to go?

dotansimha:
Hi @Ty3uK , thank you for opening this PR. Is there an issue associated with it? It lacks an issue an a description, so moving it to draft while we fill the gaps. 

Thanks

Ty3uK:
@dotansimha, oh, sorry, I didn't think that I need to fill issue :slightly_smiling_face: 

Should I write a description to a PR or fill the issue?

Ty3uK:
@dotansimha, I've added a description to this PR. :slightly_smiling_face:

Ty3uK:
@dotansimha, hey, can you please check out the PR? :slightly_smiling_face:

.changeset/fix_union_list_fieldmove_creation.md
```diff
@@ -0,0 +1,7 @@
+---
+hive-router-query-planner: patch
```
kamilkisiela, line 2:
```suggestion
hive-router-query-planner: patch
hive-router-plan-executor: patch
hive-router: patch
node-addon: patch
```

lib/query-planner/src/tests/provides.rs
dotansimha: I think if the test has a different scope, it's to leave it - there is no such thing "too many tests" 😆 

Thanks

PR #1107, opened 2026-06-08
fix(executor): handle case where only `__typename` is used for entity key

### LINKED ISSUE #1099: Hive Router skips `_entities` fetch when `requires` is only `__typename`
## Summary

Hive Router builds a correct query plan with a `Flatten` → `_entities` hop, but the executor never issues the subgraph HTTP request when the representation `requires` selection is only `__typename`.

Apollo Router with the same supergraph and subgraphs executes the plan and resolves the field.

## Environment

- **Hive Router:** `ghcr.io/graphql-hive/router:latest`
- **Apollo Router (comparison):** `ghcr.io/graphql-hive/apollo-router:router1.60.1-plugin1.1.1`
- **Federation:** 2.8 (`federation_version: =2.8.0`)
- **Config:** `query_planner.allow_expose: true`, request header `hive-expose-query-plan: true`

## Schema

Two subgraphs. `CatalogEntry` is keyed only by `__typename` on both.

**warehouse** (`http://warehouse:9501/graphql`):

```graphql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.8", import: ["@key"])

type Query {
  catalogEntry: CatalogEntry
}

type CatalogEntry @key(fields: "__typename") {
  sku: String!
}
```

**reviews** (`http://reviews:9502/graphql`):

```graphql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.8", import: ["@key"])

type Rating {
  score: Int!
}

extend type CatalogEntry @key(fields: "__typename") {
  rating: Rating!
}
```

### Query

```graphql
query {
  catalogEntry {
    sku
    rating {
      score
    }
  }
}
```

### Hive Router response

```json
{
  "data": {
    "catalogEntry": {
      "sku": "SKU-REPRO-001",
      "rating": null
    }
  }
}
```

### Observed behavior (Hive)

1. **warehouse** is called once with `{catalogEntry{__typename sku}}`.
2. **reviews** is never called.
3. `rating` is `null`.

With `hive-expose-query-plan: true`, the plan includes the expected second step:

```json
{
  "kind": "Flatten",
  "path": [{ "Field": "catalogEntry" }],
  "node": {
    "serviceName": "reviews",
    "kind": "Fetch",
    "requires": [{
      "kind": "InlineFragment",
      "typeCondition": "CatalogEntry",
      "selections": [{ "kind": "Field", "name": "__typename" }]
    }],
    "operation": "query($representations:[_Any!]!){_entities(representations: $representations){...on CatalogEntry{rating{score}}}}"
  }
}
```

The planner looks correct; the executor skips the fetch silently (no HTTP to reviews).

### Expected behavior

1. **warehouse** returns `catalogEntry { __typename, sku }`.
2. **reviews** receives `_entities(representations: [{"__typename":"CatalogEntry"}])`.
3. Response includes `rating { score: 42 }`.

### Apollo Router (same supergraph + subgraphs)

```json
{
  "data": {
    "catalogEntry": {
      "sku": "SKU-REPRO-001",
      "rating": { "score": 42 }
    }
  }
}
```

Apollo issues the reviews `_entities` request as planned.

## Minimal variant

The same failure occurs with only the cross-subgraph field:

```graphql
query {
  catalogEntry {
    rating {
      __typename
    }
  }
}
```

- warehouse: `{catalogEntry{__typename}}` only.
- reviews: still not called.
- `rating: null`.

## Suspected cause

Representation projection appears to skip `__typename` in the requires loop ("handled separately") but only emits it when other fields are projected. When `requires` lists **only** `__typename`, no representation is built and `Flatten` fetch is skipped.

Relevant logic in `lib/executor/src/projection/request.rs`:

```rust
if response_key == TYPENAME_FIELD_NAME {
    // Skip __typename field, it is handled separately
    continue;
}
// ...
if first {
    // If no fields were projected, "first" is still true,
    // so we skip writing the closing brace
    return false;
}
```

## Impact

Any federated entity whose cross-subgraph `@key` is only `__typename` may fail to resolve fields from other subgraphs.

## Standalone repro

https://github.com/mabysh/hive-router-example-2

dotansimha:
Fixes https://github.com/graphql-hive/router/issues/1099 

## TL;DR:

* `@key(fields: "__typename")` -> should create fetch call (currently, it's skipped)  
* `@key(fields: "__typename id")` + `id = null` ->  should not create fetch call (tested in `null_keys` in fed audit) 
* `@key(fields: "__typename id")` + `id = something` -> should create fetch call ("normal" flow)  

## More stuff 

As described in [issue #1099](https://github.com/graphql-hive/router/issues/1099), when an entity's `@key` is only `__typename` (e.g. `@key(fields: "__typename")`), the executor built a correct query plan but never issued the `_entities` request to the other subgraph, leaving the cross-subgraph field resolved as `null`.

The representation projection skipped the `__typename` field and only emitted it alongside other fields, so a key using only `__typename` field produced an empty representation and the entity fetch was silently dropped.

The projection now emits a `{ "__typename": ... }` representation in this case, so the entity fetch runs and the field resolves as expected.

I've added a few tests to ensure projection is handling different combos of `__typename` (only `__typename`, `__typename`+field, field+`__typename`). Also added a full e2e tests based on the original issue. 
Also, copied the `null_keys` from the Fed audit because I needed to run it here as regression test, because the test there ends up with `{__typename: "..."}` because of a null key for one of the values, and in that edge-case, we do want to drop the fetch step.

### DIFF
--- a/lib/executor/src/projection/request.rs
+++ b/lib/executor/src/projection/request.rs
@@ -24,6 +24,15 @@ fn write_response_key(first: bool, response_key: Option<&str>, buffer: &mut Vec<
     }
 }
 
+#[inline]
+fn write_typename_field(buffer: &mut Vec<u8>, type_name: &str) {
+    buffer.put(QUOTE);
+    buffer.put(TYPENAME);
+    buffer.put(QUOTE);
+    buffer.put(COLON);
+    write_and_escape_string(buffer, type_name);
+}
+
 pub fn project_requires(
     possible_types: &PossibleTypes,
     requires_selections: &Vec<SelectionItem>,
@@ -124,30 +133,47 @@ fn project_requires_map_mut(
     parent_response_key: Option<&str>,
     parent_first: bool,
 ) {
+    // First, check if __typename is present and if it's needs to be added
+    let type_name = entity_obj
+        .binary_search_by_key(&TYPENAME_FIELD_NAME, |(k, _)| k)
+        .ok()
+        .and_then(|idx| entity_obj[idx].1.as_str());
+
     for requires_selection in requires_selections {
         match &requires_selection {
             SelectionItem::Field(requires_selection) => {
                 let field_name = &requires_selection.name;
                 let response_key = requires_selection.selection_identifier();
                 if response_key == TYPENAME_FIELD_NAME {
-                    // Skip __typename field, it is handled separately
+                    // __typename is normally written when the object is opened for
+                    // its first non-__typename field (see below).
+                    //
+                    // But, for the very rate case of using only __typename in key (`@key(fields: "__typename")`),
+                    // we need to write it here instead, because the code below never runs (no other fields).
+                    //
+                    // So: if this is the first field, write __typename immediately, and we need to write __typename anyway,
+                    // we write it, and then continue to the next field (marking first=false to avoid re-write)
+                    if *first {
+                        if let Some(type_name) = type_name {
+                            write_response_key(parent_first, parent_response_key, buffer);
+                            buffer.put(OPEN_BRACE);
+                            write_typename_field(buffer, type_name);
+                            *first = false;
+                        }
+                    }
+
                     continue;
                 }
 
                 let original = entity_obj
                     .binary_search_by_key(&field_name.as_str(), |(k, _)| k)
                     .ok()
-                    .map(|idx| &entity_obj[idx].1)
                     .or_else(|| {
-                        if response_key == TYPENAME_FIELD_NAME {
-                            None
-                        } else {
-                            entity_obj
-                                .binary_search_by_key(&response_key, |(k, _)| k)
-                                .ok()
-                                .map(|idx| &entity_obj[idx].1)
-                        }
-                    });
+                        entity_obj
+                            .binary_search_by_key(&response_key, |(k, _)| k)
+                            .ok()
+                    })
+                    .map(|idx| &entity_obj[idx].1);
 
                 let Some(original) = original else {
                     continue;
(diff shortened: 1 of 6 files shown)


lib/executor/src/projection/request.rs
```diff
@@ -124,30 +133,47 @@ fn project_requires_map_mut(
...
+                    // So: if this is the first field, write __typename immediately, and we need to write __typename anyway,
+                    // we write it, and then continue to the next field (marking first=false to avoid re-write)
+                    if *first {
+                        if let Some(type_name) = type_name {
+                            write_response_key(parent_first, parent_response_key, buffer);
+                            buffer.put(OPEN_BRACE);
+                            write_typename_field(buffer, type_name);
+                            *first = false;
+                        }
+                    }
```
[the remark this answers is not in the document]
dotansimha: the feels weird, i'm skipping

lib/executor/src/projection/request.rs
```diff
@@ -24,6 +24,15 @@ fn write_response_key(first: bool, response_key: Option<&str>, buffer: &mut Vec<
     }
 }
 
+#[inline]
+fn write_typename_field(buffer: &mut Vec<u8>, type_name: &str) {
+    buffer.put(QUOTE);
+    buffer.put(TYPENAME);
+    buffer.put(QUOTE);
+    buffer.put(COLON);
```
kamilkisiela: maybe just `"__typename":`?

dotansimha: moved to a const

kamilkisiela (approved):

PR #1111, opened 2026-06-09
fix(query-planner): `partial-union` test suite 

### LINKED ISSUE #1098: Hive Router Rust strips inline fragments for partial union members
# Bug Report: Hive Router Rust strips inline fragments for partial union members

## Summary

When two federated subgraphs define a `@shareable` type containing a union, and the union has **different members across subgraphs** (a "partial union"), the Hive Router Rust query planner strips inline fragments for union members that only exist in one subgraph. This causes all fields below those union members to return `null`.

**Apollo Gateway** and **Hive Gateway JS** handle the same supergraph correctly — this is specific to the **Hive Router Rust** query planner.

## Versions

- **Hive Router Rust**: `hive-router = 0.0.65`, `hive-router-query-planner = 2.8.4`
- **Apollo Gateway**: v2.9.x (not affected)
- **Hive Gateway JS**: v2.8.1 (not affected)
- **GraphQL Spec**: Federation v2.0

### Schema Setup

**Subgraph A ("primary")** — defines full union:

```graphql
extend schema @link(url: "https://specs.apollo.dev/federation/v2.0",
  import: ["@key", "@shareable"])

type Query {
  getResponse: Response
}

type Response @shareable {
  actions: [Action!]!
  message: String
}

union Action = Alpha | Beta | Gamma

type Alpha @shareable {
  id: ID!
  value: String
}

type Beta {
  id: ID!
  name: String
  details: String
}

type Gamma {
  id: ID!
  label: String
}
```

**Subgraph B ("secondary")** — defines partial union (only `Alpha`):

```graphql
extend schema @link(url: "https://specs.apollo.dev/federation/v2.0",
  import: ["@key", "@shareable"])

type Response @shareable {
  actions: [Action!]!
  message: String
}

union Action = Alpha

type Alpha @shareable {
  id: ID!
  value: String
}
```

### Composition

Composition succeeds without errors. The supergraph contains:

```graphql
union Action
  @join__type(graph: PRIMARY)
  @join__type(graph: SECONDARY)
  @join__unionMember(graph: PRIMARY, member: "Alpha")
  @join__unionMember(graph: SECONDARY, member: "Alpha")
  @join__unionMember(graph: PRIMARY, member: "Beta")
  @join__unionMember(graph: PRIMARY, member: "Gamma")
  = Alpha | Beta | Gamma
```

Note: `Beta` and `Gamma` only have `@join__unionMember` for `PRIMARY`.

### Client Query

```graphql
{
  getResponse {
    message
    actions {
      ... on Alpha { id value }
      ... on Beta { id name details }
      ... on Gamma { id label }
    }
  }
}
```

### Expected Behavior

The router should send the full query (with all inline fragments) to the **primary** subgraph, since it is the only subgraph that defines `Beta` and `Gamma`.

### Actual Behavior (Hive Router Rust v0.0.65)

The router sends this query to the primary subgraph:

```
{getResponse{message actions{__typename ...on Alpha{id value}}}}
```

**The `...on Beta` and `...on Gamma` inline fragments are stripped.** Only `...on Alpha` (which exists in both subgraphs) is preserved.

The router does **not** send any follow-up entity resolution query for Beta/Gamma fields, because these types have no `@key` directive — they are not federation entities.

### Result

```json
{
  "data": {
    "getResponse": {
      "message": "ok",
      "actions": [
        { "id": "a1", "value": "alpha-val" },
        { "id": null, "name": null, "details": null },
        { "id": null, "label": null }
      ]
    }
  }
}
```

- `Alpha` fields resolve correctly (exists in both subgraphs)
- `Beta` fields are all `null` (fragment was stripped)
- `Gamma` fields are all `null` (fragment was stripped)

## Comparison Across Routers

| Router | Version | Query sent to primary subgraph | Beta/Gamma fields | Status |
|---|---|---|---|---|
| Apollo Gateway | v2.9 | Full query with all 3 inline fragments | Resolved correctly | **PASS** |
| Hive Gateway JS | v2.8.1 | Full query with all 3 inline fragments | Resolved correctly | **PASS** |
| **Hive Router Rust** | **v0.0.65** | Only `...on Alpha` — Beta/Gamma stripped | **null** | **FAIL** |

## Root Cause Analysis

The Hive Router Rust query planner appears to reason as follows when it encounters a `@shareable` type containing a union:

1. `Response` is `@shareable` across `PRIMARY` and `SECONDARY` → either subgraph can resolve it
2. The `actions` field returns `[Action]` — a union
3. `SECONDARY` only knows about `Alpha`, not `Beta` or `Gamma`
4. The planner attempts to "distribute" resolution by stripping fragments for types the secondary can't resolve, expecting to fetch them via entity resolution later
5. But `Beta` and `Gamma` have no `@key` directive — they are not federation entities
6. Entity resolution cannot happen → fields stay `null`

The planner should recognize that when union members lack `@key`, entity resolution is not possible, and it should include their inline fragments in the query to the subgraph that **does** define them.

## Related Issues

- [graphql-hive/gateway#2381](https://github.com/graphql-hive/gateway/issues/2381) — "Inline Fragment Type Condition Lost During Schema Stitching Delegation" (closed Jun 1, 2026)
- [apollographql/federation#1917](https://github.com/apollographql/federation/issues/1917) — "Apollo Gateway Query Plan incorrectly handles union/fragment of sharable Interfaces" (open)

## Reproducer Script

The self-contained Node.js reproducer script below:
1. Composes the supergraph from the two subgraph schemas above
2. Starts real GraphQL subgraph servers (using `@apollo/subgraph`)
3. Starts logging proxy servers to intercept what the router sends
4. Spawns the Hive Router Rust binary
5. Sends the test query
6. Reports the results

### Prerequisites

```bash
npm install @apollo/subgraph @apollo/server @apollo/composition graphql
```

### Script (`full-test-graphql.mjs`)

```javascript
import http from "node:http";
import fs from "node:fs";
import { parse } from "graphql";
import { composeServices } from "@apollo/composition";
import { buildSubgraphSchema } from "@apollo/subgraph";
import { ApolloServer } from "@apollo/server";
import { startStandaloneServer } from "@apollo/server/standalone";
import { spawn } from "node:child_process";

const PRIMARY_PORT   = 17001;
const SECONDARY_PORT = 17002;
const PROXY_PRIMARY  = 17011;
const PROXY_SECONDARY = 17012;
const ROUTER_PORT    = 17000;

// Point this at the hive-router binary
const ROUTER_BIN = process.env.HIVE_ROUTER_BIN || "./hive-router";

const primarySDL = `
  extend schema @link(url: "https://specs.apollo.dev/federation/v2.0",
    import: ["@key", "@shareable"])
  type Query { getResponse: Response }
  type Response @shareable { actions: [Action!]!; message: String }
  union Action = Alpha | Beta | Gamma
  type Alpha @shareable { id: ID!; value: String }
  type Beta { id: ID!; name: String; details: String }
  type Gamma { id: ID!; label: String }
`;

const secondarySDL = `
  extend schema @link(url: "https://specs.apollo.dev/federation/v2.0",
    import: ["@key", "@shareable"])
  type Response @shareable { actions: [Action!]!; message: String }
  union Action = Alpha
  type Alpha @shareable { id: ID!; value: String }
`;

// Compose supergraph
const result = composeServices([
  { name: "primary",   typeDefs: parse(primarySDL),   url: `http://127.0.0.1:${PROXY_PRIMARY}` },
  { name: "secondary", typeDefs: parse(secondarySDL), url: `http://127.0.0.1:${PROXY_SECONDARY}` },
]);
if (result.errors?.length) { console.error(result.errors); process.exit(1); }

fs.mkdirSync("hive-config", { recursive: true });
fs.writeFileSync("hive-config/supergraph.graphql", result.supergraphSdl);
fs.writeFileSync("hive-config/router.config.yaml",
  `supergraph:\n  source: file\n  path: ./supergraph.graphql\nhttp:\n  host: "0.0.0.0"\n  port: ${ROUTER_PORT}\n  graphql_endpoint: /graphql\nintrospection:\n  expression: 'true'\nlog:\n  level: info\n`);

// Start real GraphQL subgraphs
const primaryServer = new ApolloServer({
  schema: buildSubgraphSchema({
    typeDefs: parse(primarySDL),
    resolvers: {
      Query: {
        getResponse: () => ({
          message: "ok",
          actions: [
            { __typename: "Alpha", id: "a1", value: "alpha-val" },
            { __typename: "Beta",  id: "b1", name: "beta-name", details: "beta-details" },
            { __typename: "Gamma", id: "g1", label: "gamma-label" },
          ],
        }),
      },
      Action: { __resolveType(obj) { return obj.__typename; } },
    },
  }),
});
const secondaryServer = new ApolloServer({
  schema: buildSubgraphSchema({
    typeDefs: parse(secondarySDL),
    resolvers: { Action: { __resolveType(obj) { return obj.__typename; } } },
  }),
});

await startStandaloneServer(primaryServer,   { listen: { port: PRIMARY_PORT } });
await startStandaloneServer(secondaryServer, { listen: { port: SECONDARY_PORT } });

// Proxy servers to log what the router sends to subgraphs
const proxyQueries = { primary: [], secondary: [] };

function createProxy(name, targetPort, proxyPort) {
  return new Promise((resolve) => {
    const server = http.createServer((req, res) => {
      let body = "";
      req.on("data", (c) => { body += c; });
      req.on("end", () => {
        try {
          const p = JSON.parse(body);
          if (!p.query?.includes("IntrospectionQuery")) proxyQueries[name].push(p.query || "");
        } catch {}
        const fwd = http.request({
          hostname: "127.0.0.1", port: targetPort,
          path: req.url, method: req.method,
          headers: { ...req.headers, host: `127.0.0.1:${targetPort}` },
        }, (r) => { res.writeHead(r.statusCode, r.headers); r.pipe(res); });
        fwd.write(body); fwd.end();
      });
    });
    server.listen(proxyPort, () => resolve(server));
  });
}

await createProxy("primary", PRIMARY_PORT, PROXY_PRIMARY);
await createProxy("secondary", SECONDARY_PORT, PROXY_SECONDARY);

// Start Hive Router
const router = spawn(ROUTER_BIN, [], { cwd: `${process.cwd()}/hive-config`, stdio: ["ignore", "pipe", "pipe"] });
let ready = false;
for (let i = 0; i < 30; i++) {
  await new Promise((r) => setTimeout(r, 1000));
  try {
    const r = await fetch(`http://127.0.0.1:${ROUTER_PORT}/graphql`, {
      method: "POST", headers: { "Content-Type": "application/json" },
      body: JSON.stringify({ query: "{ __typename }" }), signal: AbortSignal.timeout(2000),
    });
    if (r.status < 500) { ready = true; break; }
  } catch {}
}
if (!ready) { console.error("Router failed to start"); process.exit(1); }

// Send test query
proxyQueries.primary.length = 0;
proxyQueries.secondary.length = 0;

const testQuery = `{ getResponse { message actions { ... on Alpha { id value } ... on Beta { id name details } ... on Gamma { id label } } } }`;
const resp = await fetch(`http://127.0.0.1:${ROUTER_PORT}/graphql`, {
  method: "POST", headers: { "Content-Type": "application/json" },
  body: JSON.stringify({ query: testQuery }),
});
const data = await resp.json();
await new Promise((r) => setTimeout(r, 500));

// Report
console.log("Query sent to primary subgraph:");
for (const q of proxyQueries.primary.filter(q => !q.includes("_service"))) console.log("  " + q);
console.log("\nGateway response:", JSON.stringify(data, null, 2));

router.kill();
await primaryServer.stop();
await secondaryServer.stop();
process.exit(0);
```

### Expected Output (demonstrating the bug)

```
Query sent to primary subgraph:
  {getResponse{message actions{__typename ...on Alpha{id value}}}}

Gateway response:
{
  "data": {
    "getResponse": {
      "message": "ok",
      "actions": [
        { "id": "a1", "value": "alpha-val" },
        { "id": null, "name": null, "details": null },
        { "id": null, "label": null }
      ]
    }
  }
}
```

Note: `...on Beta` and `...on Gamma` are completely absent from the subgraph query. The primary subgraph never sees those fragments, so it cannot return Beta/Gamma fields.

dotansimha:
Fixes https://github.com/graphql-hive/router/issues/1098 and adds lifetimes in code (agreed with Dotan to include it in the PR - we can now allocate much much less).

Let me explain the changes in a few points.

## Per-subgraph union member edges in the graph

Previously, `UnionDefinitions::intersections` computed the intersection of all union members across all subgraphs, and only some members were turned into `AbstractMove` edges. This meant subgraph-specific members were eliminated from the graph.

Now we no longer do the intersection of union members when building the graph. A subgraph that knows `Action = Common | OnlyA` gets edges only to `Common` and `OnlyA`, and a subgraph that knows `Action = Common | OnlyB` gets edges only to `Common` and `OnlyB`.

The intersection decision lives now in the walker.

## `UnionContext` in `OperationPath`

Each `OperationPath` contains an optional `UnionContext`, that has:
- `parent_type_name` and `field_name` - union field that was traversed
- `graph_id` - which subgraph we're in
- `member_name` - union member the path resolved to
- `possible_members` - the list of union members available in this subgraph

The context is set when a `FieldMove` edge represents a field returning a union type.

## Per-path union members

In `find_direct_paths` and `find_indirect_paths`, when the navigation target is a `ConcreteType` (like `... on OnlyB`), the path finder checks if the type can be resolved, before even exploring edges. If there's `UnionContext` of course.

## Intersection

The intersection of members is done in the walker module now, and it's part of `narrow_partial_union_paths` function. This function performs the intersection of `possible_members` from paths, from all subgraphs.

- drops paths where current member is not in the intersection
- replaces the `possible_members` with the intersection
- if the intersection is empty, drops all paths as no member is allowed

### DIFF
--- a/.changeset/partial_unions.md
+++ b/.changeset/partial_unions.md
@@ -7,6 +7,6 @@ node-addon: patch
 
 # Improve handling of unions
 
-The query planner improves handling of union types whose members vary between subgraphs. Previously, the planner always computed an intersection of union members, ignoreing subgraph-specific members.
+The query planner improves handling of union types whose members vary between subgraphs. Previously, the planner always computed an intersection of union members, ignoring subgraph-specific members.
 
 Fixes [#1098](https://github.com/graphql-hive/router/issues/1098)
--- a/.changeset/partial_unions.md
+++ b/.changeset/partial_unions.md
@@ -1,0 +1,12 @@
+---
+hive-router-query-planner: patch
+hive-router-plan-executor: patch
+hive-router: patch
+node-addon: patch
+---
+
+# Improve handling of unions
+
+The query planner improves handling of union types whose members vary between subgraphs. Previously, the planner always computed an intersection of union members, ignoreing subgraph-specific members.
+
+Fixes [#1098](https://github.com/graphql-hive/router/issues/1098)
--- a/bin/dev-cli/src/main.rs
+++ b/bin/dev-cli/src/main.rs
@@ -16,7 +16,6 @@ use hive_router_query_planner::planner::plan_nodes::QueryPlan;
 use hive_router_query_planner::planner::query_plan::build_query_plan_from_fetch_graph;
 use hive_router_query_planner::planner::tree::query_tree::QueryTree;
 use hive_router_query_planner::planner::walker::walk_operation;
-use hive_router_query_planner::planner::walker::ResolvedOperation;
 use hive_router_query_planner::planner::QueryPlannerOptions;
 use hive_router_query_planner::state::supergraph_state::OperationKind;
 use hive_router_query_planner::state::supergraph_state::SupergraphState;
@@ -61,8 +60,17 @@ fn main() {
             println!("{}", graph);
         }
         "paths" => {
-            let (graph, best_paths_per_leaf, _operation, _supergraph_state) =
-                process_paths(&args[2], &args[3]);
+            let (graph, operation, supergraph_state) = load_graph_operation(&args[2], &args[3]);
+            let override_context = PlannerOverrideContext::default();
+            let cancellation_token = CancellationToken::new();
+            let best_paths_per_leaf = walk_operation(
+                &graph,
+                &supergraph_state,
+                &override_context,
+                &operation,
+                &cancellation_token,
+            )
+            .unwrap();
 
             for (index, best_path) in best_paths_per_leaf
                 .root_field_groups
@@ -163,12 +171,7 @@ fn process_fetch_graph(
 }
 
 fn process_plan(supergraph_path: &str, operation_path: &str) -> QueryPlan {
-    let supergraph_sdl =
-        std::fs::read_to_string(supergraph_path).expect("Unable to read input file");
-    let parsed_schema = parse_schema(&supergraph_sdl);
-    let supergraph = SupergraphState::new(&parsed_schema);
-    let graph = Graph::graph_from_supergraph_state(&supergraph).expect("failed to create graph");
-    let operation = get_operation(operation_path, &supergraph);
+    let (graph, operation, supergraph) = load_graph_operation(supergraph_path, operation_path);
     let override_context = PlannerOverrideContext::default();
     let cancellation_token = CancellationToken::new();
 
@@ -204,9 +207,17 @@ fn process_merged_tree(
     supergraph_path: &str,
     operation_path: &str,
 ) -> (Graph, QueryTree, SupergraphState, OperationKind) {
-    let (graph, best_paths_per_leaf, operation, supergraph_state) =
-        process_paths(supergraph_path, operation_path);
+    let (graph, operation, supergraph_state) = load_graph_operation(supergraph_path, operation_path);
+    let override_context = PlannerOverrideContext::default();
     let cancellation_token = CancellationToken::new();
+    let best_paths_per_leaf = walk_operation(
+        &graph,
+        &supergraph_state,
+        &override_context,
+        &operation,
+        &cancellation_token,
+    )
+    .unwrap();
     let query_tree =
         find_best_combination(&graph, best_paths_per_leaf, &cancellation_token).unwrap();
 
@@ -230,31 +241,16 @@ fn get_operation(operation_path: &str, supergraph: &SupergraphState) -> Operatio
     operation.clone()
 }
 
-fn process_paths(
+fn load_graph_operation(
     supergraph_path: &str,
     operation_path: &str,
-) -> (
-    Graph,
-    ResolvedOperation,
-    OperationDefinition,
-    SupergraphState,
-) {
+) -> (Graph, OperationDefinition, SupergraphState) {
     let supergraph_sdl =
         std::fs::read_to_string(supergraph_path).expect("Unable to read input file");
     let parsed_schema = parse_schema(&supergraph_sdl);
     let supergraph = SupergraphState::new(&parsed_schema);
     let graph = Graph::graph_from_supergraph_state(&supergraph).expect("failed to create graph");
     let operation = get_operation(operation_path, &supergraph);
-    let override_context = PlannerOverrideContext::default();
-    let cancellation_token = CancellationToken::new();
-    let best_paths_per_leaf = walk_operation(
-        &graph,
-        &supergraph,
-        &override_context,
-        &operation,
-        &cancellation_token,
-    )
-    .unwrap();
 
-    (graph, best_paths_per_leaf, operation, supergraph)
+    (graph, operation, supergraph)
 }
--- a/bin/dev-cli/src/main.rs
+++ b/bin/dev-cli/src/main.rs
@@ -207,7 +207,8 @@ fn process_merged_tree(
     supergraph_path: &str,
     operation_path: &str,
 ) -> (Graph, QueryTree, SupergraphState, OperationKind) {
-    let (graph, operation, supergraph_state) = load_graph_operation(supergraph_path, operation_path);
+    let (graph, operation, supergraph_state) =
+        load_graph_operation(supergraph_path, operation_path);
     let override_context = PlannerOverrideContext::default();
     let cancellation_token = CancellationToken::new();
     let best_paths_per_leaf = walk_operation(
--- a/lib/query-planner/src/graph/mod.rs
+++ b/lib/query-planner/src/graph/mod.rs
@@ -10,14 +10,13 @@ mod tests;
 use std::{
     collections::{HashMap, HashSet},
     fmt::{Debug, Display},
-    hash::Hash,
 };
 
 use super::ast::normalization::utils::extract_type_condition;
 use crate::{
     ast::type_aware_selection::TypeAwareSelection,
     federation_spec::FederationRules,
-    graph::node::{SubgraphTypeSpecialization, UnionSubsetData},
+    graph::node::{SubgraphTypeSpecialization, UnionMembersData},
     state::supergraph_state::{
         OperationKind, SupergraphDefinition, SupergraphField, SupergraphState,
     },
@@ -36,7 +35,11 @@ use super::graph::{edge::Edge, node::Node};
 
 type InnerGraph = Petgraph<Node, Edge, Directed>;
 
-type UnionRegistyHashMap<'a> = HashMap<&'a str, HashMap<&'a str, HashSet<&'a str>>>;
+type UnionTypeName<'a> = &'a str;
+type SubgraphName<'a> = &'a str;
+type UnionMemberTypes<'a> = HashSet<&'a str>;
+type UnionRegistyHashMap<'a> =
+    HashMap<UnionTypeName<'a>, HashMap<SubgraphName<'a>, UnionMemberTypes<'a>>>;
 
 #[derive(Debug, Default)]
 struct UnionDefinitions<'a> {
@@ -52,7 +55,7 @@ impl<'a> UnionDefinitions<'a> {
             .iter()
             .filter(|(_, d)| matches!(d, SupergraphDefinition::Union(_)))
         {
-            let mut in_subgraphs: HashMap<&'a str, HashSet<&'a str>> = HashMap::new();
+            let mut in_subgraphs: HashMap<SubgraphName<'a>, UnionMemberTypes<'a>> = HashMap::new();
 
             for join_member in definition.join_union_members() {
                 in_subgraphs
@@ -61,7 +64,7 @@ impl<'a> UnionDefinitions<'a> {
                         e.insert(&join_member.member);
                     })
                     .or_insert_with(|| {
-                        let mut set: HashSet<&'a str> = HashSet::new();
+                        let mut set: UnionMemberTypes<'a> = HashSet::new();
                         set.insert(&join_member.member);
                         set
                     });
@@ -79,63 +82,37 @@ impl<'a> UnionDefinitions<'a> {
         self.registry.contains_key(type_name)
     }
 
-    fn members_in_subgraph(&self, type_name: &str, graph: &'a str) -> Option<&HashSet<&'a str>> {
+    fn members_in_subgraph(&self, type_name: &str, graph: &str) -> Option<&UnionMemberTypes<'a>> {
         self.registry.get(type_name).and_then(|r| r.get(graph))
     }
 
-    /// Produces a list of names of the object types.
-    pub fn intersections(
+    /// Produces the union members visible from a field resolved in a subgraph
+    pub fn members_for_field_in_graph(
         &self,
-        type_def: &'a SupergraphDefinition,
         field_def: &'a SupergraphField,
         field_type: &str,
-    ) -> HashSet<String> {
+        graph_id: &str,
+    ) -> UnionMemberTypes<'a> {
         // Collect subgraphs the field was defined in.
         // First, look for join__field(graph:),
         // If not defined, look at type's join__type(graph:).
-
-        let mut members_per_subgraph: HashMap<&str, HashSet<String>> = HashMap::new();
-
-        if field_def.join_field.is_empty() {
-            for join_type in type_def.join_types() {
-                let mut members_in_subgraph: HashSet<String> = HashSet::new();
-                for union_member in self
-                    .members_in_subgraph(field_type, &join_type.graph_id)
-                    .unwrap()
-                {
-                    members_in_subgraph.insert(union_member.to_string());
+        if let Some(join_field) = field_def.join_field.iter().find(|join_field| {
+            join_field
+                .graph_id
+                .as_ref()
+                .is_some_and(|field_graph_id| field_graph_id == graph_id)
+        }) {
+            if let Some(type_in_graph) = join_field.type_in_graph.as_ref().map(|t| t.inner_type()) {
+                // join__field(type:) can narrow a union-returning field to one concrete member.
+                if type_in_graph != field_type {
+                    return UnionMemberTypes::from([type_in_graph]);
                 }
-
-                members_per_subgraph.insert(&join_type.graph_id, members_in_subgraph);
             }
         }
 
-        for join_field in field_def.join_field.iter() {
-            if let Some(graph_id) = join_field.graph_id.as_ref() {
-                let mut members_in_subgraph: HashSet<String> = HashSet::new();
-
-                if let Some(type_in_graph) =
-                    join_field.type_in_graph.as_ref().map(|t| t.inner_type())
-                {
-                    // look for join__field(type:) - it could point to `Object` or `Union`
-                    if type_in_graph != field_type {
-                        // the field_type is a union, as we previously checked,
-                        // so if the type_in_graph is different,
-                        // it means it's an object type (one of the members).
-                        members_in_subgraph.insert(type_in_graph.to_string());
-                        members_per_subgraph.insert(graph_id, members_in_subgraph);
-                        continue;
-                    }
-                }
-
-                for union_member in self.members_in_subgraph(field_type, graph_id).unwrap() {
-                    members_in_subgraph.insert(union_member.to_string());
-                }
-                members_per_subgraph.insert(graph_id, members_in_subgraph);
-            }
-        }
-
-        intersections(members_per_subgraph.values().collect())
+        self.members_in_subgraph(field_type, graph_id)
+            .cloned()
+            .unwrap_or_default()
     }
 }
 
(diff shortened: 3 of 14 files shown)


bin/dev-cli/src/main.rs
```diff
@@ -61,8 +60,17 @@ fn main() {
             println!("{}", graph);
         }
         "paths" => {
-            let (graph, best_paths_per_leaf, _operation, _supergraph_state) =
-                process_paths(&args[2], &args[3]);
+            let (graph, operation, supergraph_state) = load_graph_operation(&args[2], &args[3]);
```
kamilkisiela, line 63: Needed to avoid lifetimes problems

lib/query-planner/src/graph/mod.rs
```diff
@@ -36,7 +35,11 @@ use super::graph::{edge::Edge, node::Node};
 
 type InnerGraph = Petgraph<Node, Edge, Directed>;
 
-type UnionRegistyHashMap<'a> = HashMap<&'a str, HashMap<&'a str, HashSet<&'a str>>>;
+type UnionTypeName<'a> = &'a str;
+type SubgraphName<'a> = &'a str;
+type UnionMemberTypes<'a> = HashSet<&'a str>;
+type UnionRegistyHashMap<'a> =
```
kamilkisiela, line 41: Clean up of &str with no meaning

lib/query-planner/src/graph/mod.rs
```diff
@@ -79,63 +82,37 @@ impl<'a> UnionDefinitions<'a> {
...
 
-    fn members_in_subgraph(&self, type_name: &str, graph: &'a str) -> Option<&HashSet<&'a str>> {
+    fn members_in_subgraph(&self, type_name: &str, graph: &str) -> Option<&UnionMemberTypes<'a>> {
         self.registry.get(type_name).and_then(|r| r.get(graph))
     }
 
-    /// Produces a list of names of the object types.
-    pub fn intersections(
+    /// Produces the union members visible from a field resolved in a subgraph
+    pub fn members_for_field_in_graph(
```
kamilkisiela, line 90: We no longer perform intersection of union members. This function returns union members for a field of a type from a subgraph.

lib/query-planner/src/graph/mod.rs
```diff
@@ -782,17 +769,18 @@ impl Graph {
                                 target_type,
                                 state.resolve_graph_id(graph_id)?,
                                 state.is_interface_object_in_subgraph(target_type, graph_id),
-                                SubgraphTypeSpecialization::UnionSubset(UnionSubsetData {
+                                SubgraphTypeSpecialization::UnionMembers(UnionMembersData {
                                     type_name: def_name.clone(),
                                     field_name: field_name.clone(),
-                                    object_type_name: member.clone(),
+                                    object_type_name: member.to_string(),
+                                    possible_members: possible_members.clone(),
```
kamilkisiela, line 776: We store possible members of the union type for the subgraph this node belongs

lib/query-planner/src/graph/node.rs
```diff
@@ -3,28 +3,38 @@ use std::fmt::{Debug, Display};
...
-pub struct UnionSubsetData {
+pub struct UnionMembersData {
     /// Represents the type owning the field
     pub type_name: String,
     /// Represents the field resolving a union type
     pub field_name: String,
     /// Represents a union member
     pub object_type_name: String,
+    /// Represents all union members reachable for the same field in this subgraph.
+    pub possible_members: Vec<String>,
```
kamilkisiela, line 14: We store possible members of the union type for the subgraph this node belongs

lib/query-planner/src/graph/node.rs
```diff
@@ -3,28 +3,38 @@ use std::fmt::{Debug, Display};
...
-    /// and from that tail we create an abstract-move edges to the object types.
-    /// (type_name, field_name, union_member_name)
-    UnionSubset(UnionSubsetData),
+    /// Node represents a union member tail for a specific subgraph.
+    ///
+    /// For union-returning field moves, we may need a tail that only exposes the
+    /// members reachable in the current subgraph. We model that by creating
+    /// per-member specialized nodes and then abstract-move edges from those tails
+    /// to the concrete member types.
+    UnionMembers(UnionMembersData),
```
kamilkisiela, line 28: Just a rename of things, as UnionSubset is too generic

lib/query-planner/src/graph/node.rs
```diff
@@ -3,28 +3,38 @@ use std::fmt::{Debug, Display};
...
+}
+
+impl SubgraphTypeSpecialization {
+    pub fn union_members_data(&self) -> Option<&UnionMembersData> {
```
kamilkisiela, line 32: Helper to extract data easily instead of importing types and do pattern matching

lib/query-planner/src/graph/node.rs
```diff
@@ -121,6 +131,19 @@ impl Node {
             Node::SubgraphType(st) => Some(&st.subgraph.0),
         }
     }
+
+    pub fn subgraph_type(&self) -> Option<&SubgraphType> {
```
kamilkisiela, line 135: Helper to extract data easily instead of importing types and do pattern matching

lib/query-planner/src/planner/fetch/fetch_graph.rs
```diff
@@ -1626,6 +1626,7 @@ fn find_satisfiable_key<'a>(
                 last_segment: None,
                 visited_edge_indices: Default::default(),
                 cost: 0,
+                union_context: None,
```
kamilkisiela, line 1629: union context is a form of state that tells us: "hey, you're now resolving a field that returns a union and here are all your union member options".

When not resolving union-type fields, it's None

lib/query-planner/src/planner/walker/mod.rs
```diff
@@ -392,21 +410,124 @@ fn process_inline_fragment<'a>(
...
+///
+/// - graph A: Action = Common | OnlyA
+/// - graph B: Action = Common | OnlyB
+///
+/// If the planner still has valid candidate paths through both A and B,
+/// the response shape must not depend on which path wins later.
+/// In that case we keep only members that are available in every candidate graph.
+///
+/// In the example above, only `Common` is safe to keep.
+fn narrow_partial_union_paths<'graph>(paths: &mut Vec<OperationPath<'graph>>) {
```
kamilkisiela, line 424: This is where we do the intersection now, when needed.

lib/query-planner/src/planner/walker/path.rs
```diff
@@ -40,6 +40,35 @@ pub struct PathSegment {
     pub condition: Option<Condition>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UnionContext<'graph> {
```
kamilkisiela, line 44: This is the struct I mentioned before that holds a state where we're resolving a union-type field.

lib/query-planner/src/planner/walker/path.rs
```diff
@@ -40,6 +40,35 @@ pub struct PathSegment {
...
+    pub parent_type_name: &'graph str,
+    pub field_name: &'graph str,
+    pub graph_id: &'graph str,
+    pub member_name: &'graph str,
+    pub possible_members: Vec<&'graph str>,
```
kamilkisiela, line 49: This part is dynamic. When we eliminate some members, this gets updated.

lib/query-planner/src/planner/walker/path.rs
```diff
@@ -141,7 +176,51 @@ impl OperationPath {
         };
         let new_segment = Rc::new(new_segment_data);
 
-        OperationPath::new(self.root_node, Some(new_segment), new_visited)
+        let union_context = match edge_ref.weight() {
```
kamilkisiela, line 179: This is where the `UnionContext` gets updated. We do it when advancing the `OperationPath`.

lib/query-planner/src/planner/walker/path.rs
```diff
@@ -141,7 +176,51 @@ impl OperationPath {
...
+            Edge::FieldMove(_) => {
+                let tail = graph.node(edge_ref.target()).ok();
+                let union_data = tail.and_then(|tail| tail.union_members_data());
+                let graph_id = tail.and_then(|tail| tail.graph_id());
+
+                match (union_data, graph_id) {
+                    (Some(union_data), Some(graph_id)) => Some(UnionContext {
```
kamilkisiela, line 186: If the field is returning a union type, we create the UnionContext

lib/query-planner/src/planner/walker/path.rs
```diff
@@ -141,7 +176,51 @@ impl OperationPath {
...
+                        field_name: &union_data.field_name,
+                        graph_id,
+                        member_name: &union_data.object_type_name,
+                        possible_members: union_data
+                            .possible_members
+                            .iter()
+                            .map(|member| member.as_str())
+                            .collect(),
+                    }),
+                    _ => None,
```
kamilkisiela, line 197: If it's not, then no UnionContext for the advanced path

lib/query-planner/src/planner/walker/path.rs
```diff
@@ -141,7 +176,51 @@ impl OperationPath {
...
+                }
+            }
+            Edge::AbstractMove(member_type_name) => self
```
kamilkisiela, line 200: If it's an abstract move, we then narrow the possible members to the target type.

lib/query-planner/src/planner/walker/path.rs
```diff
@@ -141,7 +176,51 @@ impl OperationPath {
...
+                .union_context
+                .as_ref()
+                .map(|scope| scope.narrow_to_member(member_type_name)),
+            Edge::EntityMove(_) | Edge::InterfaceObjectTypeMove(_) => None,
```
kamilkisiela, line 204: When there's an entity move, we reset

lib/query-planner/src/planner/walker/path.rs
```diff
@@ -278,5 +357,12 @@ impl OperationPath {
             previous_new_segment,
             self.visited_edge_indices.clone(),
         )
+        // The requirement continuation reuses other's tail, so it also must reuse its union context
+        .with_union_context(other.union_context.clone())
+    }
+
+    fn with_union_context(mut self, scope: Option<UnionContext<'graph>>) -> Self {
```
kamilkisiela, line 364: This is just a nicer way of doing:

```rust
let mut path = ...;
path.union_context = context;
path
```

lib/query-planner/src/planner/walker/pathfinder.rs
```diff
@@ -143,6 +156,13 @@ impl PathSearch<'_> {
             .graph_id()
             .ok_or(WalkOperationError::TailMissingInfo(tail_node_index))?;
 
+        // Respect the path's current union scope when targeting a concrete type.
+        if let NavigationTarget::ConcreteType(type_name, _) = target {
+            if !path.can_resolve_union_member(type_name) {
+                return Ok(Vec::new());
```
kamilkisiela, line 162: This is where we check whether a type is part of possible members

lib/query-planner/src/planner/walker/pathfinder.rs
```diff
@@ -372,26 +394,33 @@ pub fn find_self_referencing_direct_path(
...
+    ) -> Result<Vec<OperationPath<'graph>>, WalkOperationError> {
         let graph = self.graph;
-        let mut result: Vec<OperationPath> = vec![];
+        let mut result: Vec<OperationPath<'graph>> = vec![];
         let path_tail_index = path.tail();
 
+        // Respect the path's current union scope when targeting a concrete type.
+        if let NavigationTarget::ConcreteType(type_name, _) = target {
+            if !path.can_resolve_union_member(type_name) {
+                return Ok(result);
```
kamilkisiela, line 420: This is where we check whether a type is part of possible members

lib/query-planner/src/graph/mod.rs
dotansimha: great

.changeset/partial_unions.md
```diff
@@ -0,0 +1,12 @@
+---
+hive-router-query-planner: patch
+hive-router-plan-executor: patch
+hive-router: patch
+node-addon: patch
+---
+
+# Improve handling of unions
+
+The query planner improves handling of union types whose members vary between subgraphs. Previously, the planner always computed an intersection of union members, ignoreing subgraph-specific members.
```
dotansimha: *ignoring

lib/query-planner/src/graph/node.rs
dotansimha: Yes, good call

lib/query-planner/src/planner/walker/mod.rs
```diff
@@ -392,21 +410,124 @@ fn process_inline_fragment<'a>(
...
+    type GraphId<'graph> = &'graph str;
+    type MembersPerGraph<'graph> = Vec<(GraphId<'graph>, Vec<&'graph str>)>;
+    let mut members_per_graph: MembersPerGraph<'graph> = Vec::new();
+
+    for path in paths.iter() {
+        let context = path
+            .union_context
+            .as_ref()
+            // It's safe to unwrap as `all_same_field` checked that all paths have the same context
+            .expect("union member context should exist at this point");
```
dotansimha: nit: I think you can avoid all of these repeated `unwrap`? I know it's verified before, but you can eliminate and get the concrete Some and just avoid it. It's repeated in multiple places here...

kamilkisiela (approved):

PR #1117, opened 2026-06-10
at-least-once sampling

kamilkisiela:
The goal is to allow users to keep a low reporting sample rate while still making sure every distinct operation is visible in Hive Console at least once.

Before this change, Usage Reporting had a single top-level `sample_rate` option. That works well for reducing amount of usage reports, but it means some operations can be completely missed.

```yaml
telemetry:
  hive:
    usage_reporting:
      enabled: true
      sampling:
        rate: "10%"
        at_least_once:
          key:
            - operation_name
            - operation_body
          max_distinct_keys: 12000
```

## Breaking Change in configuration

```diff
telemetry:
  hive:
    usage_reporting:
-      sample_rate: "10%"
+      sampling:
+        rate: "10%"
```
---

## Behavior changes

Usage Reporting now runs exclusion before sampling. That means excluded operations are dropped immediately and do not affect sampling and its seen keys. Crucial to not affect at-least-once sampling.

---

## Changes in SDK

The sampling decision now lives in SDK. The same applies to exclusion over operation names.

## Seen-key storage

At-least-once sampling tracks hashes of distinct keys, not the full operation body strings, names or types. This way we store u64 (8 bytes) instead of random size of bytes (operation body or name could end up eating lots of memory).

I intentionally did not use a cache like Moka here. We do not need its features, as the only behavior we need is to:

- check whether a `u64` was seen
- insert if absent
- evict the oldest key when the storage is full

The FIFO Set I did holds 16 bytes per key so it's super cheap and efficient.

## Apollo Router Plugin 

The feature request was for Hive Router to support at-least-once sampling, that's why I did not touch our plugin for Apollo's Router. I also tried to keep the diff low (it could be lower but formatting changed a bit) and amount of changes as low as possible. Nothing here prevents us from adopting the same functionality in the plugin at some point.

---

Closes #990

The main difference between this PR and #990 is the scope of the changes.

#990 makes a larger refactor, moves configuration types into the SDK, introduces a tagged sampler, updates the Apollo Router plugin configuration etc. This PR takes a more focused approach.

The goal here is to add support for at-least-once sampling in Hive Router's Usage Reporting. I modified only the parts needed for that feaure: updating the sampling configuration in Hive Router, moving the sampling into the SDK, exclusions before sampling.

The seen-key storage is also implemented differently. The previous implementation had a flaw. Instead of storing full strings such as operation names, operation bodies, or VRL-generated values, this PR hashes the key into `u64` and stores that in a FIFO set. 

It does that to keep memory usage tiny, especially in cases where users choose keys such as `operation_body`, which can be quite huge.
I also chose not to use Moka in this implementation, because it has 10x bigger memory footprint per key and we want to store a lot of keys at small cost.

I left Apollo Router plugin untouched, since the feature request was about Hive Router.

### DIFF
--- a/.changeset/at_least_once_sampling.md
+++ b/.changeset/at_least_once_sampling.md
@@ -1,0 +1,34 @@
+---
+hive-router: minor
+hive-router-config: minor
+hive-console-sdk: minor
+hive-router-internal: patch
+hive-router-plan-executor: patch
+hive-apollo-router-plugin: patch
+---
+
+# Add at-least-once sampling for Usage Reporting
+
+Hive Router now supports at-least-once sampling for Usage Reporting.
+
+This feature is useful when you want to keep a low sampling rate, but still make sure all operations are visible in Hive at least once.
+
+The first request for each unique key is always reported. Later requests for the same key follow the configured sampling `rate`.
+
+Example configuration:
+
+```yaml
+telemetry:
+  hive:
+    usage_reporting:
+      enabled: true
+      sampling:
+        rate: "10%" # 10% of operations will be reported
+        at_least_once:
+          key: # the combination of operation's name and body makes the request unique
+            - operation_name
+            - operation_body
+          max_distinct_keys: 12000 # how many keys to track and hold in memory
+```
+
+Keys are tracked in memory, up to `max_distinct_keys` (default: `100_000`). Every key takes approximately 16 bytes of memory.
--- a/bin/router/src/pipeline/usage_reporting.rs
+++ b/bin/router/src/pipeline/usage_reporting.rs
@@ -6,16 +6,17 @@ use std::{
 use async_trait::async_trait;
 use graphql_tools::parser::schema::Document;
 use hive_console_sdk::agent::usage_agent::{
-    AgentError, ExecutionReport, OperationType, RequestDetails, UsageAgent, UsageAgentExt,
+    AgentError, ExecutionReport, OperationType, RequestDetails, SamplingKey, UsageAgent,
+    UsageAgentExt,
 };
 use hive_router_config::{
+    headers::OneOrMany,
     telemetry::hive::{is_slug_target_ref, is_uuid_target_ref, HiveTelemetryConfig},
-    usage_reporting::{UsageReportingConfig, UsageReportingExclude},
+    usage_reporting::{UsageReportingExclude, UsageReportingSamplingKeyKind},
 };
 use hive_router_internal::background_tasks::{BackgroundTask, BackgroundTasksManager};
 use hive_router_internal::telemetry::utils::resolve_value_or_expression;
 use hive_router_query_planner::state::supergraph_state::OperationKind;
-use rand::prelude::*;
 use tokio_util::sync::CancellationToken;
 
 use crate::consts::ROUTER_VERSION;
@@ -65,6 +66,7 @@ pub fn init_hive_usage_agent(
         .user_agent(user_agent)
         .endpoint(usage_config.endpoint.clone())
         .token(access_token)
+        .sample_rate(usage_config.sampling.rate.as_f64())
         .buffer_size(usage_config.buffer_size)
         .connect_timeout(usage_config.connect_timeout)
         .request_timeout(usage_config.request_timeout)
@@ -77,6 +79,20 @@ pub fn init_hive_usage_agent(
 
     if let Some(UsageReportingExclude::Expression { expression }) = &usage_config.exclude {
         agent_builder = agent_builder.exclude_expression(expression.clone());
+    }
+
+    if let Some(UsageReportingExclude::OperationNames(operation_names)) = &usage_config.exclude {
+        agent_builder = agent_builder.exclude_operation_names(operation_names.clone());
+    }
+
+    if let Some(at_least_once) = &usage_config.sampling.at_least_once {
+        agent_builder = agent_builder.at_least_once_sampling(
+            match &at_least_once.key {
+                OneOrMany::One(kind) => vec![map_sampling_key_to_sdk(kind)],
+                OneOrMany::Many(kinds) => kinds.iter().map(map_sampling_key_to_sdk).collect(),
+            },
+            at_least_once.max_distinct_keys,
+        );
     }
 
     let agent = agent_builder.build()?;
@@ -97,21 +113,9 @@ pub async fn collect_usage_report<'a>(
     operation_kind: Option<&'a OperationKind>,
     operation_body: &'a str,
     hive_usage_agent: &UsageAgent,
-    usage_config: &UsageReportingConfig,
     error_count: usize,
     request_details: Option<RequestDetails>,
 ) {
-    let sample_rate = usage_config.sample_rate.as_f64();
-    if sample_rate < 1.0 && !rand::rng().random_bool(sample_rate) {
-        return;
-    }
-    if let Some(operation_name) = operation_name {
-        if let Some(UsageReportingExclude::OperationNames(excluded_names)) = &usage_config.exclude {
-            if excluded_names.iter().any(|name| name == operation_name) {
-                return;
-            }
-        }
-    }
     let timestamp = SystemTime::now()
         .duration_since(UNIX_EPOCH)
         .unwrap()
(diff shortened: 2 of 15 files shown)


lib/hive-console-sdk/src/helpers.rs
```diff
@@ -0,0 +1,111 @@
...
+impl FifoSet {
+    pub fn new(capacity: usize) -> Self {
+        assert!(capacity > 0 && capacity < u32::MAX as usize);
+        Self {
+            eviction_order: Vec::new(),
+            seen: HashSet::default(),
+            next_evict: 0,
+            capacity: capacity as u32,
+        }
+    }
```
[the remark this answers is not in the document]
kamilkisiela: Limited to 10M, but kept the assertion with a nice error message

lib/router-config/src/usage_reporting.rs
```diff
@@ -12,6 +13,73 @@ pub enum UsageReportingExclude {
...
+    /// into one key.
+    ///
+    /// No default value.
+    pub key: UsageReportingSamplingKey,
+
+    #[serde(default = "default_max_distinct_keys")]
+    /// Maximum number of unique keys kept in memory for at-least-once sampling.
+    /// When the limit is reached, older keys may be removed.
+    ///
+    /// Every key consumes 8 bytes of memory.
```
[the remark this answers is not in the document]
kamilkisiela: corrected

lib/hive-console-sdk/src/agent/usage_agent.rs
```diff
@@ -285,22 +363,28 @@ impl UsageAgentExt for UsageAgent {
...
+            tracing::debug!(
+                "Sampling dropped report for operation \"{}\"",
+                execution_report
+                    .operation_name
+                    .clone()
+                    .or_else(|| Some("anonymous".to_string()))
+                    .unwrap()
+            );
+            return Ok(());
         }
```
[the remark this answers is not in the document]
kamilkisiela: did that, makes sense

lib/hive-console-sdk/src/agent/usage_agent.rs
```diff
@@ -372,6 +456,35 @@ impl From<RequestDetails> for VrlValue {
...
+                SamplingKey::OperationBody => report.operation_body.as_str(),
+            };
+            value.hash(&mut hasher);
+            0u8.hash(&mut hasher);
+        }
+
+        hasher.finish()
+    }
+
+    /// Marks the given key hash as seen, returning true if it was already seen.
```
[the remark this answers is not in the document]
kamilkisiela: Corrected

bin/router/src/pipeline/usage_reporting.rs
```diff
@@ -65,6 +66,7 @@ pub fn init_hive_usage_agent(
         .user_agent(user_agent)
         .endpoint(usage_config.endpoint.clone())
         .token(access_token)
+        .sample_rate(usage_config.sampling.rate.as_f64())
```
kamilkisiela, line 69: Sampling is now moved to SDK

lib/hive-console-sdk/src/agent/builder.rs
```diff
@@ -26,9 +30,25 @@ pub struct UsageAgentBuilder {
     retry_policy: ExponentialBackoff,
     user_agent: Option<String>,
     circuit_breaker: Option<AsyncRecloser>,
-    exclude_expression: Option<String>,
+    exclude: Option<BuilderExclude>,
```
kamilkisiela, line 33: `exclude` now covers expressions and operation names

lib/hive-console-sdk/src/agent/usage_agent.rs
```diff
@@ -285,22 +363,26 @@ impl UsageAgentExt for UsageAgent {
...
-                    "Excluding report for operation \"{}\" based on exclude expression evaluation",
-                    execution_report
-                        .operation_name
-                        .clone()
-                        .or_else(|| Some("anonymous".to_string()))
-                        .unwrap()
-                );
-                return Ok(());
-            }
+        if inner.should_exclude(&execution_report, request.as_ref())? {
```
kamilkisiela, line 366: I wrapped exclusion logic into `should_exclude` fn of the struct, so it's cleaner and the logic is self-contained

lib/hive-console-sdk/src/agent/usage_agent.rs
```diff
@@ -285,22 +363,26 @@ impl UsageAgentExt for UsageAgent {
...
+                "Excluding report for operation \"{}\" based on exclude expression evaluation",
+                execution_report
+                    .operation_name
+                    .as_deref()
+                    .unwrap_or("anonymous")
+            );
+            return Ok(());
+        }
+
+        if !inner.should_sample(&execution_report) {
```
kamilkisiela, line 377: I wrapped sampling logic into `should_sample` fn of the struct, so it's cleaner and the logic is self-contained

lib/hive-console-sdk/src/agent/usage_agent.rs
```diff
@@ -372,6 +454,35 @@ impl From<RequestDetails> for VrlValue {
     }
 }
 
+impl AtLeastOnceSampling {
```
kamilkisiela, line 457: The logic behind at-least-once sampling is almost entirely in this struct

lib/hive-console-sdk/src/helpers.rs
```diff
@@ -0,0 +1,114 @@
...
+///
+/// The idea here is to keep the least amount of memory footprint possible,
+/// while still providing fast lookup and eviction of the oldest entry when full.
+///
+/// Why this struct and not let's say Moka?
+///
+/// Moka's Cache eats 256 bytes per entry,
+/// This implementation takes 16 bytes per entry.
+/// For 100k entries, this implementation uses ~1.6MB, compared to ~24MB for Moka.
+pub struct FifoSet {
```
kamilkisiela, line 36: Here I explained why FIFO. We needed a bounded set and alternatives are memory expensive, and we only need limited size + u64 for hashes, so it's bare minimum.

lib/router-config/src/usage_reporting.rs
```diff
@@ -12,6 +13,73 @@ pub enum UsageReportingExclude {
...
+    ///  - `operation_name`: the name of the GraphQL operation
+    ///  - `operation_type`: the type
+    ///  - `operation_body`: the body
+    ///
+    ///
+    /// You can also provide multiple values. In that case, the router combines them
+    /// into one key.
+    ///
+    /// No default value.
+    pub key: UsageReportingSamplingKey,
```
kamilkisiela, line 42: Expression can be added at any point. The goal was to provide at-least-once sampling with ability to configure a custom key for every operation. The requirement was the operation's name, but it's not enough on its own so I added 3 options for users to choose.

It's also faster because it's precompiled.

If we ever get a request to have the key highly customizable, we can always add expressions that run for every request.

lib/router-config/src/usage_reporting.rs
```diff
@@ -12,6 +13,73 @@ pub enum UsageReportingExclude {
...
+    /// Possible values:
+    ///  - `operation_name`: the name of the GraphQL operation
+    ///  - `operation_type`: the type
+    ///  - `operation_body`: the body
+    ///
+    ///
+    /// You can also provide multiple values. In that case, the router combines them
+    /// into one key.
+    ///
+    /// No default value.
```
kamilkisiela, line 41: I debated with myself whether to have a default or not, but in the user's feature request the ask was for sampling based on operation's name, which is imo not enough for a key representing a unique operation.

I would do body + name, but did not want to decide for the user, so I left it for the user to decide.

dotansimha (approved):

lib/router-config/src/usage_reporting.rs
dotansimha: no default ✅ user needs to know when they use

.changeset/at_least_once_sampling.md
```diff
@@ -0,0 +1,34 @@
+---
+hive-router: minor
+hive-router-config: minor
+hive-console-sdk: minor
+hive-router-internal: patch
+hive-router-plan-executor: patch
+hive-apollo-router-plugin: patch
```
dotansimha, line 7: needed because of sdk bump?

kamilkisiela: `.changeset/release-sync.sh` did that changeset

PR #1122, opened 2026-06-12
fix(router): skip null entities in batched fetches

kamilkisiela:
Fixes batched entity fetching when the entity list contains `null` items.

The executor skipped `null` entities when building the batch of representations, but when the subgraph response was processed the logic still assumed every original list item had a matching result. We ended up using incorrect indexes (out of bounds).

Now `null` entities are preserved as empty slots, they are not sent to the subgraph, and the subgraph response processing logic skips those same slots when applying the results.

### DIFF
--- a/.changeset/null_entity_fix.md
+++ b/.changeset/null_entity_fix.md
@@ -1,0 +1,6 @@
+---
+hive-router-plan-executor: patch
+hive-router: patch
+---
+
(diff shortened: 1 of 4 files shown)


kamilkisiela:
@gemini-code-assist review

dotansimha (approved):

PR #1133, opened 2026-06-15
Header propagation on failure paths

### LINKED ISSUE #1088: Header manipulation does not run for errors responses
The `HeaderRules.response` config block isn’t documented as having scope conditions, but in practice the rules only fire on the successful execution path. 

Error responses (validation errors, auth failures, plugin rejections, 5xx) bypass them entirely. This wasn’t obvious from the docs and cost us some implementation time — flagging in case a docs clarification would be welcome.

kamilkisiela:
Closes #1088

This PR makes response header propagation run consistently across successful responses, partial GraphQL error responses, and execution failures. Previously, response header propagation was executed only on the successful executions.

Keeps `content-type` header excluded from response header propagation, as it's decided by Router based on `Accept`.

Preserves subgraph response headers on executor errors:
- empty response body
- response body read failure
- response deserialization failure
- subgraph's internal server errors

> [!WARNING]  
> WebSocket response header propagation is not covered and is left as a TODO

---

Introduces a per-request `ResponseHeaderSink` used to collect response headers. Applies collected response headers at the final router -> client response, including deduped request responses.

### DIFF
--- a/.changeset/header_propagation_non_happy.md
+++ b/.changeset/header_propagation_non_happy.md
@@ -1,0 +1,8 @@
+---
+hive-router: patch
+hive-router-plan-executor: patch
+---
+
+# Fix response header propagation on error paths
+
+Response header rules now run consistently for successful responses, partial GraphQL error responses, deduped requests, and execution failures.
--- a/bin/router/src/lib.rs
+++ b/bin/router/src/lib.rs
@@ -14,6 +14,7 @@ mod utils;
 
 use std::ops::ControlFlow;
 use std::sync::Arc;
+use tracing::error;
 
 use crate::{
     consts::ROUTER_VERSION,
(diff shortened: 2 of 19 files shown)


dotansimha:
@kamilkisiela 
While working on demand control PR, I needed a way to return response headers from the pipeline flow, and I didn't have a way to do that. Now with that new Sink, do you think it could cover this case as well? (so we won't have two different ways to do that) 

See https://github.com/graphql-hive/router/blob/b2073f4b227cd098a1ad64269a99d2bb41c3621f/bin/router/src/pipeline/error.rs#L169 
And https://github.com/graphql-hive/router/blob/b2073f4b227cd098a1ad64269a99d2bb41c3621f/bin/router/src/pipeline/error.rs#L217-L223

dotansimha (approved):

PR #1136, opened 2026-06-15
bug: introspection leaks fields hidden by on_graphql_validation + with_schema override

praguevara:
## What this demonstrates

When a plugin overrides the validation schema via `on_graphql_validation` / `with_schema`, the router correctly rejects queries that use removed fields. However, `__type` and `__schema` introspection still resolve against the **full supergraph**, so the hidden fields remain fully visible through introspection.

This PR adds a **failing test** to the `feature_flags` example that reproduces the leak. It is intentionally not a fix — the goal is to agree on the right approach before sending a full diff.

## How to reproduce

```
cargo test --manifest-path plugin_examples/feature_flags/Cargo.toml \
  introspection_should_hide_fields_removed_by_validation_schema_override
```

**Failure output:**

```
thread 'test::tests::introspection_should_hide_fields_removed_by_validation_schema_override'
  panicked at '`shippingEstimate` leaked through introspection even though the feature is disabled.
Actual response:
{
  "data": {
    "__type": {
      "fields": [
        { "name": "upc" },
        { "name": "weight" },
        { "name": "price" },
        { "name": "inStock" },        <-- should be hidden
        { "name": "shippingEstimate" }, <-- should be hidden
        { "name": "name" },
        { "name": "reviews" },
        ...
      ]
    }
  }
}'
```

## Context

We are migrating the commercetools GraphQL federation gateway to Hive Router. One customization presents each tenant a **filtered view of the supergraph** — elements gated behind inactive `@feature` directives are removed. For this to work correctly, validation **and** introspection must present the same filtered surface; otherwise the hidden elements are still discoverable via `__type`.

The `feature_flags` example is the exact pattern we use (minus the external tags service). The current behaviour — `with_schema` silently not covering introspection — is the blocker.

### DIFF
--- a/plugin_examples/feature_flags/src/test.rs
+++ b/plugin_examples/feature_flags/src/test.rs
@@ -3,6 +3,57 @@ mod tests {
     use e2e::testkit::{ClientResponseExt, TestRouter, TestSubgraphs};
     use hive_router::ntex;
 
+    /// `on_graphql_validation` + `with_schema` swaps the schema used for
+    /// validation, but `__type`/`__schema` introspection still resolves against
+    /// the full supergraph today, so feature-gated fields remain visible through
+    /// introspection even when they are blocked for queries.
+    ///
+    /// This test FAILS on stock hive-router main — see the assertion message for
+    /// the actual leaked output.  The desired behaviour is for introspection to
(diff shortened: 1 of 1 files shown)


praguevara:
Moving this to https://github.com/praguevara/router/pull/1 to avoid noise

PR #1155, opened 2026-06-17
feat(executor): log subgraph WebSocket connect and handshake failures…

Meemaw:
## Summary

Surface subgraph subscription failures (WebSocket handshake, HTTP connect, SSE stream, etc.) in the router logs at `error` level. Previously these were silent on the router side — the only signal was the `SubgraphExecutorError` returned to the client, while the router itself logged nothing above `debug`.

In practice this meant a client seeing

```
WebSocket protocol handshake with 'ws://<subgraph>:.../subscriptions' failed: Connection closed before acknowledgement
```

while operators looking at router logs at the default `info` level saw nothing.

## Why subscriptions specifically

Non-subscription executor errors already log through `QueryPlanExecution::log_error` at `plan.rs:1135`. The subscription path skips that handler — it yields `FailedExecutionResult` directly to the client stream at `plan.rs:312-323` and `plan.rs:362-368` without ever invoking it. That gap is what makes failed WebSocket subscriptions invisible to operators.

## Changes

`lib/executor/src/execution/plan.rs`:

- Extract the body of `log_error()` into a free `log_plan_execution_error(error: &PlanExecutionError)` helper, sharing the same format string as before (`"Error executing plan with subgraph '{}': {}"` when the subgraph name is known, otherwise `"Error executing plan: {}"`).
- Call it at the two subscription-stream error yield sites in `execute_query_plan` before yielding the `FailedExecutionResult`.
- `QueryPlanExecution::log_error()` now delegates to the same helper, so non-subscription paths log identically to before.

`lib/executor/src/executors/websocket.rs`:

- An earlier iteration of this PR added `error!` logs at the four `WsClient` connect / handshake failure sites; those have been removed in favour of the central handler above. No behavior change vs. `main` in this file.

## Why this approach over per-executor logging

- Matches the existing pattern: `log_error` is already the central handler for non-subscription executor errors. Subscriptions were the one path that bypassed it.
- Covers HTTP-callback and SSE subscription failures with the same fix, not just WebSocket.
- Keeps `lib/executor/src/executors/websocket.rs` purely transport-level, no logging-policy concerns.

## Test plan

- [ ] Point a subgraph at a refused WebSocket endpoint and confirm one `error` line with `subgraph` and the connect error appears (instead of the previous silence).
- [ ] Point at a subgraph that closes the socket after the upgrade without sending `connection_ack` and confirm the handshake-failure `error` line appears.
- [ ] Same check with HTTP-callback and SSE subscription failures — confirm they now also produce an error log.
- [ ] Confirm non-subscription subgraph errors continue to log exactly as on `main` (format string unchanged).
- [ ] Confirm a healthy subscription path produces no new error logs at steady state.

### DIFF
--- a/.changeset/log_subscription_errors.md
+++ b/.changeset/log_subscription_errors.md
@@ -1,0 +1,8 @@
+---
+hive-router-plan-executor: patch
+hive-router: patch
+---
+
+# Log subgraph subscription failures at error level
+
+Subgraph subscription failures (WebSocket handshake, HTTP-callback connect, SSE stream, etc.) are now logged at `error` level via the central `plan.rs` handler, matching how non-subscription subgraph errors are already logged. Previously these failures only reached the client; the router itself logged nothing above `debug`.
--- a/.changeset/log_subscription_errors.md
+++ b/.changeset/log_subscription_errors.md
@@ -1,6 +1,5 @@
 ---
 hive-router-plan-executor: patch
-hive-router: patch
 ---
 
 # Log subgraph subscription failures at error level
--- a/lib/executor/src/execution/plan.rs
+++ b/lib/executor/src/execution/plan.rs
@@ -187,6 +187,18 @@ fn early_http_response_into_execution_output(
     }
 }
 
+fn log_plan_execution_error(error: &PlanExecutionError) {
+    if let Some(subgraph_name) = error.subgraph_name() {
+        tracing::error!(
+            "Error executing plan with subgraph '{}': {}",
+            subgraph_name,
+            error
+        );
+    } else {
+        tracing::error!("Error executing plan: {}", error);
+    }
+}
+
 pub async fn execute_query_plan<'exec>(
     opts: QueryPlanExecutionOpts<'exec>,
 ) -> Result<QueryPlanExecutionResult, PlanExecutionError> {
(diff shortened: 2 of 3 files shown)


Meemaw:
Thanks for the review. Pushed [27e823a](https://github.com/graphql-hive/router/pull/1155/commits/27e823a) which addresses the centralization feedback: the four `error!` calls in `lib/executor/src/executors/websocket.rs` are gone, and the logging now lives in `plan.rs` alongside the existing `QueryPlanExecution::log_error` handler. Both paths share a new free `log_plan_execution_error` helper. Side benefit: HTTP-callback and SSE subscription failures are now logged the same way, not just WebSocket.

Note on the cited rule — I couldn't find a "centralize error logging" rule in `.gemini/styleguide.md`. The actual project pattern (e.g. `executors/http_callback.rs:236,261`, `executors/coprocessor/runtime.rs`) is to log at the executor level too. I still moved the logging because `plan.rs` already centralizes non-subscription executor errors via `log_error`, and the subscription path was the one gap — so this happens to be the right call here for a different reason than the styleguide claim.

enisdenjo (approved):

dotansimha (approved):

dotansimha:
Thanks @Meemaw ! Can I please ask you to add a `changeset` to the executor? It could be added by installing `knope` and then running `knope changeset`. bumping the executor should be just fine, then it will propagate to the router automatically.

Meemaw:
Done in e6e6f04 — dropped `hive-router: patch` from the changeset so only `hive-router-plan-executor` is bumped, and the router consumer picks it up transitively.

dotansimha:
Thanks @Meemaw ! 
We got another PR merged right before this one and I think it has conflicts now - can you please take a look?

Meemaw:
Rebased onto upstream main. Only conflict was around the helper I added in `plan.rs` — upstream's #1133 added `early_http_response_into_execution_output` next to where I placed `log_plan_execution_error`. Kept both, no behavioral changes.

dotansimha (approved):

PR #1157, opened 2026-06-17
feat(query-planner): experimental fold any matching interface (opt-in)

kamilkisiela:
This PR adds an opt-in query planner optimization behind:

```yaml
query_planner:
  experimental_abstract_type_folding: true
```

or

```rust
QUERY_PLANNER_EXPERIMENTAL_ABSTRACT_TYPE_FOLDING=true
```

The option is disabled by default, so existing query planning behavior stays unchanged unless users explicitly enable it.

The feature folds multiple matching concrete object inline fragments into a shared interface fragment in subgraph fetches, even when that interface is not the declared return type of the parent field.

Example `product-service` subgraph fetch:

```graphql
query {
  products {
    ... on Book  { id title }
    ... on Movie { id title }
  }
}
```

is turned into:

```diff
query {
  products {
-    ... on Book  { id title }
-    ... on Movie { id title }
+    ... on Media { id title }
  }
}
```

The `products` field returns `Product` interface, but one object-type member of this interface called `Album` is not present in the query, therefore `... on Product {...}` is not possible to use (default behavior). With the feature flag enabled, both fragments are folded into `... on Media { ... }`, because `Book` and `Movie` are the only members of the `Media` interface in the `product-service` subgraph.

---

The main internal change is that the query planner now accepts `QueryPlannerOptions`:

```rust
pub struct QueryPlannerOptions {
    pub experimental_abstract_type_folding: bool,
}
```

Those options are stored on `Planner` and passed down into fetch graph optimization, specifically into `fold_concrete_selections_to_interfaces`.

The previous behavior folded concrete object-type fragments only into the expected interface matching return type of the parent field.

Closes #1149

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -302,6 +302,8 @@ jobs:
   federation-audit:
     name: test / federation-audit
     runs-on: ubuntu-latest
+    env:
+      GATEWAY_TIMEOUT: 1800000 # 30 minutes, to prevent CI timeout for long-running tests
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
       - name: setup rust
@@ -312,8 +314,6 @@ jobs:
           node-version-file: .node-version
       - name: Run Audits
         run: npm run test:federation-all
-        env:
-          GATEWAY_TIMEOUT: 1800000 # 30 minutes, to prevent CI timeout for long-running tests
         working-directory: audits
         continue-on-error: true # continue even if the test fails
       - name: Upload logs
@@ -332,3 +332,9 @@ jobs:
           comment_mode: off
           files: |
             audits/reports/*.xml
+      - name: Run Audits (experimental_abstract_type_folding)
+        run: npm run test:federation-all
+        env:
+          QUERY_PLANNER_EXPERIMENTAL_ABSTRACT_TYPE_FOLDING: true
+        working-directory: audits
+        continue-on-error: true # continue even if the test fails
--- a/bin/router/src/schema_state.rs
+++ b/bin/router/src/schema_state.rs
@@ -28,9 +28,8 @@ use hive_router_plan_executor::{
     response::graphql_error::GraphQLError,
     SubgraphExecutorMap,
 };
-use hive_router_query_planner::planner::plan_nodes::QueryPlan;
 use hive_router_query_planner::{
-    planner::{Planner, PlannerError},
+    planner::{plan_nodes::QueryPlan, Planner, PlannerError, QueryPlannerOptions},
     utils::parsing::safe_parse_schema,
 };
 use moka::future::Cache;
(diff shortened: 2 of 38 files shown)


.github/workflows/ci.yaml
```diff
@@ -302,6 +302,8 @@ jobs:
   federation-audit:
     name: test / federation-audit
     runs-on: ubuntu-latest
+    env:
+      GATEWAY_TIMEOUT: 1800000 # 30 minutes, to prevent CI timeout for long-running tests
```
kamilkisiela, line 306: now the `GATEWAY_TIMEOUT` is shared for two audit runs

.github/workflows/ci.yaml
```diff
@@ -332,3 +332,9 @@ jobs:
           comment_mode: off
           files: |
             audits/reports/*.xml
+      - name: Run Audits (experimental_abstract_type_folding)
+        run: npm run test:federation-all
+        env:
+          QUERY_PLANNER_EXPERIMENTAL_ABSTRACT_TYPE_FOLDING: true
```
kamilkisiela, line 338: second audit run that tests Router when the FF is enabled

lib/query-planner/src/tests/testkit.rs
```diff
@@ -54,10 +54,11 @@ pub fn read_supergraph(fixture_path: &str) -> String {
     std::fs::read_to_string(supergraph_path).expect("Unable to read input file")
 }
 
-pub fn build_query_plan_with_context(
+pub fn build_query_plan(
```
kamilkisiela, line 57: I unified those build query plan functions and now we have two variants:

- `build_query_plan` that accepts all the things with no defaults
- `build_query_plan_with_defaults` that accepts only bare minimum and uses defaults for everything

We had 2 variants and now with the options we would have had 3. Ridiculous.

lib/router-config/src/env_overrides.rs
```diff
@@ -55,6 +55,10 @@ pub struct EnvVarOverrides {
     // Tracing overrides
     #[envconfig(from = "TELEMETRY_TRACING_SAMPLING_RATE")]
     pub tracing_sampling_rate: Option<f64>,
+
+    // Query planner overrides
+    #[envconfig(from = "QUERY_PLANNER_EXPERIMENTAL_ABSTRACT_TYPE_FOLDING")]
```
kamilkisiela, line 60: env var

lib/router-config/src/query_planner.rs
```diff
@@ -40,3 +52,7 @@ fn default_query_planning_allow_expose() -> bool {
 fn default_query_planning_timeout() -> Duration {
     Duration::from_secs(10)
 }
+
+fn default_experimental_abstract_type_folding() -> bool {
+    false
```
kamilkisiela, line 57: opt-in

bin/router/src/schema_state.rs
```diff
@@ -280,7 +279,14 @@ impl SchemaState {
         parsed_supergraph_sdl: Document,
         callback_subscriptions: CallbackSubscriptionsMap,
     ) -> Result<SupergraphData, SupergraphManagerError> {
-        let planner = Planner::new_from_supergraph(&parsed_supergraph_sdl)?;
+        let planner = Planner::new_from_supergraph(
```
kamilkisiela, line 282: This is where Router talks with Query Planner

lib/node-addon/src/lib.rs
```diff
@@ -30,9 +30,10 @@ impl QueryPlanner {
         let parsed_supergraph =
             safe_parse_schema(&supergraph_sdl).map_err(QueryPlanError::SchemaParse)?;
 
-        let planner = Planner::new_from_supergraph(&parsed_supergraph).map_err(|err| {
-            napi::Error::from_reason(format!("Failed to create query planner: {}", err))
-        })?;
+        let planner = Planner::new_from_supergraph(&parsed_supergraph, Default::default())
```
kamilkisiela, line 33: JS side stays unchanged

lib/query-planner/src/tests/testkit.rs
dotansimha: perfect

dotansimha (approved):

PR #1162, opened 2026-06-17
fix(executor): proper handling of `null` and non-null propagation

### LINKED ISSUE #1110: Non-nullable fields return null instead of error-propagating when an entity is unresolved in its owning subgraph (Apollo Router compatibility gap)
## Summary

When an entity reference cannot be resolved by its owning subgraph (its `__resolveReference` returns `null`), Hive Router returns a response in which **non-nullable fields are set to `null`**, violating the supergraph schema. Apollo Router, by contrast, correctly performs null value completion / error propagation and nullifies up to the nearest nullable parent.

> ⚠️ **The core problem: `email` is declared `String!` in the schema but is returned as `null`. The response does not conform to the schema, and no error is reported** — a client trusting the schema's non-null guarantee receives invalid data silently.

Since Hive Router advertises full compatibility with Apollo Router, this is a behavioral divergence on a schema-correctness guarantee.

**Version:** `hive-router@0.0.65`

## Minimal reproduction

Two subgraphs. Subgraph **A** owns `Order` and references `User` (providing only its `id`). Subgraph **B** owns `User`, including the non-nullable `email`.

**Subgraph A (`orders`)**
```graphql
type Query {
  orders: [Order!]!
}

type User @key(fields: "id", resolvable: false) {
  id: ID! @external
}

type Order @key(fields: "id") {
  id: ID!
  user: User! @provides(fields: "id")
}
```

**Subgraph B (`users`)**
```graphql
type User @key(fields: "id") {
  id: ID!
  name: String      # nullable
  email: String!    # NON-nullable
}
```

**Resolvers (key part):** subgraph A returns an order whose `user.id` (e.g. `"5"`) does **not** exist in subgraph B, and subgraph B's reference resolver returns `null` for a missing user (the behavior recommended by Apollo's docs):
```js
// Subgraph B
User: {
  __resolveReference: (ref) => users.find((u) => u.id === ref.id) ?? null,
}
```

### Query

```graphql
query ExampleQuery {
  orders {
    id
    user {
      id
      name
      email
    }
  }
}
```

### Actual behavior (Hive Router `0.0.51`)

```json
{
  "data": {
    "orders": [
      { "user": { "id": "5", "name": null, "email": null } }
    ]
  }
}
```

> ❌ **`email` is declared `String!` but is returned as `null` — the response does not conform to the schema, and no error is reported.**
>
> This is the heart of the issue: the response is silently invalid against the published supergraph schema. A client relying on the non-null contract has no way to detect that the data is malformed.

### Expected behavior (matches Apollo Router)

Apollo Router reports null-value-completion errors and propagates `null` up to the nearest nullable position (here, all the way to `data: null`):

```json
{
  "data": null,
  "extensions": {
    "valueCompletion": [
      { "message": "Null value found for non-nullable type String", "path": ["orders", 0, "user"] },
      { "message": "Null value found for non-nullable type User",   "path": ["orders", 0, "user"] },
      { "message": "Null value found for non-nullable type Order",  "path": ["orders", 0] },
      { "message": "Null value found for non-nullable type [Order!]", "path": ["orders"] }
    ]
  }
}
```

## Notes

- For the same setup but requesting only nullable fields (`user { id name }`), **both** routers return a schema-valid response (`name: null`). The divergence only appears once a **non-nullable** field of the unresolved entity is selected.
- `@provides(fields: "id")` on subgraph A means the `User` is partially materialized (its `id` is known before the entity fetch). This appears relevant: the router seems to keep the partial `User` object alive via the provided `id` and leaves the other non-nullable fields as `null` instead of treating the entity fetch returning `null` as a non-null violation.

### LINKED ISSUE #1154: Non-null field errors does not propagate correctly
Per the GraphQL spec (https://spec.graphql.org/draft/#sec-Handling-Field-Errors), when a non-null field errors, the `null` must bubble up to the nearest nullable ancestor. 

For a non-null *root* field with no nullable ancestor, the `null` propagates all the way to `data`, so the response must be `"data": null`.

At the moment, this doesn't happen correctly, and the error remain on the same level where the error happened.

dotansimha:
## Overview

This PR implements null propagation for fields, executed during projection of fields. Since `null` can be returned by a subgraph (for example: in an entity call), we need to validate that the final response conforms to the contract and the definition of each field.

From the GraphQL spec:
> If during [ExecuteCollectedFields](https://spec.graphql.org/September2025/#ExecuteCollectedFields())() a [response position](https://spec.graphql.org/September2025/#response-position) with a non-null type raises an [execution error](https://spec.graphql.org/September2025/#execution-error) then that error must propagate to the parent response position (the entire selection set in the case of a field, or the entire list in the case of a list position), either resolving to null if allowed or being further propagated to a parent response position.

> If this occurs, any sibling response positions which have not yet executed or have not yet yielded a value may be cancelled to avoid unnecessary work.

And:
> Since Non-Null response positions cannot be null, execution errors are propagated to be handled by the parent [response position](https://spec.graphql.org/September2025/#response-position). If the parent response position may be null then it resolves to null, otherwise if it is a Non-Null type, the execution error is further propagated to its parent [response position](https://spec.graphql.org/September2025/#response-position).

> If a List type wraps a Non-Null type, and one of the [response position](https://spec.graphql.org/September2025/#response-position) elements of that list resolves to null, then the entire list [response position](https://spec.graphql.org/September2025/#response-position) must resolve to null. If the List type is also wrapped in a Non-Null, the execution error continues to propagate upwards.

> If every [response position](https://spec.graphql.org/September2025/#response-position) from the root of the request to the source of the execution error has a Non-Null type, then the "data" entry in the [execution result](https://spec.graphql.org/September2025/#execution-result) should be null.

If this occurs, any sibling response positions which have not yet executed or have not yet yielded a value may be cancelled to avoid unnecessary work.

## Overview of the changes 

* In projection plan, we now store a new `struct FieldNullability` instead of `is_non_null: bool`, since we need to know the inner/outer nullability in case of list-fields.
* `project_selection_set` and `project_selection_set_with_map` now returns a `Decision` value that indicates the caller (= parent selection) on the value it used. If we encounter a `null` field in a field that doesn't allow to have null `!`), then the parent will process and collapse. This bubbles up all the way to `data` in case on a chain of non-null fields.

Fixes https://github.com/graphql-hive/router/issues/1154
Fixes https://github.com/graphql-hive/router/issues/1110

### DIFF
--- a/bin/router/src/pipeline/authorization/collector.rs
+++ b/bin/router/src/pipeline/authorization/collector.rs
@@ -165,7 +165,7 @@ pub(super) fn collect_authorization_statuses<'op>(
                 // to trigger null bubbling behavior (which invalidates the entire response).
                 let status = if let Some(field_info) = type_fields.and_then(|f| f.get(&field.name))
                 {
-                    if field_info.is_non_null {
+                    if field_info.nullability.is_non_null() {
                         has_non_null_unauthorized = true;
                         FieldAuthStatus::UnauthorizedNonNullable
                     } else {
@@ -261,7 +261,7 @@ fn process_field_selection<'op, 'ctx>(
 
     let status = if is_authorized {
         FieldAuthStatus::Authorized
-    } else if field_info.is_non_null {
+    } else if field_info.nullability.is_non_null() {
         FieldAuthStatus::UnauthorizedNonNullable
     } else {
         FieldAuthStatus::UnauthorizedNullable
--- a/lib/executor/src/introspection/schema.rs
+++ b/lib/executor/src/introspection/schema.rs
@@ -6,10 +6,60 @@ use graphql_tools::parser::{
 };
 use hive_router_query_planner::consumer_schema::ConsumerSchema;
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct FieldTypeInfo {
     pub output_type_name: String,
-    pub is_non_null: bool,
+    pub nullability: FieldNullability,
+}
+
+/// The nullability shape of a field's type. Similar to the GraphQL type wrappers.
+#[derive(Debug, Clone)]
+pub enum FieldNullability {
+    Leaf {
+        non_null: bool,
+    },
+    List {
+        non_null: bool,
+        item: Box<FieldNullability>,
+    },
+}
+
+impl FieldNullability {
+    // __typename: String! (https://spec.graphql.org/September2025/#sec-Type-Name-Introspection)
+    #[inline]
+    pub fn type_name() -> Self {
+        FieldNullability::Leaf { non_null: true }
+    }
+
+    fn leaf(non_null: bool) -> Self {
+        FieldNullability::Leaf { non_null }
+    }
+
+    fn list(non_null: bool, item: FieldNullability) -> Self {
+        FieldNullability::List {
+            non_null,
+            item: Box::new(item),
+        }
+    }
+
+    /// Whether this position's value is Non-Null, so a `null` here must bubble to its parent.
+    #[inline]
+    pub fn is_non_null(&self) -> bool {
+        match self {
+            FieldNullability::Leaf { non_null } | FieldNullability::List { non_null, .. } => {
+                *non_null
+            }
+        }
+    }
+
+    /// The element's nullability shape, when this position is a list.
+    #[inline]
+    pub fn list_item(&self) -> Option<&FieldNullability> {
+        match self {
+            FieldNullability::List { item, .. } => Some(item),
+            FieldNullability::Leaf { .. } => None,
+        }
+    }
 }
 
 #[derive(Debug, Default)]
@@ -114,7 +164,7 @@ impl SchemaWithMetadata for ConsumerSchema {
                             field.name.to_string(),
                             FieldTypeInfo {
                                 output_type_name: field_type_name,
-                                is_non_null: field.field_type.is_non_null(),
+                                nullability: field.field_type.field_nullability(),
                             },
                         );
                     }
@@ -136,7 +186,7 @@ impl SchemaWithMetadata for ConsumerSchema {
                             field.name.to_string(),
                             FieldTypeInfo {
                                 output_type_name: field_type_name,
-                                is_non_null: field.field_type.is_non_null(),
+                                nullability: field.field_type.field_nullability(),
                             },
                         );
                     }
@@ -196,6 +246,7 @@ impl SchemaWithMetadata for ConsumerSchema {
 
 trait TypeName {
     fn type_name(&self) -> String;
+    fn field_nullability(&self) -> FieldNullability;
 }
 
 impl TypeName for Type<'_, String> {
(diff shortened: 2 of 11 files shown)


lib/executor/src/projection/response.rs
```diff
@@ -194,41 +212,57 @@ fn project_selection_set<'a>(
     buffer: &mut Vec<u8>,
     parent_type_name: TypeName<'a>,
     schema_metadata: &'a SchemaMetadata,
-) -> Result<(), ProjectionError> {
+    nullability: &'a FieldNullability,
+) -> Result<NullPropagationDecision, ProjectionError> {
     match data {
         Value::Array(arr) => {
+            let null_propagation_checkpoint = buffer.len();
+            let list_item_nullability = nullability.list_item();
```
dotansimha, line 222: If it's a list, we get the inner one, and use that after. This way inner nulls and nested lists work

lib/executor/src/projection/response.rs
```diff
@@ -238,29 +272,43 @@ fn project_selection_set<'a>(
                         &mut first,
                         schema_metadata,
                     )?;
+
+                    if null_propagation_decision.should_propagate() {
```
dotansimha, line 278: If the child propagated, we continue to bubble up.

lib/executor/src/projection/response.rs
```diff
@@ -344,12 +393,19 @@ fn project_selection_set_with_map<'a>(
                                 buffer,
                                 parent_type_name.clone(),
                                 schema_metadata,
-                            )?;
+                                &plan.nullability,
+                            )?
                         } else {
                             // If the field is not found in the object, set it to Null
                             buffer.put(NULL);
+                            NullPropagationDecision::PropagateNonNull
```
dotansimha: This one is tricky: if the field is not found, the projection logic right now is using `null` for it. 

We treat this as something that needs to bubble up, as it's invalid state and we don't want to silently do `null`.

lib/executor/src/projection/response.rs
```diff
@@ -344,12 +393,19 @@ fn project_selection_set_with_map<'a>(
...
                         }
                     }
+                };
+
+                // A `null` value in a non-null position bubbles up
+                if null_propagation_decision.should_propagate() && plan.nullability.is_non_null() {
```
dotansimha, line 409: this is the gate that makes the decision if we need to bubble more or not.

bin/router/src/pipeline/authorization/collector.rs
```diff
@@ -165,7 +165,7 @@ pub(super) fn collect_authorization_statuses<'op>(
                 // to trigger null bubbling behavior (which invalidates the entire response).
                 let status = if let Some(field_info) = type_fields.and_then(|f| f.get(&field.name))
                 {
-                    if field_info.is_non_null {
+                    if field_info.nullability.is_non_null() {
```
dotansimha, line 168: just a tiny refactor because i changed the way it's stored (bool vs struct)

lib/executor/src/introspection/schema.rs
```diff
@@ -9,7 +9,63 @@ use hive_router_query_planner::consumer_schema::ConsumerSchema;
 #[derive(Debug, Default)]
 pub struct FieldTypeInfo {
     pub output_type_name: String,
-    pub is_non_null: bool,
+    pub nullability: FieldNullability,
+}
+
+/// The nullability shape of a field's type. Similar to the GraphQL type wrappers.
+#[derive(Debug, Clone)]
+pub enum FieldNullability {
```
dotansimha, line 17: initially i did `is_non_null: bool, list_nullability: Vec<bool>` (and then `[true, false, true, false]`. But it's hard to read. A nested struct works well

kamilkisiela (approved):

PR #1172, opened 2026-06-19
feat(cache-control): restrictive merge of subgraph Cache-Control headers

### LINKED ISSUE #1022: restrictive Cache-Control header propagation policy
## Summary
  Please add support for **propagating `Cache-Control` headers from subgraph responses using a restrictive merge policy**, configurable globally and per-subgraph.

  The current header merge options (`first` / `last` / `append`) are not safe for `Cache-Control`: since subgraph resolution order is non-deterministic, we either get the wrong directive (`first` /
  `last`) or multiple conflicting directives (`append`). This can result in private data being cached publicly.

  ## Scenarios we need to handle correctly
  **Scenario 1 — private data must not be cached**
  - Subgraph A: `Cache-Control: max-age=0, no-cache, no-store, private`
  - Subgraph B: `Cache-Control: public, max-age=300`
  - Expected: `Cache-Control: no-store, no-cache` (the strictest wins).

  **Scenario 2 — pick the shortest TTL**
  - Subgraph A: `Cache-Control: public, max-age=500`
  - Subgraph B: `Cache-Control: public, max-age=300`
  - Expected: `Cache-Control: public, max-age=300`.
  
  **Scenario 3 — any subgraph error disables caching**
  - Subgraph A: returns an error
  - Subgraph B: `Cache-Control: public, max-age=300`
  - Expected: `Cache-Control: no-store, no-cache, must-revalidate`.

  None of these are achievable today with `first` / `last` / `append`.
  ## Why this matters
  - **Data security.** Without restrictive merging, a `private` / `no-store` response from one subgraph can be silently overridden by a permissive `public, max-age=N` from another, causing CDNs and browsers to cache data that was explicitly marked uncacheable.
  - **Cache correctness.** An aggregated response is only as cacheable as its strictest participant.
  - **Operational impact.** Teams running federated graphs that span public + private data domains currently have no safe option — they must either disable Cache-Control propagation entirely (losing
   CDN benefits) or accept the risk.

  ## Prior art — WunderGraph Cosmo
  Cosmo already ships exactly this feature. Their algorithm:
  1. `no-cache` / `no-store` from any subgraph wins over everything else.
  2. Otherwise, the smallest `max-age` across subgraphs is selected.
  3. If `Expires` is present, the earliest value is used.
  4. Mutations are automatically `no-cache` when the policy is enabled.
  5. Responses containing errors are automatically `no-store, no-cache, must-revalidate`.
  Example config:

```code
  cache_control_policy:
    enabled: true
    value: "max-age=180, public"
    subgraphs:
          • name: "products"
        value: "max-age=60, public"
          • name: "pricing"
        value: "no-cache"
```
  Docs: https://cosmo-docs.wundergraph.com/router/configuration/cache-control-policy

enisdenjo:
Closes #1022

Implementataion of restrictive merging of `Cache-Control` headers across all subgraph responses before forwarding to the client. The most conservative directive always wins. No opt-in required.

### Why always-on?

Making this opt-in means an illinformed or unaware developer gets unsafe behavior by default: a `public` subgraph can silently override a `private` or `no-cache` subgraph and accidentally cache responses that should not be cached. Or even worse, a mutation or an errored response getting cached because some subgraph has caching set up.

Everyone propagating `Cache-Control` headers should be merging them using the restrictive policies - otherwise they are facing dangers like those stated in #1022 - you'd never want the unsafe behaviour (as far as I can think of).

The safe behavior should be the default behavior, and it should be enforced.

### Merge algorithm

Given N subgraph responses:

1. Any `no-store`, `no-cache`, or `private` directive short-circuits the result to `no-store, no-cache` immediately.
1. Otherwise `max-age` = minimum of all present values (missing = ignored).
1. `public` only when every subgraph is `public`.
1. `must-revalidate` if any subgraph sets it.

Forces `no-store, no-cache, must-revalidate` regardless of subgraph headers when:

- Any subgraph executor error (network failure, bad status, etc.)
- Any GraphQL-level error in a subgraph response (`errors` array non-empty)
- The operation is a mutation

Completely removes the `Cache-Control` header when no subgraph sends a valid `Cache-Control`, like non-UTF-8 bytes or parinsg returns `None` (invalid/empty/whitespace-only string).

### Configuration

Users augment cache behavior through existing header rules:

```yaml
headers:
  all:
    response:
      # forward subgraph cache-control; append keeps every value for the merge
      - propagate:
          named: cache-control
          algorithm: append
          # default emitted unless a subgraph provides one
          default: "public, max-age=180"
  subgraphs:
    pricing:
      response:
        # pin this subgraph regardless of what it returns
        - insert:
            name: cache-control
            value: "no-cache"
```

Subgraph-level `insert` or `remove` rules let operators pin or drop a specific subgraph's contribution before the merge runs.

We're using the full powers of header rules to augment the Cache-Control. No need for a separate config option because the rules already provide the full set of primitives, and we're opting everyone in to restrictive merging by design.

Configuring `propagate: named: cache-control` without merging is rejected at compile time with an error that goes straight to the point: propagating `Cache-Control` across subgraphs without the restrictive merge is unsafe.

If no propagation of the `Cache-Control` is set up - the merge is inactive. This is the safer approach because you have to _not_ propagate the headers to disable.

### TODO

- [x] docs https://github.com/graphql-hive/docs/pull/135

### DIFF
--- a/lib/executor/src/headers/cache_control.rs
+++ b/lib/executor/src/headers/cache_control.rs
@@ -1,5 +1,6 @@
 use crate::headers::{plan::HeaderAggregationStrategy, response::ResponseHeaderAggregator};
 use http::HeaderValue;
+use tracing::warn;
 
 #[derive(Clone, Default)]
 struct ParsedCacheControl {
@@ -175,12 +176,18 @@ pub fn finalize(aggregator: &mut ResponseHeaderAggregator, force_no_store: bool)
     }
 
     if let Some(merged) = acc {
-        let value = HeaderValue::from_str(&to_header_value(&merged))
-            .unwrap_or_else(|_| HeaderValue::from_static("no-store"));
+        let serialized = to_header_value(&merged);
+        // safety: to_header_value only produces ASCII
+        let value = HeaderValue::from_str(&serialized).expect("to_header_value produced non-ASCII");
         aggregator.entries.insert(
             http::header::CACHE_CONTROL,
             (HeaderAggregationStrategy::Last, vec![value]),
         );
+    } else {
+        // no valid values found, but there were cache-control headers
+        // do the safe thing and graceful thing - completely omit the header
+        warn!("no valid cache-control values found, removing header");
+        aggregator.entries.remove(&http::header::CACHE_CONTROL);
     }
 }
 
@@ -681,6 +688,36 @@ mod tests {
         let mut agg = ResponseHeaderAggregator::default();
         finalize(&mut agg, false);
         assert!(agg.entries.get(&http::header::CACHE_CONTROL).is_none());
+    }
+
+    // empty string: parse() returns None, acc stays None, entry left unchanged
+    #[test]
+    fn finalize_empty_string_removes_header() {
+        let mut agg = make_aggregator(&[""]);
+        finalize(&mut agg, false);
+        assert_eq!(cc_value(&agg).as_deref(), None);
+    }
+
+    // non-UTF-8: to_str() fails, acc stays None, entry left unchanged
+    #[test]
+    fn finalize_invalid_utf8_removes_header() {
+        let mut agg = ResponseHeaderAggregator::default();
+        let invalid = http::HeaderValue::from_bytes(&[0xFF, 0xFE]).unwrap();
+        agg.write(
+            &http::header::CACHE_CONTROL,
+            &invalid,
+            HeaderAggregationStrategy::Append,
+        );
+        finalize(&mut agg, false);
+        assert_eq!(cc_value(&agg).as_deref(), None);
+    }
+
+    // unrecognized directive: parse() returns Some(default), serializes to no-store
+    #[test]
+    fn finalize_unrecognized_directive_removes_header() {
+        let mut agg = make_aggregator(&["bogus-directive"]);
+        finalize(&mut agg, false);
+        assert_eq!(cc_value(&agg).as_deref(), None);
     }
 
     #[test]
--- a/lib/executor/src/headers/cache_control.rs
+++ b/lib/executor/src/headers/cache_control.rs
@@ -1,0 +1,649 @@
+use crate::headers::{plan::HeaderAggregationStrategy, response::ResponseHeaderAggregator};
+use http::HeaderValue;
+
+#[derive(Clone, Default)]
+struct ParsedCacheControl {
+    no_store: bool,
+    no_cache: bool,
+    must_revalidate: bool,
+    is_private: bool,
+    is_public: bool,
+    max_age: Option<u32>,
+}
+
+fn parse(header: &str) -> Option<ParsedCacheControl> {
+    let trimmed = header.trim();
+    if trimmed.is_empty() {
+        return None;
+    }
+
+    let mut p = ParsedCacheControl::default();
+
+    for part in trimmed.split(',') {
+        let part = part.trim();
+        if part.is_empty() {
+            continue;
+        }
+        let (token, value) = match part.split_once('=') {
+            Some((t, v)) => (t.trim(), Some(v.trim())),
+            None => (part, None),
+        };
+        match token.to_ascii_lowercase().as_str() {
+            "no-store" => p.no_store = true,
+            "no-cache" => p.no_cache = true,
+            "must-revalidate" => p.must_revalidate = true,
+            "private" => p.is_private = true,
+            "public" => p.is_public = true,
+            "max-age" => {
+                if let Some(v) = value {
+                    match v.parse::<u32>() {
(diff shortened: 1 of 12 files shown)


lib/executor/src/headers/cache_control.rs
```diff
@@ -0,0 +1,739 @@
...
+    }
+
+    let mut acc: Option<CacheControl> = None;
+    for v in values {
+        if let Ok(s) = v.to_str() {
+            if let Some(parsed) = parse(s) {
+                merge_into(&mut acc, parsed);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
enisdenjo: it compiles

lib/executor/src/headers/compile.rs
```diff
@@ -85,12 +85,35 @@ impl HeaderRuleCompiler<Vec<ResponseHeaderRule>> for config::ResponseHeaderRule
...
             config::ResponseHeaderRule::Propagate(rule) => {
                 let aggregation_strategy = rule.algorithm.into();
+
                 let spec = materialize_match_spec(
                     &rule.spec,
                     rule.rename.as_ref(),
                     rule.default.as_ref(),
                 )?;
 
+                // named path has no exclude at runtime, so exclude can only
```
dotansimha: nit: this logic could go into some `validate_cache_control(...)` just to make it a bit more scoped

lib/executor/src/headers/cache_control.rs
```diff
@@ -0,0 +1,765 @@
...
+struct CacheControl {
+    no_store: bool,
+    no_cache: bool,
+    must_revalidate: bool,
+    is_private: bool,
+    is_public: bool,
+    max_age: Option<u32>,
+}
+
+fn parse(header: &str) -> Option<CacheControl> {
```
dotansimha, line 20: Any reason we use `&str` here and not a parsed `HeaderValue`? Or, just use `HeaderValue::parse` to do the heavy lifting of parsing?

lib/executor/src/headers/cache_control.rs
```diff
@@ -0,0 +1,765 @@
...
+                    None => {
+                        warn!("cache-control max-age is missing a value");
+                        return None;
+                    }
+                };
+                p.max_age = max_age;
+            }
+            v => {
+                // one invalid part is enough to stop parsing and discard the header
+                warn!("cache-control has unrecognized directive: {v}");
```
dotansimha: should be: 

```
warn!(directive = v, "cache-control has unrecognized directive");
```

lib/executor/src/headers/cache_control.rs
```diff
@@ -0,0 +1,765 @@
...
+    force_no_store: bool,
+    total_responses: usize,
+) {
+    let Some((_, values)) = aggregator.entries.get(&http::header::CACHE_CONTROL) else {
+        // there's no cache-control headers anywhere, so nothing to merge or poison - just leave it absent
+        return;
+    };
+
+    if force_no_store {
+        let value = HeaderValue::from_static("no-store, no-cache, must-revalidate");
```
dotansimha: nit: Since this lives on the hotpath of a request, and is a static value, it can be `lazy_static`.

lib/executor/src/headers/cache_control.rs
```diff
@@ -0,0 +1,765 @@
...
+        // safety: to_header_value only produces ASCII
+        let value = HeaderValue::from_str(&serialized).expect("to_header_value produced non-ASCII");
+        aggregator.entries.insert(
+            http::header::CACHE_CONTROL,
+            (HeaderAggregationStrategy::Last, vec![value]),
+        );
+    } else {
+        // no valid values found, but there were cache-control headers
+        // do the safe thing and graceful thing - completely omit the header
+        warn!("no valid cache-control values found, removing header");
```
dotansimha, line 216: should we also consider a `debug` level log line here, that prints the value that we have?

dotansimha (approved):
Thanks @enisdenjo . I added a few comments/questions and some additional `nit`picking that you may fix or ignore. 

I really like the approach of having a single strict solution for merging these values, and it makes perfect sense in terms of the way it's being handled, in order to prevent leakage/over-permissive cache. 

Thanks, great job! 🏆

lib/executor/src/headers/cache_control.rs
enisdenjo: which `HeaderValue::parse` are we talking about? `HeaderValue` itself only validates that the bytes are visible ascii, it doesn't parse the directives. the actual directive splitting/parsing still has to be done manually regardless. so, if parse would take `&HeaderValue` then the `to_str` would be moved inside anyways haha, no real gain. or am I missing something?

PR #1208, opened 2026-06-28
feat(router): add subscription and websocket metrics

Meemaw:
## What

Adds observability for GraphQL subscriptions and WebSocket clients. There was previously no way to see how many subscriptions/clients are active, how often clients subscribe/unsubscribe, or how often the router drops events because it can't keep up with a subgraph (that was only a `warn!` log).

Metrics are split by direction:

**Client → Router** (the router's server side):
| Metric | Type | Labels |
|---|---|---|
| `hive.router.subscriptions.active` | UpDownCounter | — |
| `hive.router.websocket.connections.active` | UpDownCounter | — |
| `hive.router.subscriptions.operations_total` | Counter | `subscription.operation` (`subscribe`/`unsubscribe`) |

**Router → Subgraph** (the router's client side), all tagged with `subgraph.name`:
| Metric | Type | Labels |
|---|---|---|
| `hive.router.subgraph.subscriptions.active` | UpDownCounter | `subgraph.name` |
| `hive.router.subgraph.subscriptions.operations_total` | Counter | `subgraph.name`, `subscription.operation` |
| `hive.router.subgraph.subscriptions.dropped_messages_total` | Counter | `subgraph.name` |

## How

- **Client-side** metrics use RAII guards at the two consumer sites — the WS `Subscribe` handler (`websocket_server.rs`) and the HTTP streaming response (`shared_state.rs::into_response`, SSE/multipart) — so the active gauge decrements and the `unsubscribe` counter fires on disconnect/stream-end. Covers all client transports.
- **Upstream** metrics are tracked at the single `SubgraphExecutorMap::subscribe` chokepoint (`map.rs`), wrapping the returned stream so one guard covers every subgraph subscription transport (WS, multipart/SSE, HTTP callback).
- **Dropped messages** replaces the per-drop `warn!` in `subscription_buffer.rs::drain_into` (now demoted to `trace!`) with the subgraph-tagged counter.
- All metrics follow the existing OTel catalog pattern (`METRIC_SPECS` + debug-assert on attributes).

## Notes

- With subscription dedup active, the **upstream** active count reflects deduped subgraph subscriptions (one per shared upstream) while the **client** active count reflects every client subscription including joiners — the correct semantic for each side, but the two gauges won't match 1:1.
- Metrics are no-ops when telemetry is disabled (instruments are `None` internally).

### DIFF
--- a/bin/router/src/pipeline/mod.rs
+++ b/bin/router/src/pipeline/mod.rs
@@ -430,7 +430,7 @@ pub async fn graphql_request_handler(
             },
         );
 
-        shared_response.into_response(response_mode)
+        shared_response.into_response(response_mode, &shared_state.telemetry_context.metrics)
     }
     .instrument(span_clone)
     .await
--- a/bin/router/src/pipeline/websocket_server.rs
+++ b/bin/router/src/pipeline/websocket_server.rs
@@ -98,6 +98,10 @@ async fn ws_service(
     request_context: SharedRequestContext,
 ) -> Result<impl Service<ws::Frame, Response = Option<ws::Message>, Error = io::Error>, web::Error>
 {
+    // Tracks the active WebSocket connection in metrics; only set for accepted connections so
+    // rejected handshakes are not counted. Moved into `on_shutdown` so it is dropped (and the
+    // gauge decremented) when the connection closes.
+    let mut connection_metrics_guard = None;
     if !has_accepted_subprotocol {
         debug!("WebSocket connection rejecting due to unacceptable subprotocol");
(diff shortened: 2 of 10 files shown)


enisdenjo (commented):
An active connection can also be from HTTP streaming, not only from WebSockets. I think we should count these in as well.

What about using labels to disambiguate active connections over the supported transports? Same for subgraphs (active connections labeled by transport router<->subgraph). Same for subscriptions (would that be necessary)? 

Some E2E tests would be great too.

PR #1217, opened 2026-06-30
fix(query-planner): prevent `@requires` creating a circular dependency across subgraphs

kamilkisiela:
The path finder could enter an infinite loop when:
1. A field with `@requires` needs to entity-move to another subgraph
2. The entity-move edge's requirements include the field being resolved
3. This creates a circular dependency: "to get field X, use edge requiring X"

The fix detects this cycle by skipping edges whose requirements include the
target field being searched, allowing alternative entity-move paths to be used.

### DIFF
--- a/lib/query-planner/src/planner/walker/pathfinder.rs
+++ b/lib/query-planner/src/planner/walker/pathfinder.rs
@@ -22,6 +22,7 @@ use crate::{
         tree::query_tree_node::QueryTreeNode,
         walker::best_path::{find_best_paths, BestPathTracker},
     },
+    state::supergraph_state::SupergraphState,
 };
 
 use super::{error::WalkOperationError, excluded::ExcludedFromLookup, path::OperationPath};
@@ -100,6 +101,7 @@ impl<'op> From<&'op NavigationTarget<'op>> for NavigationTargetKey<'op> {
 
 struct PathSearch<'graph> {
     graph: &'graph Graph,
+    supergraph: &'graph SupergraphState,
     override_context: &'graph PlannerOverrideContext,
     cancellation_token: &'graph CancellationToken,
     /// Edges currently being checked in this path search.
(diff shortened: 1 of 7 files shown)


lib/query-planner/src/tests/issues.rs
```diff
@@ -101,35 +104,36 @@ fn issue_281_test() -> Result<(), Box<dyn Error>> {
...
-              ... on Product {
-                __typename
-                pid
-              }
-            } =>
-            {
-              ... on Product {
-                c
-                pid
+        Parallel {
```
kamilkisiela, line 107: it even improves the query plan in this scenario.

Instead of Sequence(1,2) we do now Parallel(1,2)

lib/query-planner/src/planner/walker/pathfinder.rs
```diff
@@ -213,6 +213,28 @@ impl<'graph> PathSearch<'graph> {
                     continue;
                 }
 
+                // If we're searching for a field and this edge's
```
dotansimha, line 263: good catch

lib/query-planner/src/planner/walker/pathfinder.rs
```diff
@@ -213,6 +213,28 @@ impl<'graph> PathSearch<'graph> {
...
+                        if requirements
+                            .selection_set
+                            .items
+                            .iter()
+                            .any(|item| matches!(item, SelectionItem::Field(f) if &f.name == target_field_name))
+                        {
+                            continue;
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
dotansimha: @kamilkisiela you broke your own guideline 😆

lib/query-planner/src/tests/issues.rs
dotansimha: nice

dotansimha (approved):

dotansimha (approved):

Issue #1194, opened 2026-06-24
Subgraph error masking with ability to turn on/off


dotansimha:
At the moment, our Router exposes the subgraph errors to the consumer. Ideally, we should mask these, but allow developers to enable and expose those as-is (also probably can be enabled via this https://github.com/graphql-hive/router/issues/1087 in the future).

dotansimha:
Available in `0.0.85` 🎉 

Product update: https://the-guild.dev/graphql/hive/product-updates/2026-07-26-hive-router-subgraph-error-masking
Docs:  https://the-guild.dev/graphql/hive/docs/router/security/error-masking

PR #1259, opened 2026-07-07
feat(router): subgraph error masking

dotansimha:
Related https://github.com/graphql-hive/router/issues/1194 
Docs PR:  https://github.com/graphql-hive/docs/pull/143

## Background 

This PR introduces error masking for subgraph errors. Subgraphs are considered internal, and not front-facing, so developers can safely assume that a subgraph service does not need to perform masking on that layer, and trust the router to perform masking for the error produced by the subgraph. 

By default, error masking is enabled, in order to prevent any leakage of internal/sensitive details from the errors. 

## How subgraph errors are detected? 

Any error that has `extensions.service` field, either constructed by the upstream service, or if it was syntethically created by the Router (in case of subgraph network errors).  

## When do we perform error masking? 

Error masking is performed last, after all metrics/tracing/logging took place. It's only impacting the final, user-facing response. 

## Configuration

* Developers can customize the error message used (default: `Unexpected error`) 
* Can enable/disable masking via config, or via env var (`DISABLE_SUBGRAPH_ERROR_MASKING=true`)
* Can mask `error.extesions` via deny/allow list 
* Can configure either `all` or per-subgraph configuration (overrides `all`)

Example: 

```yaml
              error_masking:
                all:
                  enabled: true
                  extensions:
                    mode: allow
                    keys: [] # allow none!
                subgraphs:
                  products:
                    enabled: false
                    extensions:
                      mode: deny
                      keys: [] # dont deny anything!
``` 

## More stuff in this PR 

* Since now error masking is enabled by default, some e2e needed to configure it as disabled, in order to pass the snapshot testing 
* I fixed a small issue in the way inline config in e2e tests are loaded, becase it was not using env overrides at all 

## TODO 

- [x] initial impl 
- [x] initial e2e
- [x] config is now "all or nothing" - we need to make each part of `all` and `subgraphs` and `Option` so user can override only part of the config (`all: { error_message: false, extension: { ... } }` + `subgraphs: { named: { error_message: true }}` should use the `error_message: true` AND the `exntesions` from `all`. 
- [x] some sanity testing
- [x] more e2es? 
- [x] does the detection of subgraph-specific errors is correct? should we have a branded error to detect it? 
- [x] rebase after https://github.com/graphql-hive/router/pull/1233 and confirm that it works correctly with the new errors added there
- [x] env var to disable masking  
- [x] docs
- [x] changeset

### DIFF
--- a/.changeset/subgraph_error_masking.md
+++ b/.changeset/subgraph_error_masking.md
@@ -24,7 +24,7 @@ error_masking:
         - code
   subgraphs:
     products:
-      error_message: false
+      enabled: false
 ```
 
 - `error_message` toggles message redaction; `extensions` redacts extension keys via an `allow`/`deny` list.
--- a/.changeset/subgraph_error_masking.md
+++ b/.changeset/subgraph_error_masking.md
@@ -1,0 +1,36 @@
+---
+hive-router-config: patch
+hive-router: patch
+hive-router-plan-executor: patch
+hive-router-internal: patch
+---
+
+# Subgraph Error Masking
+
+Mask subgraph errors before they reach clients, preventing internal details from leaking.
+
+Masking is **enabled by default**: subgraph error messages are replaced with `"Unexpected error"`. It runs last in the pipeline, so metrics, tracing, and logging still see the original error.
+
+Configure it under `error_masking`:
+
+```yaml
+error_masking:
+  redacted_error_message: "Unexpected error"
+  all:
+    error_message: true
+    extensions:
+      mode: allow # allow | deny
+      keys:
+        - code
+  subgraphs:
+    products:
+      error_message: false
+```
+
+- `error_message` toggles message redaction; `extensions` redacts extension keys via an `allow`/`deny` list.
+- `subgraphs.<name>` overrides `all` per subgraph, inheriting any field it doesn't set.
+- Set `DISABLE_SUBGRAPH_ERROR_MASKING=true` to disable message masking without editing the config.
+
+[Documentation](http://the-guild.dev/graphql/hive/docs/router/security/error-masking)
+
+Fixes https://github.com/graphql-hive/router/issues/1194
--- a/.changeset/subgraph_error_masking.md
+++ b/.changeset/subgraph_error_masking.md
@@ -15,9 +15,10 @@ Configure it under `error_masking`:
 
 ```yaml
 error_masking:
+  enabled: true # enabled by default 
   redacted_error_message: "Unexpected error"
-  all:
-    error_message: true
+  all: # default config, unless you override it with subgraph-specific 
+    enabled: true
     extensions:
       mode: allow # allow | deny
       keys:
--- a/lib/executor/src/execution/error_masking.rs
+++ b/lib/executor/src/execution/error_masking.rs
@@ -1,29 +1,40 @@
 use std::collections::{HashMap, HashSet};
 
 use crate::response::graphql_error::{GraphQLError, GraphQLErrorExtensions};
-use hive_router_config::error_masking::{
-    ErrorMaskingConfig, ExtensionsMaskingConfig, SubgraphErrorMaskingConfig,
-};
+use hive_router_config::error_masking::{ErrorMaskingConfig, ExtensionsMaskingConfig};
 
 pub struct ErrorMaskingRuntime {
-    default_redacted_error_message: String,
-    per_subgraph_config: HashMap<String, ErrorMaskingCompiledConfig>,
-    default_config: ErrorMaskingCompiledConfig,
+    redacted_error_message: String,
+
+    // Error masking for the error message, split by "all" and subgraph-specific
+    default_error_masking: bool,
+    per_subgraph_error_masking: HashMap<String, bool>,
+
+    // Error masking for the error extensions, split by "all" and subgraph-specific
+    default_extensions_masking: Option<RedactExtensionsPlan>,
+    per_subgraph_extensions_masking: HashMap<String, Option<RedactExtensionsPlan>>,
 }
 
 impl ErrorMaskingRuntime {
     pub fn apply(&self, err: &mut GraphQLError) {
         if let Some(service_name) = &err.extensions.service_name {
-            let effective_config = self
-                .per_subgraph_config
+            let should_mask_message = self
+                .per_subgraph_error_masking
                 .get(service_name)
-                .unwrap_or(&self.default_config);
+                .copied()
+                .unwrap_or(self.default_error_masking);
 
-            if effective_config.redact_error_message {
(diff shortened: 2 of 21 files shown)


lib/executor/src/execution/error_masking.rs
```diff
@@ -0,0 +1,145 @@
...
+            "affectedPath" => {
+                extensions.affected_path = None;
+                return;
+            }
+            _ => {
+                extensions.extensions.remove(key);
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
dotansimha: no longer correct

lib/executor/src/execution/error_masking.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn compile_extensions_config(
+        extensions_config: &ExtensionsMaskingConfig,
+    ) -> RedactExtensionsPlan {
+        match extensions_config {
+            ExtensionsMaskingConfig::AllowList { keys } => {
+                RedactExtensionsPlan::Allow(keys.clone())
+            }
+            ExtensionsMaskingConfig::DenyList { keys } => RedactExtensionsPlan::Deny(keys.clone()),
+        }
+    }
```
enisdenjo: maybe?

```rs
impl From<&ExtensionsMaskingConfig> for RedactExtensionsPlan {
    fn from(config: &ExtensionsMaskingConfig) -> Self {
        match config {
            ExtensionsMaskingConfig::AllowList { keys } => Self::Allow(keys.clone()),
            ExtensionsMaskingConfig::DenyList { keys } => Self::Deny(keys.clone()),
        }
    }
}
```

lib/router-config/src/error_masking.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl Default for SubgraphErrorMaskingConfig {
+    fn default() -> Self {
+        Self {
+            error_message: default_subgraph_redact_error_message(),
+            extensions: None,
+        }
+    }
+}
```
enisdenjo: None's already the default of Option

```suggestion
    /// Whether to redact the `error_message` in errors, for that specific subgraph.
    ///
    /// Configuring this will override the global `all.error_message` setting.
    #[serde(default)]
    pub error_message: Option<bool>,
    /// Whether to redact the `extensions` in errors, for that specific subgraph.
    /// Configuring this will override the global `all.extensions` setting.
    ///
    /// You may pick the execution mode by setting `mode: allow` or `mode: deny`.
    /// Note: only root-level fields are supported.
    #[serde(default)]
    pub extensions: Option<ExtensionsMaskingConfig>,
}
```

enisdenjo (changes requested):
I don't perticularily like the `error_masking.*.error_message` naming.

reading this is weird:

```yaml
error_masking:
  all:
    error_message: false
```

and it does not tell me immediately what `error_message: false` does. without reading the docs, I would think it disables the error message (masking enabled) interpretaing "error message false" as "no error message"

why not simply use `enabled`?

```yaml
error_masking:
  all:
    enabled: true / false
```

---

`DISABLE_SUBGRAPH_ERROR_MASKING=true` only changes `all.error_message` but not `subgraphs.<name>.error_message: true` - is this intentional? should it be? I think that overriding with an env var should be applied globally

---

(cant leave comment on unchanged lines)

subscription failures bypass error masking. `error_masking_runtime` is used by `execute_query_plan_with_data`, but both the `stream_result` error branch at lines 347-359 and the fatal branch at lines 402-409 in lib/executor/src/execution/plan.rs serialize `err.into()` directly through `FailedExecutionResult`

additionally, there's no e2e tests covergin subscriptions error masking

lib/executor/src/execution/error_masking.rs
dotansimha: yes, fixed

dotansimha:
> why not simply use enabled?

Renamed `<>.error_message` to be `<>.enabled` 

> DISABLE_SUBGRAPH_ERROR_MASKING=true only changes all.error_message but not subgraphs.<name>.error_message: true - is this intentional? should it be? I think that overriding with an env var should be applied globally


Good point. we need a higher level `enabled` flag for the feature, in order to make it easy to configure whole, and then `DISABLE_SUBGRAPH_ERROR_MASKING` toggles that. Added.

dotansimha:
> subscription failures bypass error masking. error_masking_runtime is used by execute_query_plan_with_data, but both the stream_result error branch at lines 347-359 and the fatal branch at lines 402-409 in lib/executor/src/execution/plan.rs serialize err.into() directly through FailedExecutionResult

Yeah, missed those. Thanks for flagging. Push a change to fully mask those as well. 

> additionally, there's no e2e tests covergin subscriptions error masking

Done, added 3 new tests: 2 for the failures i missed earlier and you flagged, and once for the regular subgraph-originated errors.

.changeset/subgraph_error_masking.md
```diff
@@ -0,0 +1,36 @@
...
+
+Masking is **enabled by default**: subgraph error messages are replaced with `"Unexpected error"`. It runs last in the pipeline, so metrics, tracing, and logging still see the original error.
+
+Configure it under `error_masking`:
+
+```yaml
+error_masking:
+  redacted_error_message: "Unexpected error"
+  all:
+    error_message: true
```
enisdenjo:
```suggestion
  all:
    enabled: true
```

lib/router-config/src/env_overrides.rs
```diff
@@ -192,6 +196,16 @@ impl EnvVarOverrides {
...
         }
 
+        if let Some(disable_subgraph_error_masking) = self.disable_subgraph_error_masking.take() {
+            debug!(
+                "[config-override] 'disable_subgraph_error_masking' = {}",
+                disable_subgraph_error_masking
+            );
+
+            config =
+                config.set_override("error_masking.enabled", !disable_subgraph_error_masking)?;
```
enisdenjo, line 206: we're still overriding only the root/all and not `subgraphs.*.enabled`?

enisdenjo (approved):
some nits, otherwise lgtm

lib/router-config/src/env_overrides.rs
dotansimha: No, we have now: `error_masking.enabled` (default: true) which controls the whole feature. And then `error_masking.all.enabled` controls the default masking for the error message. 

The env var controls  `error_masking.enabled` (the whole feature).

.changeset/subgraph_error_masking.md
dotansimha: thanks!

lib/router-config/src/env_overrides.rs
dotansimha: So basically we allow users to control the whole feature by using the env var, instead of specific configs

PR #1261, opened 2026-07-07
feat(router): subscriptions metrics

### LINKED ISSUE #921: Usage reporting and metrics of subscriptions
Improvement to #620.

enisdenjo:
Somewhat inspired by and therefore closes #1208, closes #921

Full observability into GraphQL subscription activity - both between clients and the router, and between the router and subgraphs - via OTEL metrics, so one can answer "how many subscriptions are live right now", "what is the subscription churn rate", and "are we dropping or losing subscription events" per transport and per subgraph.

### Live-state gauges

| Metric                                            | Labels                                    | Unit             | Meaning                                                     |
| ------------------------------------------------- | ----------------------------------------- | ---------------- | ----------------------------------------------------------- |
| `hive.router.subscriptions.clients.active`        | `subscription.transport`                  | `{subscription}` | Active subscription operations from clients to the router   |
| `hive.router.subscriptions.clients.connections`   | `subscription.transport`                  | `{connection}`   | Active transport connections carrying client subscriptions  |
| `hive.router.subscriptions.subgraphs.active`      | `subgraph.name`                           | `{subscription}` | Active subscription operations from the router to subgraphs |
| `hive.router.subscriptions.subgraphs.connections` | `subgraph.name`, `subscription.transport` | `{connection}`   | Active transport connections from the router to subgraphs   |

Operations and connections are separate because one connection can (will be able) multiplex multiple operations (WebSocket) or subscription deduplication can fan one subgraph operation out to many clients.

### Lifecycle counters

| Metric                                              | Labels                                              | Unit             | Meaning                                                      |
| --------------------------------------------------- | --------------------------------------------------- | ---------------- | ------------------------------------------------------------ |
| `hive.router.subscriptions.clients.started_total`   | `subscription.transport`                            | `{subscription}` | Client subscriptions started (churn/rate, survives restarts) |
| `hive.router.subscriptions.clients.ended_total`     | `subscription.transport`, `subscription.end_reason` | `{subscription}` | Client subscriptions ended                                   |
| `hive.router.subscriptions.subgraphs.started_total` | `subgraph.name`                                     | `{subscription}` | Subgraph subscriptions started                               |
| `hive.router.subscriptions.subgraphs.ended_total`   | `subgraph.name`                                     | `{subscription}` | Subgraph subscriptions ended                                 |

Every start is matched by exactly one end, regardless of how the subscription ends (explicit unsubscribe, connection drop, stream completion, error).

`hive.router.subscriptions.clients.ended_total` additionally carries `subscription.end_reason`, one of:

- `completed` - the upstream/broadcast source finished normally (subscription ran to completion).
- `error` - the subscription ended because an error event was delivered to the client.
- `client_disconnected` - the client closed the connection, unsubscribed, or the stream was otherwise dropped from the router's side. This is also the default recorded if a guard drops without an explicit reason set, so no end is ever left unattributed.

### Saturation/error counters

| Metric                                                       | Labels                   | Unit        | Meaning                                                                                                                      |
| ------------------------------------------------------------ | ------------------------ | ----------- | ---------------------------------------------------------------------------------------------------------------------------- |
| `hive.router.subscriptions.clients.sent_messages_total`      | `subscription.transport` | `{message}` | Messages successfully sent to a client subscriber                                                                            |
| `hive.router.subscriptions.clients.lagged_messages_total`    | `subscription.transport` | `{message}` | Messages skipped for slow clients on the broadcast fan-out (incremented by the actual number skipped); subscription survives |
| `hive.router.subscriptions.subgraphs.dropped_messages_total` | `subscription.transport` | `{message}` | Subgraph messages dropped because the internal subscription buffer was full (consumer too slow); subscription survives       |

`sent_messages_total` and `lagged_messages_total` together give the % of messages a client actually received vs skipped (`sent / (sent + lagged)`), per transport.

No subscription is ever terminated because of a slow consumer. All buffer-full situations (subgraph WebSocket, HTTP multipart/SSE, HTTP callback) now drop the single message, record it, and keep the subscription alive - mirroring the broadcast lag semantics on the client side.

This includes a behavior change on the HTTP callback endpoint: a full buffer no longer terminates the subscription and returns 503; it responds OK and drops the message.

### Label semantics

`subscription.transport` values: `websocket`, `http_multipart`, `http_sse`, `http_callback` - covering all client-facing transports (WebSocket, incremental delivery / Apollo multipart HTTP, SSE) and all subgraph executors (WebSocket, HTTP multipart/SSE, HTTP callback).

`subscription.end_reason` values (client-side `ended_total` only): `completed`, `error`, `client_disconnected`.

### Why custom units?

The convention for what goes in units comes from the OTEL metrics spec, which says units should be UCUM (Unified Code for Units of Measure) case-sensitive codes.

Two kinds of values:

1. Real physical units - UCUM codes like s, ms, By (bytes), %. E.g. `http.server.request.duration` is defined with unit s in the semantic conventions.
2. Counts of things - UCUM has "annotations": a label in curly braces, like `{subscription}`, `{connection}`, `{request}`, `{message}`.
   Semantically it means "dimensionless count of X"; the braces are UCUM syntax for "this is a human hint, not a real unit". OTEL semantic conventions use this everywhere, e.g. `http.server.active_requests` has unit `{request}`, JVM thread count `{thread}`. Also per convention, when the unit is an annotation like this, the metric name should NOT repeat it (no `.subscriptions_count` suffix) - the unit field carries it

What it actually buys us:

- Exporters use it. The Prometheus exporter appends real units to the metric name (`_seconds`, `_bytes`) and uses it for the `# UNIT/# HELP` metadata; brace annotations are stripped from the name but still land in metadata.
- Backends/UIs use it. Grafana, Datadog etc. pick axis formatting and legends from it.
- It's documentation at the instrument - `{connection}` on `clients.connections` tells the reader the gauge counts connections, not bytes or operations.

### Why started and ended counters?

Why not derive ends from the `active` gauge (`ends = starts - active`)? Because the derivation is only correct between restarts. Backends (like Prometheus and alike) have special handling for counters: when a counter drops (100 -> 0), `rate()`/`increase()` recognize it as a reset and keep computing correctly across it. Gauges have no reset semantics - when the router restarts, `active` drops hard to 0 and any expression involving `delta(active)` sees that as a mass unsubscribe - producing nonsense. A real `ended_total` counter is correct across restarts.

What the start/end rates answer that `active` alone can't:

- Mass disconnects. End-rate spike with flat start rate = something killed a batch of subscriptions (deploy, LB idle timeout, subgraph WS outage). `active` alone shows only a dip you can't attribute.
- Flapping/churn. Start and end rates both high while `active` is flat = clients constantly reconnecting (broken keepalive, client retry loop). Invisible in the gauge.
- Average subscription lifetime = `active / rate(ended_total)`. A median lifetime of 3 seconds means a client is probably subscribing per-request instead of holding the stream.
- Invariant checking. `started_total - ended_total` should equal `active`; drift between them indicates a leaked lifecycle guard.

### Removed `CallbackError::ClientTooSlow`?

Slow consumer should not hold up or drop the subscription, messages should simply be dropped. This is in line with all other streaming executors.

### WebSocket server records GraphQL errors

Previously, GraphQL errors (like jwt, validation, parsing, etc.) were not recorded in the WebSocket server. [Now they are.](https://github.com/graphql-hive/router/pull/1261/commits/566e11cbcfd0d0ed98989050448cf5671698850d)

### TODO

- [x] docs https://github.com/graphql-hive/docs/pull/141
- [x] changeset
  - dropped client too slow http callback stop error, instead we drop the message like with other transports
  - relaxed logs relating to lagging/dropping messages from warn to debug
  - subscriptions metrics

### DIFF
--- a/bin/router/src/pipeline/websocket_server.rs
+++ b/bin/router/src/pipeline/websocket_server.rs
@@ -17,6 +17,7 @@ use std::time::Instant;
 use tokio::sync::mpsc;
 use tracing::{debug, error, trace, warn, Instrument};
 
+use hive_router_internal::telemetry::metrics::catalog::values::SubscriptionEndReason;
 use hive_router_internal::telemetry::metrics::subscription_metrics::{
     ActiveClientConnectionGuard, SubscriptionTransport,
 };
@@ -602,7 +603,7 @@ async fn handle_text_frame(
                         // making cancellation impossible
                         rt::spawn(async move {
                             let _guard = guard;
-                            let _client_op_guard = client_op_guard;
+                            let mut client_op_guard = client_op_guard;
                             let mut cancelled = false;
 
                             loop {
@@ -614,6 +615,7 @@ async fn handle_text_frame(
                                                 metrics.subscriptions.record_client_sent(SubscriptionTransport::WebSocket);
                                             }
                                             Ok(SubscriptionEvent::Error(errors)) => {
+                                                client_op_guard.set_end_reason(SubscriptionEndReason::Error);
                                                 let _ = sink.send(ServerMessage::error(&id_for_loop, &errors)).await;
                                                 break;
                                             }
@@ -626,6 +628,7 @@ async fn handle_text_frame(
                                                 continue;
                                             }
                                             Err(tokio::sync::broadcast::error::RecvError::Closed) => {
+                                                client_op_guard.set_end_reason(SubscriptionEndReason::Completed);
                                                 break;
                                             }
                                         }
--- a/bin/router/src/pipeline/websocket_server.rs
+++ b/bin/router/src/pipeline/websocket_server.rs
@@ -17,6 +17,9 @@ use std::time::Instant;
 use tokio::sync::mpsc;
 use tracing::{debug, error, trace, warn, Instrument};
 
+use hive_router_internal::telemetry::metrics::subscription_metrics::{
+    ActiveClientConnectionGuard, SubscriptionTransport,
+};
 use hive_router_internal::telemetry::traces::spans::graphql::GraphQLOperationSpan;
 use hive_router_plan_executor::executors::graphql_transport_ws::{
     ClientMessage, CloseCode, ConnectionInitPayload, ServerMessage, WS_SUBPROTOCOL,
@@ -108,6 +111,18 @@ async fn ws_service(
         debug!("WebSocket connection accepted");
     }
 
+    let conn_guard: Option<ActiveClientConnectionGuard> = if has_accepted_subprotocol {
+        Some(
+            shared_state
+                .telemetry_context
+                .metrics
+                .subscriptions
+                .active_client_connection(SubscriptionTransport::WebSocket),
+        )
+    } else {
+        None
+    };
+
     let ws_uri: Rc<http::Uri> = Rc::new(
         shared_state
             .router_config
@@ -175,6 +190,8 @@ async fn ws_service(
         // in turn cancel all active subscription streams and perform
         // the cleanup in there
         state.borrow_mut().subscriptions.clear();
+        // drop conn_guard here to decrement the connection counter
+        drop(conn_guard);
     });
 
     Ok(chain(service).and_then(on_shutdown))
@@ -573,6 +590,8 @@ async fn handle_text_frame(
                             .receiver
                             .unwrap_or_else(|| response.body.subscribe());
 
+                        let client_op_guard = shared_state.telemetry_context.metrics.subscriptions.active_client_operation(SubscriptionTransport::WebSocket);
+
                         trace!(id = %id, "Subscription started");
 
                         let sink = sink.clone();
@@ -582,6 +601,7 @@ async fn handle_text_frame(
                         // making cancellation impossible
                         rt::spawn(async move {
                             let _guard = guard;
+                            let _client_op_guard = client_op_guard;
                             let mut cancelled = false;
 
                             loop {
--- a/bin/router/src/pipeline/websocket_server.rs
+++ b/bin/router/src/pipeline/websocket_server.rs
@@ -617,7 +617,10 @@ async fn handle_text_frame(
                                                 break;
                                             }
                                             Err(tokio::sync::broadcast::error::RecvError::Lagged(n)) => {
-                                                warn!(id = %id_for_loop, lagged = n, "Broadcast receiver lagged, dropping message");
+                                                // NOTE: not warn to avoid log spam when receiver starts
+                                                // lagging. users should rely on the lagged_messages
+                                                // metric to detect slow consumers and tune accordingly
+                                                debug!(id = %id_for_loop, lagged = n, "Broadcast receiver lagged, dropping message");
                                                 metrics.subscriptions.record_client_lag(SubscriptionTransport::WebSocket, n);
                                                 continue;
                                             }
--- a/bin/router/src/pipeline/websocket_server.rs
+++ b/bin/router/src/pipeline/websocket_server.rs
@@ -591,6 +591,7 @@ async fn handle_text_frame(
                             .unwrap_or_else(|| response.body.subscribe());
 
                         let client_op_guard = shared_state.telemetry_context.metrics.subscriptions.active_client_operation(SubscriptionTransport::WebSocket);
+                        let metrics = shared_state.telemetry_context.metrics.clone();
 
                         trace!(id = %id, "Subscription started");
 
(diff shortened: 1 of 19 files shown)


bin/router/src/shared_state.rs
```diff
@@ -183,6 +204,7 @@ impl SharedRouterStreamResponse {
                     }
                     Err(tokio::sync::broadcast::error::RecvError::Lagged(n)) => {
                         warn!(lagged = n, "Broadcast receiver lagged, dropping message");
+                        metrics.subscriptions.record_client_lag(transport, n);
```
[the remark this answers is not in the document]
enisdenjo: unnecessary, rust autocast

bin/router/src/pipeline/websocket_server.rs
```diff
@@ -597,6 +618,7 @@ async fn handle_text_frame(
                                             }
                                             Err(tokio::sync::broadcast::error::RecvError::Lagged(n)) => {
                                                 warn!(id = %id_for_loop, lagged = n, "Broadcast receiver lagged, dropping message");
+                                                metrics.subscriptions.record_client_lag(SubscriptionTransport::WebSocket, n);
```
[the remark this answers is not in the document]
enisdenjo: unnecessary, rust autocast

lib/internal/src/telemetry/metrics/subscription_metrics.rs
```diff
@@ -0,0 +1,314 @@
...
+pub struct SubscriptionMetrics {
+    subgraphs_active: Option<UpDownCounter<i64>>,
+    subgraphs_connections: Option<UpDownCounter<i64>>,
+    subgraphs_started_total: Option<Counter<u64>>,
+    subgraphs_ended_total: Option<Counter<u64>>,
+    subgraphs_dropped_messages_total: Option<Counter<u64>>,
+    clients_active: Option<UpDownCounter<i64>>,
+    clients_connections: Option<UpDownCounter<i64>>,
+    clients_started_total: Option<Counter<u64>>,
+    clients_ended_total: Option<Counter<u64>>,
```
kamilkisiela, line 38: For `_ended_total`, do you think we could track the reason why the subscriptions ended?

lib/internal/src/telemetry/metrics/subscription_metrics.rs
```diff
@@ -0,0 +1,314 @@
...
+    clients_lagged_messages_total: Option<Counter<u64>>,
```
kamilkisiela, line 39: Do we reject subscriptions? If so, maybe a metric for rejections (with a reason) would be good to have.

lib/internal/src/telemetry/metrics/subscription_metrics.rs
```diff
@@ -0,0 +1,314 @@
...
+        self.into()
+    }
+}
+
+pub struct SubscriptionMetrics {
+    subgraphs_active: Option<UpDownCounter<i64>>,
+    subgraphs_connections: Option<UpDownCounter<i64>>,
+    subgraphs_started_total: Option<Counter<u64>>,
+    subgraphs_ended_total: Option<Counter<u64>>,
+    subgraphs_dropped_messages_total: Option<Counter<u64>>,
```
kamilkisiela, line 34: We have a counter for `lagged` and `dropped` messages, but no way to measure sent messages. Without this, users won't able to track the % of total messages.

enisdenjo: https://github.com/graphql-hive/router/pull/1261/commits/af627ed1bb1941f87029ffd01266a97831230b1b

lib/internal/src/telemetry/metrics/subscription_metrics.rs
enisdenjo: https://github.com/graphql-hive/router/pull/1261/commits/5081175174f8abf882e6d2fc6e9cfcaaf59f02e8

lib/internal/src/telemetry/metrics/subscription_metrics.rs
enisdenjo: rejections are covered by `metrics.graphql.record_error` actually, until the subscription starts - it's a regular graphql request. but I did [add the error recording to the websocket server](https://github.com/graphql-hive/router/pull/1261/commits/566e11cbcfd0d0ed98989050448cf5671698850d), which was missing before.

the remaining things that are not covered on the ws side are - protocol-level closes (bad subprotocol, init timeout, duplicate id, unparseable frame). those are client bugs (I'd say noise), they're even `debug!`-logged, and nobody will alert on them because the clients are acting up.

kamilkisiela:
Let's add a changeset and merge

kamilkisiela (approved):
but don't forget about a changeset

PR #1272, opened 2026-07-12
chore(deps): downgrade `serde_with` in `graphql_tools` 

TuvalSimha:
This pull request downgrades `serde_with` to ensure Fastly SDK compatibility (requires time = "=0.3.23") for graphql-tools migration.


Cargo.toml
```diff
@@ -37,7 +37,7 @@ serde_json = "1.0.150"
 sonic-rs = "0.5.3"
 insta = { version = "1.42.1", features= ["filters"] }
 criterion = { version = "0.8", features = ["html_reports", "async_tokio"] }
-lazy_static = "1.5.0"
+lazy_static = "1"
```
dotansimha, line 40: We prefer not to downgrade here, if you need an explicit version in graphql-tools, then please specify it there, instead of using `{ workspace = true }`

Cargo.toml
```diff
@@ -46,7 +46,7 @@ http = "1.4.1"
 http-serde = "2.1.1"
 http-body-util = "0.1.3"
 hyper = { version = "1.10.1" }
-thiserror = "2.0.14"
+thiserror = "1"
```
dotansimha, line 49: We prefer not to downgrade here, if you need an explicit version in graphql-tools, then please specify it there, instead of using `{ workspace = true }`

dotansimha: Also, why downgrade this one as well? your PR mentions serde_with

lib/graphql-tools/Cargo.toml
```diff
@@ -17,7 +17,7 @@ harness = false
 lazy_static = { workspace = true }
 serde = { version = "1.0.200", features = ["derive"] }
 serde_json = "1.0"
-serde_with = "3.21.0"
+serde_with = "2.3"
```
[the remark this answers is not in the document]
dotansimha: can you please share an overview of what do we lose when doing that? 
I mean, we prefer using the latest obviously, but we do want to keep things compatible.

TuvalSimha:
@dotansimha 
I'm downgrading a few package versions because I want to create a PR for the parser migration that is only related to the parser itself. I got feedback that the original PR, which also included the Rust, Fastly, and Viceroy changes, was too large.

PR #1279, opened 2026-07-14
feat(router): `OnGraphqlAnalysis` hook and shared `OperationFilter` for authorization and plugins

### LINKED ISSUE #1086: Router-Level Auth Field Stripping for Federated GraphQL
# Router-Level Auth Field Stripping for Federated GraphQL

## Problem

We use schema directives (`@authorize`) to mark fields that require authentication. When an unauthenticated client sends a query containing a mix of public and auth-required fields, the router currently forwards the **entire** query to subgraphs — including the auth-required fields the client cannot access.

```graphql
# Client sends (no JWT):
{
  streakBoost { currentDayIndex }   # requires @authorize
  trendingCollections { name }      # public
}
```

Today, both fields reach the subgraph. The subgraph then checks auth, returns `null` + an auth error for `streakBoost`, and returns real data for `trendingCollections`. This works correctly, but the subgraph still receives and processes a field it will always reject — wasting compute.

**What we want:** The router strips auth-required fields from the query before it reaches any subgraph when the request has no valid JWT. The subgraph should only receive:

```graphql
{
  trendingCollections { name }
}
```

The router itself would return `null` + an auth error extension for `streakBoost`, matching the response shape the subgraph would have produced.

For the pure case where **all** root fields require auth and there's no JWT, we've already implemented early rejection at the router (the query never reaches the subgraph at all). Field stripping is the natural next step for the mixed case.

## Why This Matters

- **Reduced subgraph load**: Auth-required fields in unauthenticated requests are guaranteed to return null. Forwarding them wastes subgraph CPU, memory, and downstream calls (DB queries, gRPC fanout, etc.).
- **Latency**: The subgraph may do expensive work (resolve nested fields, hit caches) before the auth check rejects the field. Stripping at the router avoids this entirely.
- **Security surface**: Fewer unnecessary requests reaching subgraphs means less attack surface for unauthenticated traffic.

## What We've Built So Far

We have a router plugin (`auth.rs`) that:

1. Extracts and verifies JWTs in `on_graphql_params` / `on_execute`
2. Reads `@authorize` directives from the supergraph schema at startup (`schema_directives.rs`)
3. Walks the operation's `SelectionSet` to determine which root fields require auth
4. Rejects the entire query in `on_execute` when **all** root fields require auth and there's no JWT

The logic to identify which fields require auth already exists. What's missing is the ability to **remove** those fields from the operation before the query planner runs.

## Current Plugin API Constraint

The router plugin lifecycle is:

```
on_graphql_params  →  on_graphql_parse  →  on_graphql_validation  →  on_query_plan  →  on_execute  →  on_subgraph_execute  →  on_subgraph_http_request
```

At every hook after `on_graphql_params`, the operation AST is exposed as an **immutable reference**:

| Hook | Field | Type |
|------|-------|------|
| `on_query_plan` | `filtered_operation_for_plan` | `&OperationDefinition` |
| `on_execute` | `operation_for_plan` | `&OperationDefinition` |
| `on_subgraph_execute` | `execution_request.query` | `&str` |

There is no hook where a plugin can modify the parsed operation AST before query planning.

The only mutable entry points are:
- `on_graphql_params` — can replace the raw query **string** (pre-parse), requiring us to parse, filter, and re-serialize the query ourselves
- `on_subgraph_http_request` — can replace the raw HTTP **body bytes** (post-plan), which is too late and very hacky

Neither is ideal. The natural place to do this is between validation and query planning, operating on the typed AST rather than raw strings.

## Potential Solutions

1. **Mutable operation in `on_query_plan`**: If `filtered_operation_for_plan` were `&mut OperationDefinition`, plugins could call `selection_set.items.retain(...)` to filter fields before planning. Minimal API change.

2. **Operation replacement method**: A `set_operation(OperationDefinition)` method on the `on_query_plan` payload, letting plugins swap in a modified operation. Avoids `&mut` lifetime concerns.

3. **New dedicated hook** (e.g. `on_operation_filter`) between validation and planning, receiving an owned `OperationDefinition` that the plugin returns (possibly modified).

Any of these would let us filter auth-required fields at the AST level — no string manipulation, no re-parsing, and the query planner only generates fetch nodes for fields the client can actually access.

kamilkisiela:
Introduces a new `OnGraphqlAnalysis` plugin hook that runs right before query planning. It serves similar purpose as coprocessor's `graphql.analysis` stage, one example would be injection of flags for progressive override.

It's a perfect place to perform rate limiting or authorization on the operation (its fields) as it's the last step before query planning and has all the context needed to make decisions.  Closes #1086

Instead of letting users to replace the entire operation, without guarantee of its correctness and it getting out of sync with the projection plan, I created an API that lets users filter fields out - it takes care of null-bubbling, adjustments of the field projection plans etc.

```rust
    async fn on_graphql_analysis<'exec>(
        &'exec self,
        payload: &mut OnGraphqlAnalysisHookPayload<'exec>,
    ) -> OnGraphqlAnalysisHookResult {
        payload.filter_operation(|selection| match selection {
            Selection::Field(field)
                if field.parent_type_name == "User" && field.field_name == "email" =>
            {
                selection.null(GraphQLError::from_message_and_code(
                    "Access to field 'email' is not allowed",
                    "FIELD_ACCESS_DENIED",
                ))
            }
            _ => selection.keep(),
        });

        OnGraphqlAnalysisHookResult::Proceed
    }
```

Our authorization pipeline performs similar logic - it nulls unauthorized fields - so the new `OperationFIlter` struct is reused there too now. Plugins get now the same correctness guarantees that authorization already had.

---

I also moved things around to clean things up


lib/executor/src/operation_filter.rs
```diff
@@ -0,0 +1,263 @@
...
+            .get_type_fields(parent_type_name)
+            .and_then(|fields| fields.get(field_name))
+            .ok_or_else(|| OperationFilterError::FieldNotFound {
+                parent_type_name: parent_type_name.to_string(),
+                field_name: field_name.to_string(),
+            })?;
+        let output_type_name = field_type_info.output_type_name.as_str();
+        let is_non_null = field_type_info.nullability.is_non_null();
+
+        let decision = match self.decisions.entry((parent_type_name, field_name)) {
```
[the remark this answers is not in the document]
kamilkisiela: it's expected

lib/executor/src/plugins/hooks/on_graphql_analysis.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        let schema_metadata = self.schema_metadata;
+        let operation = self.filtered_operation_for_plan;
+
+        let root_type_name = match operation.operation_kind {
+            None | Some(OperationKind::Query) => schema_metadata.query_type.as_deref(),
+            Some(OperationKind::Mutation) => schema_metadata.mutation_type.as_deref(),
+            Some(OperationKind::Subscription) => schema_metadata.subscription_type.as_deref(),
+        }
+        .expect("root type name not found in schema metadata");
```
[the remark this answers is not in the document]
kamilkisiela: this is incorrect state in general

bin/router/benches/router_benches.rs
```diff
@@ -92,14 +92,17 @@ fn authorization_benchmark(c: &mut Criterion) {
...
+            black_box(
+                apply_authorization_to_operation(
+                    bubble_up.normalized_payload,
+                    bubble_up.auth_metadata,
+                    bubble_up.schema_metadata,
+                    bubble_up.variable_payload,
+                    &jwt_req_details,
+                    false,
+                )
+                .unwrap(),
```
kamilkisiela, line 104: now `apply_authorization_to_operation` returns `Result<T,E>`

bin/router/src/pipeline/authorization/metadata.rs
```diff
@@ -32,6 +33,18 @@ impl UserAuthContext {
                 .collect(),
         }
     }
+
+    pub fn from_jwt(
```
kamilkisiela, line 37: It fits better here

bin/router/src/pipeline/authorization/metadata.rs
```diff
@@ -74,15 +87,6 @@ where
         visited: &mut HashSet<String>,
     ) -> bool;
 
-    fn is_type_authorized(&self, type_name: &str, user_context: &UserAuthContext) -> bool;
```
kamilkisiela, line 77: those required `UserAuthContext` and I moved them to `AuthorizationChecker` that contains everything that's needed and is next to where the functions are called.

bin/router/src/pipeline/authorization/mod.rs
```diff
@@ -108,37 +184,18 @@ pub fn enforce_operation_authorization(
...
-                    normalized_operation_hash: hashes.combined_operation_hash,
-                    root_type_name: normalized_payload.root_type_name,
-                    projection_plan: Arc::new(new_projection_plan),
-                    operation_identity: normalized_payload.operation_identity.clone(),
-                }),
-                errors,
-            )
-        }
+        } => (
+            normalized_payload.with_operation(new_operation_definition, new_projection_plan),
```
kamilkisiela, line 194: The hashing, cloning etc, basically the part that was before, is now within `with_operation` method and reused in a few places.

bin/router/src/pipeline/authorization/mod.rs
```diff
@@ -152,80 +209,89 @@ pub fn apply_authorization_to_operation(
...
+
+    let operation_filter_output = OperationFilter::new(schema_metadata).filter(
         normalized_payload.root_type_name,
-        schema_metadata,
+        &normalized_payload.operation_for_plan.selection_set,
         variable_payload,
-        auth_metadata,
-        &user_context,
-    );
+        |selection| match selection {
```
kamilkisiela, line 236: The new OperationFilter struct that is used in the new hook and also for authorization directives.

In future it will be reused for `@policy` and field-level rate limiting and shit like that

bin/router/src/pipeline/authorization/mod.rs
```diff
@@ -152,80 +209,89 @@ pub fn apply_authorization_to_operation(
...
-        return AuthorizationDecision::NoChange;
+        return Ok(AuthorizationDecision::NoChange);
     }
 
-    // Phase 1: Collect authorization status for all fields
+    // Phase 1 & 2: Walk the operation, decide per-field/per-fragment
+    // authorization, and let `OperationFilter` bubble non-null rejections up to
+    // the nearest nullable ancestor.
 
-    let collection_result = collect_authorization_statuses(
```
kamilkisiela, line 169: the entire collector is now replaced with OperationFilter

bin/router/src/pipeline/authorization/mod.rs
```diff
@@ -152,80 +209,89 @@ pub fn apply_authorization_to_operation(
...
-    auth_metadata: &AuthorizationMetadata,
-) -> UserAuthContext {
-    match jwt_request_details {
-        JwtRequestDetails::Authenticated { scopes, .. } => {
-            UserAuthContext::new(true, scopes.as_deref().unwrap_or(&[]), auth_metadata)
-        }
-        JwtRequestDetails::Unauthenticated => UserAuthContext::new(false, &[], auth_metadata),
-    }
+        errors,
+    })
```
kamilkisiela, line 294: Same logic as before

bin/router/src/pipeline/nullify/rebuilder.rs
```diff
@@ -41,139 +41,123 @@ pub(super) fn rebuild_authorized_operation<'op>(
     }
 }
```
kamilkisiela, line 45: Same logic as before

bin/router/src/pipeline/execution.rs
```diff
@@ -37,7 +37,7 @@ pub struct PlannedRequest<'req> {
     pub query_plan_payload: &'req QueryPlan,
     pub variable_payload: Arc<CoerceVariablesPayload>,
     pub client_request_details: Arc<ClientRequestDetails<'req>>,
-    pub authorization_errors: Vec<AuthorizationError>,
+    pub initial_errors: Vec<GraphQLError>,
```
kamilkisiela, line 40: We needed to accept errors coming from the operation filter as well, so I merged them into just `initial_errors` of `GraphQLError` type.

bin/router/src/pipeline/mod.rs
```diff
@@ -673,6 +681,38 @@ pub async fn execute_pipeline<'exec>(
         }
     }
 
+    let client_request_details = Arc::new(client_request_details.freeze());
```
kamilkisiela, line 684: It was freezed before, just a bit later in code, no impact.

bin/router/src/pipeline/mod.rs
```diff
@@ -673,6 +681,38 @@ pub async fn execute_pipeline<'exec>(
...
+            match plugin.on_graphql_analysis(&mut analysis_payload).await {
+                OnGraphqlAnalysisHookResult::Proceed => {}
+                OnGraphqlAnalysisHookResult::EndWithResponse(response) => {
+                    return Ok(QueryPlanExecutionResult::Single(response));
+                }
+            }
+        }
+
+        let filter_output = analysis_payload.run_operation_filters()?;
+        if filter_output.has_changes() {
```
kamilkisiela, line 711: If operation filter produced some changes, we rebuild the normalized payload

bin/router/src/pipeline/mod.rs
```diff
@@ -711,13 +751,16 @@ pub async fn execute_pipeline<'exec>(
...
 
-    let client_request_details = client_request_details.freeze();
     let planned_request = PlannedRequest {
         normalized_payload: normalize_payload,
         query_plan_payload: &query_plan_payload,
         variable_payload: variable_payload.clone(),
-        client_request_details: client_request_details.into(),
-        authorization_errors,
+        client_request_details: client_request_details.clone(),
+        initial_errors: authorization_errors
```
kamilkisiela, line 759: see, we now combine auth errors with plugin errors into one list of errors

bin/router/src/pipeline/normalize.rs
```diff
@@ -51,6 +57,49 @@ impl<'a> From<&'a OperationIdentity> for GraphQLSpanOperationIdentity<'a> {
...
+/// based on the filter output and the original payload.
+pub(crate) trait FilterOutputExt<'exec> {
+    fn apply_to(
+        self,
+        payload: &GraphQLNormalizationPayload,
+    ) -> (Arc<GraphQLNormalizationPayload>, Vec<GraphQLError>);
+}
+
+impl<'exec> FilterOutputExt<'exec> for OperationFilterOutput<'exec> {
+    fn apply_to(
```
kamilkisiela, line 92: This is where we "connect" OperationFilter (executor crate) with GraphQLNormalizationPayload (router crate)

e2e/src/authorization_directives_filter.rs
```diff
@@ -258,14 +258,14 @@ mod authorization_directives_in_filter_mode_e2e_tests {
               "message": "Unauthorized field or type",
               "extensions": {
                 "code": "UNAUTHORIZED_FIELD_OR_TYPE",
-                "affectedPath": "me"
+                "affectedPath": "topProducts.price"
```
kamilkisiela, line 261: different order of errors

e2e/src/authorization_directives_filter.rs
```diff
@@ -1622,6 +1621,8 @@ mod authorization_directives_in_filter_mode_e2e_tests {
           }
         "#);
 
+        assert!(res.status().is_success(), "Expected 200 OK");
```
kamilkisiela, line 1624: I moved it to see graphql errors when they happen

e2e/src/authorization_directives_reject.rs
```diff
@@ -198,21 +198,21 @@ mod authorization_directives_in_reject_mode_e2e_tests {
               "message": "Unauthorized field or type",
               "extensions": {
                 "code": "UNAUTHORIZED_FIELD_OR_TYPE",
-                "affectedPath": "me"
+                "affectedPath": "topProducts.shippingEstimate"
```
kamilkisiela, line 201: different order of errors

lib/executor/src/introspection/schema.rs
```diff
@@ -71,6 +71,9 @@ pub struct SchemaMetadata {
     pub scalar_types: HashSet<String>,
     pub union_types: HashSet<String>,
     pub interface_types: HashSet<String>,
+    pub query_type: Option<String>,
```
kamilkisiela, line 74: The root types were never part of the SchemaMetadata and I needed them.
Instead of relying on hard-coded `Query` or `Mutation`, I added logic that extracts them from either the schema definition or object type definitions (when there's no schema definition available)

lib/executor/src/plugins/hooks/on_graphql_analysis.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        let mut visitors = self.operation_filter_visitors;
+        OperationFilter::new(schema_metadata).filter(
+            root_type_name,
+            &operation.selection_set,
+            self.variable_payload,
+            |selection| {
```
kamilkisiela, line 101: We run visitors from all plugins when iterating over an operation, instead of iterating an operation for every visitor.

lib/executor/src/plugins/hooks/on_graphql_analysis.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub fn filter_operation(
+        &mut self,
+        visitor: impl FnMut(&Selection<'exec>) -> FilterDecision + Send + 'exec,
+    ) {
+        self.operation_filter_visitors.push(Box::new(visitor));
+    }
+
+    pub fn run_operation_filters(
+        self,
+    ) -> Result<OperationFilterOutput<'exec>, OperationFilterError> {
```
kamilkisiela, line 81: Notice the `Result<T,E>` type. When an error occurs during operation filtering, we don't want to expose that to the plugin author, we just take care of it, short-circuit the execution and turn it into a PipelineError

lib/executor/src/operation_filter.rs
```diff
@@ -0,0 +1,302 @@
...
+
+pub struct TypeConditionInfo<'exec> {
+    /// `User` in `... on User { ... }`
+    pub type_condition: &'exec str,
+    /// The type the fragment is written inside before it becomes more specific
+    /// `Node` in `node { ... on User { ... } }`
+    pub parent_type_name: &'exec str,
+}
+
+impl Selection<'_> {
```
kamilkisiela, line 39: Two helpers so users would:
```rust
|selection| {
  // ...
  selection.keep()
  // or
  selection.reject(...)
}
```

instead of:
```rust
use hive_router::path::to::FieldDecision;

|selection| {
  // ...
  FieldDecision::Keep
  // or
  FieldDecision::Reject { error: ... }
}
```

lib/executor/src/operation_filter.rs
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    pub fn filter(
+        &self,
+        root_type_name: &'exec str,
+        selection_set: &'exec SelectionSet,
+        // We need variables to be able to evaluate `@skip`/`@include` directives,
+        // and ignore selections that are not included.
+        variable_payload: &'exec CoerceVariablesPayload,
+        mut visitor: impl FnMut(&Selection<'exec>) -> FilterDecision,
```
kamilkisiela, line 110: I did the visitor in a way that does not require user to define the argument type:

```rust
visitor: |selection| { ... }
// instead of:
visitor: |selection: &Selection<'exec>| { ... }
```

plugin_examples/field_nulling/src/plugin.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    fn on_plugin_init(payload: OnPluginInitPayload<Self>) -> OnPluginInitResult<Self> {
+        payload.initialize_plugin_with_defaults()
+    }
+
+    async fn on_graphql_analysis<'exec>(
+        &'exec self,
+        payload: &mut OnGraphqlAnalysisHookPayload<'exec>,
+    ) -> OnGraphqlAnalysisHookResult {
+        payload.filter_operation(|selection| match selection {
```
kamilkisiela, line 34: and here's the example

kamilkisiela, line 34: You see? No type in the argument

plugin_examples/field_nulling/src/plugin.rs
```diff
@@ -0,0 +1,48 @@
...
+            Selection::Field(field)
+                if field.parent_type_name == "User" && field.field_name == "email" =>
+            {
+                selection.reject(GraphQLError::from_message_and_code(
```
kamilkisiela, line 38: No `use` statements to pull some structs and enums

kamilkisiela, line 38: I deliberately used `reject` instead of `null` to force users to pass the error and leave some space for (not sure ever needed) `null()`.

dotansimha: love it

dotansimha: Yes, explicit > implicit

bin/router/src/pipeline/authorization/metadata.rs
dotansimha: yes

bin/router/src/pipeline/execution.rs
dotansimha: ok

dotansimha (approved):

PR #1283, opened 2026-07-15
feat(router): logger improvements, access logs 

### LINKED ISSUE #503: logger identifier for the schema version

### LINKED ISSUE #755: Correlate log lines with request identifiers

### LINKED ISSUE #765: Access logs

dotansimha:
Closes https://github.com/graphql-hive/router/issues/755 
Closes https://github.com/graphql-hive/router/issues/765 
Closes https://github.com/graphql-hive/router/issues/503 

Docs PR: https://github.com/graphql-hive/docs/pull/147 
Docs preview: https://a507d5b7-hive-platform-docs.theguild.workers.dev/graphql/hive/docs/router/observability/logging

## Background 

Extracted only the crucial changes from https://github.com/graphql-hive/router/pull/775 and re-implements major parts of it. 

### Disable logging from internal crates 

A new config flag `log.log_internals` (default: `false`) has been added, that creates a `Targets` filter. It has a hardcoded list of log targets (mostly `ntex`) to remove and avoid printing. Most of `ntex` internal logging are noise, and we would like to avoid logging it, even when `debug` level is set. 

A user can enable these by setting `log.log_internals=true`, or by setting `LOG_INTERNALS=true` in the env. 

### Improved existing log lines 

This PR also iterated all of the existing log lines (specifically the debug/info/warn/error levels) and adjusted the following:

* User-based errors (like validation/parsing/invalid requests) are logged as warning, as they do not imply a router error. 
* Router-based/internal-errors (like failure to load schema or apply an internal process in the pipeline) are printed as errors level.
* All log lines have `target` configured explicitly, so log lines are now much more clear and easy to read or filter.
* Everything related to the router workflow/pipeline is printed as `debug`. No log lines are printed on the hotpath, expect for the summary (see below).
* No nested `fields` or nested structued are printed, to allow log ingestors to process logs easily. 
* Added new log lines in decision areas of the Router, all under `DEBUG` level. 

### Structured Logging 

I've iterated all existing log lines and modified their attributes to be printed as-is, instead of printing it using a raw string. This allow JSON logging to be printed nicely and inspected by tools like DataDog or Grafana. 

The `message` of each log line is only there to indicate and identify the even, and no log message contains information about the event. All data and fields are tracked as standalone fields, and they are not nested. 

### Performance impact 

`tracing` crate internal processes can be heavy. To avoid a major perf regression, while still allowing usersto get meaningful logs that are all-or-nothing, we've implement several mechanism and applied decisions to find the right balance:

- We do not use `tracing`'s internal request-id correlation mechanism that is based on `span`s - this has a major impact since it records too many data items. Instead, we have a task-local implementation that tracks exactly what we need. 
- Custom logs formatters are in charge of correlation the log line with the request. 
- JSON log formatter is implemented from scratch, without performing any serialization (no serde / sonic_rs or any other itermediate representation). 
- The default log level is `INFO` and it prints only the request summary (see below), everything else is `DEBUG`. 
- All log lines that are lower than `DEBUG` in our code ( = `TRACE`) are dropped at compile time.

### Custom log formatters 

For `json` logging, we've implemented a custom formatter that does not construct any object, and avoid any JSON object allocations, or serialization. This means, that we construct the JSON output log lines, with a simple `buffer` that's also shared to avoid re-allocation for every log line. 

For `text` logging, we use the `compact` format, but avoid re-allocation of new objects when possible.

The new, custom, log formatters are also in charge of injecting the request identifiers into every log line. 

### Access Logs / Request Summary

While running the request, we attached a `RequestSummary` object that has a set of pre-defined attributes that are relevant for a request summary: 

```rs
pub struct RequestSummary {
    pub client_name: OnceLock<String>,
    pub client_version: OnceLock<String>,
    pub operation_name: OnceLock<String>,
    pub operation_type: OnceLock<&'static str>,
    pub operation_hash: OnceLock<String>,
    pub persisted_document_id: OnceLock<String>,
    pub subgraph_requests: AtomicU32,
    pub involved_subgraphs: Mutex<HashSet<String>>,
    pub error_count: AtomicU32,
    pub partial_response: AtomicBool,
    pub response_code: OnceLock<&'static str>,
    pub response_mode: OnceLock<&'static str>,
    pub status_code: AtomicU16,
    pub payload_bytes: AtomicI64,
    pub duration_ms: AtomicU64,
}
```

During execution, a task-local struct is attached to the request execution, and different part of the codebase can `record` information and hints about the execution. I tried to avoid any allocations where possible, but we can't avoid them all due to the need to use `'static` lifetime for the task-local variable. 

Once a request completes, a single log line is printed on the `INFO` level, summarizes the request overall: 

> `2026-07-22T06:03:46.293394Z  INFO router::request: operation_name=“IntrospectionQuery” operation_hash=“1a9246e236afb66a7cccaf18c07c38f2" subgraph_requests=0 involved_subgraphs=“” error_count=0 partial_response=false response_mode=“dual” status_code=200 payload_bytes=17302 duration_ms=3 request_id=629431863395501115` 

### Logs Correlation

In order to implement full request correlation, every incoming request is being inspected for both request-id and traceparent information. Both can be configured, or fully disabled from the config. 

```yaml
log:
  correlation:
    id_header: x-request-id
    trace_propagation: true
```

image

In order to do that, we create a tokio task-specific variable that holds the request identifiers extracted from the request. Then the task-local data is passed to the request processing `Future`. 

If a request doesn't have a `X-Request-ID` header, then Router will generate one for it, so logs lines are always correlated.  

### Subscriptions support 

Same as HTTP, subscriptions also extract correlations and attached them, either based on the initial message headers, or based on the per-operation headers. 

We do not print a log line for the initial HTTP part of Subscriptions, and we consider every operation execution on the subscriptions channel as a separate execution when it comes to requst-id correlation. 

Request correlation is based on the headers that can be passed over the init message, or subscribe message. 

The request-id and request-summary are being passed to the execution thread (as `Arc`) of subscriptions, so they can also correlate log lines and report summary details without effort.

For stream-based HTTP requests, I've implemented mechanism to "bind"/attach the summary and request-identifier to the response, so when `ntex` finally drops the response, it will `Drop` the summary object that will print the correct summary information. 
Also, `duration_ms` for stream-based reports the final time of the request, since it was staretd. For `payload_bytes` it will be an accumulation of all bytes sent in all events. 

### Improved log filter

The existing `log.log_filter` (or `LOG_FILTER` env var) can now mute noise from the router itself, by doing something like `LOG_FILTER=router::supergraph=off`. This allow more granular control over specific targets and parts of the Router logs. 

### Non-blocking stdout logging 

Instead of using the default `stdout`, we are now using `tracing-appender`'s solution for non-blocking writing https://docs.rs/tracing-appender/latest/tracing_appender/#non-blocking-writer.

Also, the output is buffered so it won't print every character to the `stdout` immediately, to prevent over-flushing of data to the stdout. 

### Testing 

Added some e2e to cover the important log lines, log attributes, req-id correlations, and also to test and cover the impact on OTEL (due to the "fake" span we create for the logging). 

## Breaking Changes 

* Log level `trace` no longer available 
* Log formats `pretty`, `tree` and `pretty-compact` no longer available, only `text` and `json` are available now.

## TODO 

- [x] apply core changes from previous PR
- [x] changes to code to make it more logging-friendly
- [x] http request logging
- [x] subgraph calls logging
- [x] graphql layer logging
- [x] env vars control 
- [x] bring back env-based log filter
- [x] `trace` level in development only
- [x] e2e testing
- [x] subscriptions 
- [x] figure out perf regression
- [x] subgraph call target 
- [x] subgraph http logging 
- [x] schema id in the log line 
- [x] subscriptions support (correlation over `spawn`)
- [x] stream support (`duration_ms` + `payload_bytes`) 
- [x] sonyflake?
- [x] changeset
- [x] docs


bin/router/src/pipeline/parser.rs
```diff
@@ -151,6 +155,8 @@ pub async fn parse_operation_with_cache(
                         err.0.errors.first()
                     {
                         if *msg == "Token limit exceeded" {
+                            warn!(target: targets::GRAPHQL_PARSING, "token limit exceeded");
+
```
enisdenjo, line 159: since INFO is the default, and this is WARN - an unauthenticated client can produce an extra log line for every malformed request, in addition to the request summary. this creates a straightforward log amplification path.

in general, I'd keep client-caused details at DEBUG and let the INFO request summary carry the error code and count

```suggestion
                            debug!(target: targets::GRAPHQL_PARSING, "token limit exceeded");
```

bin/router/src/pipeline/parser.rs
```diff
@@ -161,22 +167,23 @@ pub async fn parse_operation_with_cache(
                             );
                         }
                     }
-                    error!("Failed to parse GraphQL operation: {}", err);
+
+                    warn!(target: targets::GRAPHQL_PARSING, error = ?err, "failed to parse GraphQL operation");
```
enisdenjo: same reasoning as above

```suggestion
                    debug!(target: targets::GRAPHQL_PARSING, error = ?err, "failed to parse GraphQL operation");
```

lib/internal/src/telemetry/logging/request_id.rs
```diff
@@ -0,0 +1,162 @@
...
+            return Some(context_ref.trace_id());
+        }
+
+        None
+    }
+
+    fn extract_req_id(&self, headers: &impl HeaderLookup) -> RequestId {
+        if let Some(req_id_header) = headers.lookup_str(self.cfg.id_header.get_header_ref()) {
+            return RequestId::FromRequest(req_id_header.to_string());
+        }
```
enisdenjo: remember that the request ID header is an untrusted header, not having a length limit is an issue. a large header value can multiply into many large log lines and allocations

```suggestion
        if let Some(req_id_header) = headers
            .lookup_str(self.cfg.id_header.get_header_ref())
            .filter(|value| !value.is_empty() && value.len() <= 128)
        {
            return RequestId::FromRequest(req_id_header.to_string());
        }
```

maybe also have a character allowlist? making sure the request id is easily readable and can be pasted in all log consumers as is

lib/internal/src/telemetry/logging/request_id.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    pub fn trace_id(&self) -> Option<&str> {
+        self.trace_id.as_deref()
+    }
+}
+
+impl RequestIdentifierExtractor {
+    pub fn new(cfg: CorrelationConfig) -> Self {
+        Self {
+            generator: Sonyflake::new().expect("Sonyflake generator should initialize successfully; ensure system clock is correct and network interfaces are available"),
```
enisdenjo, line 47: yeah copilot asked already, but are we sure? how often and when can this happen? is it truly that rare?

bin/router/src/pipeline/websocket_server.rs
```diff
@@ -360,329 +375,362 @@ async fn handle_text_frame(
...
+                  debug!(target: targets::GRAPHQL_EXECUTION, error_count = shared_response.error_count(), "graphql request completed");
+
+                  summary::record(|s| {
+                      s.error_count.store(
+                          shared_response.error_count() as u32,
+                          std::sync::atomic::Ordering::Relaxed,
+                      );
+                      s.set_duration(started_at.elapsed());
+                  });
+                  summary::emit();
```
enisdenjo: WebSocket failures before line 362, including missing or retired supergraphs, happen before request correlation and summary scopes are installed. furthermore, parse, validation, authorization, and execution failures later in the inner future return early before `summary::emit()` at this line. these operations therefore produce no request summary.

maybe create the merged headers, identifiers, and summary before entering the operation body, then emit once after the body returns rather than only on the success path?

bin/router/src/lib.rs
```diff
@@ -120,20 +131,90 @@ impl BackgroundTask for CallbackServer {
...
+            "http request completed",
+        );
+
+        summary::record(|s| {
+            s.status_code
+                .store(status_code, std::sync::atomic::Ordering::Relaxed);
+            s.payload_bytes
+                .store(payload_bytes, std::sync::atomic::Ordering::Relaxed);
+            s.set_duration(started_at.elapsed());
+        });
```
enisdenjo, line 209: this emits the request summary as soon as the response object is created. `duration_ms` does not include stream delivery and `payload_bytes` is always `-1`, so the event is not a "completed" request access log. this is can be misleading for subscriptions that may remain open for hours (specifiaclly the duration)

maybe for streams log "http request started" + "http request ended"?

bin/router/src/telemetry.rs
```diff
@@ -368,62 +391,48 @@ async fn metrics_handler(registry: web::types::State<prometheus::Registry>) -> H
...
+pub fn init_logging<S>(config: &LoggingConfig) -> Result<(DynLayer<S>, WorkerGuard), ParseError>
 where
     S: tracing::Subscriber
         + for<'span> tracing_subscriber::registry::LookupSpan<'span>
         + Send
         + Sync,
 {
+    let stdout_stream = std::io::stdout();
+    let is_terminal = stdout_stream.is_terminal();
+    let (stdout_writer, stdout_guard) = tracing_appender::non_blocking(stdout_stream);
```
enisdenjo: [`tracing_appender::non_blocking` is lossy by default!](https://github.com/tokio-rs/tracing/blob/d9d4c542de10f5d3a711b7a45ffe450fd0666437/tracing-appender/src/non_blocking.rs#L203) logs cant ever be lossy

```suggestion
    let (stdout_writer, stdout_guard) =
        tracing_appender::non_blocking::NonBlockingBuilder::default()
            .lossy(false)
            .finish(stdout_stream);
```

lib/internal/src/telemetry/logging/format_text.rs
```diff
@@ -0,0 +1,49 @@
...
+            let line = buf.strip_suffix('\n').unwrap_or(&buf);
+            writer.write_str(line)?;
+
+            let _ = REQUEST_IDENTIFIERS.try_with(|ids| -> std::fmt::Result {
+                write!(writer, " request_id={}", ids.req_id())?;
+                if let Some(trace_id) = ids.trace_id() {
+                    write!(writer, " trace_id={}", trace_id)?;
+                }
+                Ok(())
+            });
```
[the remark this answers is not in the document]
enisdenjo: ^ yes

lib/internal/src/telemetry/logging/request_id.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+impl RequestId {
+    pub fn as_str(&self) -> &str {
+        match self {
+            RequestId::FromRequest(s) => s.as_str(),
+            RequestId::Generated(s) => s.as_str(),
+        }
+    }
+}
```
enisdenjo: the whole purpose of `HeaderLookup` is to bridge two concrete header map types and `RequstId` has two variants with identical storage and behavior. I dont think that either abstractions are necessary

instead just use a `String` for the ID and pass the looked-up `Option<&str>` into the extractor - i.e. pluck out the header at the request identifying call sites

```diff
pub struct RequestIdentifiers {
-   req_id: RequestId,
+   req_id: String,
    trace_id: Option<String>,
}
```

enisdenjo (changes requested):
- do we really want to maintain our own JSON writer for logging? do we have concrete evidence that this is really faster? maybe add a focused benchmark just for evaluating whether it's really faster, for proof and for potential future switching of the JSON writer if we find out that this one has (security) issues
- `RequestSummary` with atomics, a mutex, a hash set, and several locks is allocated for every request even when `router::request=off`
- https://github.com/graphql-hive/router/pull/1283/changes#r3631758193

dotansimha:
> * do we really want to maintain our own JSON writer for logging? do we have concrete evidence that this is really faster? maybe add a focused benchmark just for evaluating whether it's really faster, for proof and for potential future switching of the JSON writer if we find out that this one has (security) issues

I noticed a slight perf improvement, but that's not everything. We can't use the built-in `fmt().json()` formatter because that's the part that today injects the contentual parameters from the `span` (when you use `.instrument(...)`). This is a major perf killer because of so maybe thing tracked within that `span` and the way it's being processed. 
So to print a custom attribute (our req_id/trace_id), we have to use a custom formatter. If we already do that, the extra effort to just print the string instead of using intermediate JSON objects just makes sense. 

> `RequestSummary` with atomics, a mutex, a hash set, and several locks is allocated for every request even when router::request=off

Good point. Fixed in https://github.com/graphql-hive/router/pull/1283/commits/a6f0c809f6ca070959daa4e0da5b10452b6d1c13

lib/internal/src/telemetry/logging/request_id.rs
dotansimha: Good point. Fixed in https://github.com/graphql-hive/router/pull/1283/commits/6c0a0f61a409f1b2cc9afd470263265579c5dd1a

bin/router/src/telemetry.rs
dotansimha: Good catch. Fixed in [`5eaaf40` (this PR)](https://github.com/graphql-hive/router/pull/1283/commits/5eaaf4079ee54c26c550ce374eb1a3a56647b8d5)

lib/internal/src/telemetry/logging/request_id.rs
dotansimha: Super rare: https://codeberg.org/enra/sonyflake-rs/src/branch/main/src/builder.rs#L67-L98 

But for the sake of avoiding this complexity - do you think we should just replace it now with `ulid` and "accept" the extra `ns`? (initial benchmark shows it has a super low impact)

lib/internal/src/telemetry/logging/request_id.rs
dotansimha: super good catch. I've added a sanitation in https://github.com/graphql-hive/router/pull/1283/commits/3f875a8c624697e46164328219b2051c8619b508

bin/router/src/pipeline/parser.rs
dotansimha: good point, thanks

bin/router/src/pipeline/parser.rs
dotansimha: fixed

bin/router/src/telemetry.rs
```diff
@@ -368,62 +391,51 @@ async fn metrics_handler(registry: web::types::State<prometheus::Registry>) -> H
...
         + for<'span> tracing_subscriber::registry::LookupSpan<'span>
         + Send
         + Sync,
 {
+    let stdout_stream = std::io::stdout();
+    let is_terminal = stdout_stream.is_terminal();
+    let (stdout_writer, stdout_guard) =
+        tracing_appender::non_blocking::NonBlockingBuilder::default()
+            .lossy(false)
+            .finish(stdout_stream);
```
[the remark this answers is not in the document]
dotansimha: @enisdenjo what do you think?

bin/router/src/pipeline/websocket_server.rs
dotansimha: fixed in https://github.com/graphql-hive/router/pull/1283/commits/a8513a06ac58a8ccd5ab02375c79a5e3ada16b35

bin/router/src/lib.rs
dotansimha: good point, thanks. 

> maybe for streams log "http request started" + "http request ended"? 

you mean, in addition to the summary?

lib/internal/src/telemetry/logging/request_id.rs
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        None
+    }
+
+    fn extract_req_id(&self, headers: &impl HeaderLookup) -> String {
+        if let Some(req_id_header) = headers.lookup_str(self.cfg.id_header.get_header_ref()) {
+            return req_id_header.to_string();
+        }
+
```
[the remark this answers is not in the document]
dotansimha: fixed

bin/router/src/pipeline/validation/mod.rs
```diff
@@ -146,11 +151,8 @@ pub async fn validate_operation_with_cache(
         }
 
         if !errors.is_empty() {
-            error!(
-                "GraphQL validation failed with total of {} errors",
-                errors.len()
-            );
-            trace!("Validation errors: {:?}", errors);
+            warn!(target: targets::GRAPHQL_VALIDATION, total_errors = errors.len(), "GraphQL validation failed");
+            debug!(target: targets::GRAPHQL_VALIDATION, errors = ?errors, document = %validation_operation, "validation errors");
```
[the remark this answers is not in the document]
dotansimha: printed on purpose

bin/router/src/lib.rs
enisdenjo: `duration_ms` is misleading for long living connections and if it were to show the true duration, no matter how long - I think that would be very useful info for the user.

technically a request does not _end_ after accepting a HTTP stream, so request summery/complete does not even make sense here. should the summary be moved to the end of the subscription? or split in two? what are your thoughts?

bin/router/src/telemetry.rs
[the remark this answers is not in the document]
enisdenjo: I think there's never a good argument to drop logs. even it comes at a cost of performance

lib/internal/src/telemetry/logging/request_id.rs
enisdenjo: super rare is good enough for me

bin/router/src/telemetry.rs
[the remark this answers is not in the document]
dotansimha: agree

bin/router/src/lib.rs
dotansimha: Thanks, that's great feedback. I've implement full tracking for subscriptions :) 

Meaning:
- incoming req identifiers are also captured and shared as scope for the `spawn` calls of subscriptions, so now they can contribute to overall request summary, and their log lines are not correlated (it was missing before). 
- implement a guard that's being attached to the `Response`, so only when the response completes (=stream closes), it prints the summary, so `duration_ms` is real now. 
- subscription payloads are acumulated now so `payload_bytes` contains the total payload sent.

lib/internal/src/telemetry/mod.rs
```diff
@@ -62,36 +69,41 @@ impl TelemetryContext {
 
         let mut propagators: Vec<Box<dyn TextMapPropagator + Send + Sync>> = Vec::new();
 
-        if config.trace_context {
+        if telemetry_config.trace_context {
             propagators.push(Box::new(TraceContextPropagator::new()));
         }
```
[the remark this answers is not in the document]
dotansimha: No, this one does not impact the actual logging infra. 
It's the OTEL that can contribute to the logging extractors, not the other way around. Logging can just disable it fully.

lib/internal/src/telemetry/logging/format_text.rs
```diff
@@ -0,0 +1,49 @@
+use std::cell::RefCell;
```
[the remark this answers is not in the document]
dotansimha: fixed

PR #1288, opened 2026-07-18
fix(router, query-planner): root types are no longer hard-coded

### LINKED ISSUE #1231: Renamed root operation type produces malformed subgraph operations (root fields wrapped in an inline fragment)
> **Related to #1227** (same class, different code path). #1227 fixed one instance of this pattern — deciding root/operation semantics from a hard-coded root type **name/output type** instead of `operation_kind` — in mutation detection. This is another instance of the same pattern in the fetch-selection path (`selections.rs`); noticing it while looking at #1227’s fix. Provenance note: unlike #1227, this is **not a recent regression** — the literal root-name comparison here appears longstanding, present since the single-type fetch-selection handling was written. The fix is the same shape as #1227: key on `operation_kind` / the schema’s actual root type name rather than the literals.

## Summary

If a supergraph renames a root operation type in its `schema` block (e.g. `schema { query: Query, mutation: MutationRoot }`), the query planner emits subgraph operations whose **root fields are wrapped in a top-level inline fragment** (`{ ... on MutationRoot { ... } }`) instead of a normal `mutation { ... }` selection set — a malformed operation that a subgraph will reject or mis-handle.

## Reproduction

Load a supergraph whose mutation root type is renamed (via `supergraph: source: file`), for example a schema block like:

```graphql
schema
  @link(url: "https://specs.apollo.dev/link/v1.0")
  @link(url: "https://specs.apollo.dev/join/v0.3", for: EXECUTION)
{
  query: Query
  mutation: MutationRoot   # <-- renamed root
}
```

with a `MutationRoot` type exposing a couple of fields, then plan a multi-field mutation. The emitted subgraph operation comes out as, roughly:

```graphql
# actual (malformed): root fields wrapped in an inline fragment on the renamed type,
# with no `mutation {` keyword
{ ... on MutationRoot { five: add(num: 5) ... } }

# expected: a normal mutation operation
mutation { five: add(num: 5) ... }
```

The router loads the renamed-root supergraph without complaint (`safe_parse_schema` → `Planner::new_from_supergraph`), the renamed name is preserved end-to-end (`SupergraphState.mutation_type == Some("MutationRoot")`), and the plan wraps every root fetch in `... on MutationRoot`.

## Root cause

`lib/query-planner/src/planner/fetch/selections.rs:47` decides whether to unwrap a single-type fetch's selection by comparing the type name against hard-coded literals:

```rust
if type_name == "Query" || type_name == "Mutation" || type_name == "Subscription" {
    return selections.clone();   // unwrap to a bare root selection set
}
```

The type name here is the schema's actual (renameable) root type name (`SupergraphState` reads it from the `schema` block — `state/supergraph_state.rs:212-214`, only defaulting to the literals when the block is absent). A renamed root fails this literal comparison, so the root selection is never unwrapped and stays fragment-wrapped.

## Impact

Malformed subgraph operations for any supergraph with a renamed root operation type — the subgraph rejects them or mis-executes them. It's a correctness/data-integrity issue for that configuration (operation ordering, e.g. mutation serialization, is unaffected — only the root-field shape is wrong).

## Caveat / reachability

This is triggered by an **operator-supplied supergraph** with a renamed root, not by client input — so it's a configuration footgun, not an externally-triggerable bug. I confirmed it end-to-end through the `supergraph: source: file` load path (the load path preserves the renamed name with no normalization and no rejection). I was **not** able to confirm whether standard managed composition (Apollo/Hive) ever emits a renamed root into the composed supergraph — my understanding is composition typically canonicalizes root operation types to `Query`/`Mutation`/`Subscription`, in which case managed deployments wouldn't hit this and it mainly affects hand-authored or non-standard-composed supergraphs. Worth a maintainer confirming whether a renamed root can reach the router in the managed flow.

## Suggested fix

Key the root-selection unwrap on the operation kind / the schema's actual root type name (which `SupergraphState` already holds) rather than the hard-coded string literals. Happy to send a PR.

kamilkisiela:
Removes the assumption that GraphQL root types are always named `Query`, `Mutation`, and `Subscription`

We now reads the real root type names from the schema and tracks the operation kind. It also adds tests for renamed root types.

Closes https://github.com/graphql-hive/router/issues/1231

### DIFF
--- a/bin/router/src/pipeline/demand_control/runtime.rs
+++ b/bin/router/src/pipeline/demand_control/runtime.rs
@@ -295,7 +295,7 @@ impl DemandControlRuntime {
         actual_plans_by_fetch_hash: &mut Option<AHashMap<u64, CompiledSubgraphActualCostPlan>>,
     ) -> FormulaFetchNode {
         let default_list_size = self.default_list_size_for_subgraph(service_name);
-        let root_type = supergraph_state.root_type_name(operation_kind);
+        let root_type = supergraph_state.expect_root_type_name(operation_kind);
         if let Some(actual_plans_by_fetch_hash) = actual_plans_by_fetch_hash {
             actual_plans_by_fetch_hash
                 .entry(operation.hash)
--- a/lib/executor/src/execution/plan.rs
+++ b/lib/executor/src/execution/plan.rs
@@ -123,7 +123,8 @@ pub struct QueryPlanExecutionOpts<'exec> {
     pub extensions: ExecutionResultExtensions<'exec>,
     pub client_request: Arc<ClientRequestDetails<'exec>>,
     pub introspection_context: Arc<IntrospectionContext>,
-    pub operation_type_name: &'static str,
+    pub operation_type_name: String,
+    pub operation_kind: OperationKind,
     pub executors: Arc<SubgraphExecutorMap>,
     pub jwt_auth_forwarding: Option<Arc<JwtAuthForwardingPlan>>,
     pub graphql_error_recorder: Option<GraphQLErrorMetricsRecorder>,
@@ -380,7 +381,8 @@ pub async fn execute_query_plan<'exec>(
                         path_params: client_path_params.clone(),
                     }.into(),
                     introspection_context: opts.introspection_context.clone(),
-                    operation_type_name: opts.operation_type_name,
+                    operation_type_name: opts.operation_type_name.clone(),
+                    operation_kind: opts.operation_kind.clone(),
                     executors: opts.executors.clone(),
                     jwt_auth_forwarding: opts.jwt_auth_forwarding.clone(),
                     initial_errors,
@@ -434,7 +436,7 @@ async fn execute_query_plan_with_data<'exec>(
 ) -> Result<PlanExecutionOutput, PlanExecutionError> {
     let mut errors = opts.initial_errors;
 
-    let dedupe_subgraph_requests = opts.operation_type_name == "Query";
+    let dedupe_subgraph_requests = opts.operation_kind.is_query();
 
     let mut on_end_callbacks = vec![];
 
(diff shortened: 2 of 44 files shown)


bin/router/src/pipeline/demand_control/runtime.rs
```diff
@@ -295,7 +295,7 @@ impl DemandControlRuntime {
         actual_plans_by_fetch_hash: &mut Option<AHashMap<u64, CompiledSubgraphActualCostPlan>>,
     ) -> FormulaFetchNode {
         let default_list_size = self.default_list_size_for_subgraph(service_name);
-        let root_type = supergraph_state.root_type_name(operation_kind);
+        let root_type = supergraph_state.expect_root_type_name(operation_kind);
```
kamilkisiela, line 298: Past validation, so root type is expected

lib/executor/src/execution/plan.rs
```diff
@@ -434,7 +436,7 @@ async fn execute_query_plan_with_data<'exec>(
 ) -> Result<PlanExecutionOutput, PlanExecutionError> {
     let mut errors = opts.initial_errors;
 
-    let dedupe_subgraph_requests = opts.operation_type_name == "Query";
+    let dedupe_subgraph_requests = opts.operation_kind.is_query();
```
kamilkisiela, line 439: We pass now `operation_kind`, so it's easier to check whether it's a query or mutation

lib/executor/src/introspection/schema.rs
```diff
@@ -240,6 +272,9 @@ impl SchemaWithMetadata for ConsumerSchema {
             scalar_types,
             union_types,
             interface_types,
+            query_type_name,
```
kamilkisiela, line 275: Caches so we don't have to find root types again and again

lib/executor/src/execution/plan.rs
dotansimha: nice

dotansimha (approved):
LGTM. Thanks @kamilkisiela . I rebased and fixed a few conflicts :)

PR #1316, opened 2026-07-23
chore(router): expose lab props via router config, update lab to latest

jonathanawesome:
This PR lets router operators pre-populate the embedded Hive Laboratory from the router config, and bumps @graphql-hive/laboratory to the latest release (0.2.0 → 0.2.3, see changeset).

  Until now the only Laboratory setting was enabled, with no way to seed a default operation or header short of patching the binary. Adds three optional keys under laboratory:

  - `global_headers`: a map of header name to value sent on every request the Laboratory makes to the router. Not shown or editable in the UI; a header set on an individual operation overrides a
  global one of the same name.
  - `operations`: named operations that each open in a pre-filled tab, with optional variables, headers and extensions as native YAML maps.
  - `collections`: named groups of operations shown in the Laboratory's sidebar.

```yaml
  laboratory:
    enabled: true
    global_headers:
      X-Env: staging
    operations:
      - name: GetHello
        query: "query GetHello { hello }"
```


lib/router-config/src/laboratory.rs
```diff
@@ -9,16 +9,360 @@ pub struct LaboratoryConfig {
...
+    ///   finishes becomes the base headers of the operation.
+    /// - `lab.prompt(placeholder, defaultValue)`: returns a `Promise` that resolves with the value
+    ///   the user enters.
+    /// - `CryptoJS`: the full crypto-js library.
+    /// - `console.log` / `.warn` / `.error` / `.info`: forwarded to the Laboratory's preflight log.
+    ///
+    /// Environment variables can also be referenced as `{{name}}` inside an operation's `headers`,
+    /// `variables` and `extensions`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub preflight: Option<LaboratoryPreflightConfig>,
```
dotansimha: Based on our discussion, I would prefer avoid this and just allow users to specify pre-defined headers, using some key-value. Otherwise, this might introduce a possible security concern, as we allow the config file to inject runtime parameters to the HTML of lab, which is fully executable.

dotansimha: Also, if the pre-defined headers are already part of the `operations[].headers`, then we might not need this? or do you see a need for global headers?

lib/router-config/src/laboratory.rs
```diff
@@ -9,16 +9,360 @@ pub struct LaboratoryConfig {
...
+    /// ```yaml
+    /// laboratory:
+    ///   operations:
+    ///     - name: GetHello
+    ///       query: |
+    ///         query GetHello {
+    ///           hello
+    ///         }
+    ///       variables: '{}'
+    ///       headers: '{"X-Env": "staging"}'
```
dotansimha: Any reason headers are a string?

lib/router-config/src/laboratory.rs
```diff
@@ -9,16 +9,360 @@ pub struct LaboratoryConfig {
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub variables: Option<String>,
+    /// Headers to send with this operation, as a JSON object encoded in a string.
+    ///
+    /// These apply only to this operation, and are merged on top of any headers set by the
+    /// preflight script. To set headers on every operation, use `preflight` instead.
+    ///
+    /// Supports `{{name}}` references to the Laboratory's environment variables.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub headers: Option<String>,
```
dotansimha: Any reason this map is a string? We can use `HashMap`  (or `BTreeMap` if you want to keep the order of headers)

lib/router-config/src/laboratory.rs
```diff
@@ -9,16 +9,360 @@ pub struct LaboratoryConfig {
...
+    /// The name of the operation. Used as the tab title, and must be unique across all seeded
+    /// operations.
+    pub name: String,
+    /// The GraphQL document of the operation.
+    pub query: String,
+    /// The operation's variables, as a JSON object encoded in a string.
+    ///
+    /// Supports `{{name}}` references to the Laboratory's environment variables.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub variables: Option<String>,
```
dotansimha: Same as headers - why not use  native object/map here, instead of a string?

lib/router-config/src/laboratory.rs
```diff
@@ -9,16 +9,360 @@ pub struct LaboratoryConfig {
...
+    #[serde(default)]
+    pub script: String,
+}
+
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[serde(deny_unknown_fields)]
+pub struct LaboratoryOperationConfig {
+    /// The name of the operation. Used as the tab title, and must be unique across all seeded
+    /// operations.
+    pub name: String,
```
dotansimha, line 94: isn't the name, also reflect the operation name? is it really required?

lib/router-config/src/laboratory.rs
```diff
@@ -9,16 +9,360 @@ pub struct LaboratoryConfig {
...
+    /// preflight script. To set headers on every operation, use `preflight` instead.
+    ///
+    /// Supports `{{name}}` references to the Laboratory's environment variables.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub headers: Option<String>,
+    /// The operation's GraphQL extensions, as a JSON object encoded in a string.
+    ///
+    /// Supports `{{name}}` references to the Laboratory's environment variables.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub extensions: Option<String>,
```
dotansimha: Same, native map is preferred over a string.

bin/router/src/http_utils/laboratory.rs
```diff
@@ -0,0 +1,1058 @@
...
+/// `%` must be encoded first: encoding `:` first would turn a literal `%3A` and a real `:` into the
+/// same `%253A`, reintroducing the collision this exists to prevent.
+fn encode_id_segment(segment: &str) -> String {
+    segment.replace('%', "%25").replace(':', "%3A")
+}
+
+/// The Laboratory calls `JSON.parse` on these fields, so validating here turns a confusing
+/// browser-side failure into a startup error naming the offending operation. `location` is the
+/// config path of the operation, e.g. `laboratory.operations[0]`.
+fn validate_json_object(
```
dotansimha: this can be avoided if we would just offload the parsing to the YAML config

dotansimha (changes requested):

lib/router-config/src/laboratory.rs
jonathanawesome: It's a JSON string in Lab, but makes more sense as a YAML map here. Updating!

lib/router-config/src/laboratory.rs
jonathanawesome: Went with `BTreeMap`, which won't preserve the authoring order but will sort them alphabetically.

lib/router-config/src/laboratory.rs
jonathanawesome: It's not strictly necessary here, but if the config user leaves out the name from the operation itself the tab is "Untitled". So this is a guard for the tab title.

lib/router-config/src/laboratory.rs
jonathanawesome: Went with `global_headers` per our discussion.

dotansimha (approved):

dotansimha (approved):

PR #1335, opened 2026-07-27
fix(router): don't count regular queries toward max_long_lived_clients

Meemaw:
## Problem

`max_long_lived_clients` classifies a request as "long-lived" from the `Accept` header alone (`is_long_lived_request` in `long_lived_client_limit.rs`): any request whose `Accept` header can negotiate `text/event-stream` or `multipart/mixed` reserves a slot, and holds it until the response body is fully sent.

Popular GraphQL clients advertise streaming support on **every** operation — urql sends `Accept: application/graphql-response+json, application/json, text/event-stream, multipart/mixed` on plain queries and mutations. With such clients, every in-flight request counts toward the limit, so once concurrent requests exceed the limit (default `128` per process), regular queries are rejected with `503 "Too many long-lived clients"`.

This contradicts the config's own documentation:

> Regular non-streaming requests are not counted toward this limit.

and the router's own content negotiation, which resolves these requests to `ResponseMode::Dual` — "queries are single responses, subscriptions are streams" — and correctly answers them with a single `application/json` response.

We hit this in production: a urql-based web app with subscriptions enabled saw bursty `503`s on ordinary queries at peak traffic (hundreds of concurrent in-flight requests, while actual subscription clients numbered in the single digits). The rejections were also hard to diagnose because the middleware short-circuits before any tracing/metrics run, so they were invisible server-side.

## Fix

Split the accounting by *when* each kind of long-lived client becomes identifiable:

- **WebSocket connections** — unchanged: still reserved in the middleware at upgrade time, since the upgrade headers alone are proof of a long-lived client. The guard is still attached to the response body so the slot is held for the lifetime of the connection.
- **HTTP streaming (subscriptions over SSE/multipart)** — no longer classified from the `Accept` header. The pipeline reserves a slot in `graphql_request_handler` once the parsed operation is known to be a subscription — still **before** any planning/execution work, so over-limit subscriptions are rejected as cheaply as before, with the same `503` / `Retry-After: 5` / `"Too many long-lived clients"` response. The guard is threaded into the client-facing stream (`SharedRouterStreamResponse::into_response`) and released when that client's stream ends.

Regular queries and mutations never touch the counter, regardless of their `Accept` header.

Behavior notes:

- Dedupe joiners each reserve their own slot, preserving per-client counting (same as before).
- Subscriptions over WebSocket are not double-counted: the WS *connection* holds the one slot, matching the documented "WebSocket connections" semantics (same as before).
- The middleware is now only enabled when `websocket.enabled` is set (it only handles upgrades); the pipeline check covers the subscriptions side.
- Over-limit subscription rejections now happen inside the request span, so they produce a trace instead of being invisible to telemetry.
- Config docs updated to make the semantics explicit (`docs/README.md` regenerated).

## Testing

- New e2e test `max_long_lived_clients_ignores_regular_queries`: exhausts the limit with two real SSE subscriptions, then asserts a query sent with urql's exact `Accept` header still succeeds while a third subscription is still rejected with `503`.
- Existing `max_long_lived_clients_rejects_over_limit` passes unchanged (same rejection status, header, and body).
- Full `subscriptions` e2e module passes; `cargo clippy` clean. (`telemetry::logging::test_logging_of_subscriptions` is flaky when the module runs in parallel — fails identically on unmodified `main`, passes in isolation.)

## Possible follow-up (not in this PR)

There is still no metric counting these rejections. A counter (e.g. `router.long_lived_clients.rejected`) would make hitting the limit observable without log scraping — happy to add it in a follow-up if there's interest.


enisdenjo (approved):
great work, thanks for the contribution!

PR #1342, opened 2026-07-29
feat(router): add `limits.max_request_header_size` to reject oversized request headers

Meemaw:
## Description

Adds a new `limits.max_request_header_size` config option (default: `64KiB`) that rejects GraphQL requests whose total header size exceeds the limit with **`431 Request Header Fields Too Large`**, before any processing happens.

### Motivation

The router propagates client headers (cookies, JWTs) to every subgraph. Most subgraph servers enforce their own request header size limit — e.g. Tomcat's default is 8KB — so users with large cookies get an opaque HTML 400 from the subgraph that surfaces at the router as a subgraph error (e.g. `SUBREQUEST_HTTP_ERROR`), which is confusing and wastes a full pipeline execution + subgraph fan-out. With this option the router can reject such requests up front with a clear, spec-appropriate error:

```yaml
limits:
  max_request_header_size: 8KiB
```

```json
{"errors":[{"message":"Request headers exceed the maximum allowed size","extensions":{"code":"REQUEST_HEADER_FIELDS_TOO_LARGE"}}]}
```

### Implementation notes

- The strict per-request check lives at the top of `graphql_request_handler`, mirroring the existing `limits.max_request_body_size` check (including draining a small amount of the body before responding, so the connection closes cleanly instead of being reset).
- ntex's own `max_buf_size` (the h1 parse buffer, hard-coded default 64KiB) is **not** a reliable per-request limit — it only triggers when a request head is still incomplete once the buffer is full, so heads that arrive in a single read pass through regardless of size. That's why the check is done on the parsed request. The parse buffer is still raised to `max(configured_limit, 64KiB)` so limits **above** 64KiB are actually reachable (previously any head over 64KiB that arrived fragmented was rejected by ntex with a plain 400).
- The counted size approximates the raw header block: per header, name + value + 4 bytes of on-wire overhead (`": "` + CRLF).
- Default of `64KiB` matches ntex's existing parse buffer, so there is no behavior change for existing deployments.

### Verified

- e2e tests: `header_limit.rs` (431 over limit incl. error body snapshot, 200 under limit); full `limit`-filtered e2e suite passes.
- Manually: with `max_request_header_size: 2KiB`, a 4KB cookie gets 431 and a normal request gets 200; with `128KiB`, a 100KB cookie is accepted (impossible before) and a 150KB one rejected.

### DIFF
--- a/.changeset/max_request_header_size_limit.md
+++ b/.changeset/max_request_header_size_limit.md
@@ -1,0 +1,11 @@
+---
+hive-router-config: minor
+hive-router-internal: patch
+hive-router: minor
+---
+
+# Add `limits.max_request_header_size`
+
+Adds a new `limits.max_request_header_size` configuration option (default: `64KiB`) that rejects requests whose HTTP headers exceed the configured size with `431 Request Header Fields Too Large`, before the request is processed.
+
+Since the router propagates client headers (cookies, JWTs) to subgraphs, requests with oversized headers would previously be forwarded and rejected by the subgraph server's own header limit (e.g. Tomcat's 8KB default), surfacing as a confusing subgraph error. With this limit, such requests are rejected at the router with a clear error.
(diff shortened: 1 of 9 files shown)


dotansimha (approved):
Thanks! I made a few adjustments and added more tests.

PR #1352, opened 2026-08-02
fix(router): expose correlations to plugin system

### LINKED ISSUE #1351: Logger identifiers correlation is not passed to custom plugins hooks
Some custom plugin hooks, like `on_http_request`, are running from `ntex` middleware service, and those are executed on a separate flow, before the graphql request handler is executed. 

Currently, identifiers are extracted inside the graphql request handler, which runs after, so `on_http_request` hook can't see the identifier, so custom log lines that are printed inside `on_http_request` are not correlated.

dotansimha:
Fixes https://github.com/graphql-hive/router/issues/1351 

The proposed fix moves the request correlation extraction higher (to ntex's middleware) so other middlewares and other aspects of the execution can access it fully. 

By moving it up the hirerchy, we allow `on_http_req` hook to also be fully correlated, so plugin developers can benefit from using the logger without the needs to deal with correlation manually.

### DIFF
--- a/e2e/src/telemetry/logging.rs
+++ b/e2e/src/telemetry/logging.rs
@@ -6,11 +6,18 @@ use crate::{
         Started, TestRouter, TestRouterBuilder, TestSubgraphs,
     },
 };
+use hive_router::{
+    async_trait,
+    plugins::hooks::on_http_request::{OnHttpRequestHookPayload, OnHttpRequestHookResult},
+    plugins::hooks::on_plugin_init::{OnPluginInitPayload, OnPluginInitResult},
+    plugins::plugin_trait::{RouterPlugin, StartHookPayload},
+};
 use hive_router_internal::telemetry::logging::targets;
 use http::{HeaderMap, HeaderName, HeaderValue};
 use insta::assert_json_snapshot;
 use serde_json::{Map, Value};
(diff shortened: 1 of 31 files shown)


e2e/src/telemetry/logging.rs
```diff
@@ -661,3 +668,55 @@ log:
     let http_start = log_line(&stdout_log, LOG_HTTP_REQUEST_START);
     assert!(find_attr(http_start, "trace_id").is_none());
 }
+
+#[derive(Default)]
+struct TestDebugLogPlugin;
```
enisdenjo, line 673: this test is better fitting in the plugin_examples. moving it there also has the benefit of being an example for the users and you can reference it from the docs.

enisdenjo (changes requested):

e2e/src/telemetry/logging.rs
dotansimha: Ah yes, I did a full example + a test in https://github.com/graphql-hive/router/pull/1348 and it will land after. Added a local test here just to have a confirmation also in this PR :)

enisdenjo (approved):

kamilkisiela (approved):

PR #1363, opened 2026-08-03
feat(router): authorize `@requiresScopes` from a configurable JWT claim (`jwt.scopes_claim`)

### LINKED ISSUE #1361: Support for authorization directive and Azure Entra JWT tokens
Currently the @requiresScopes directive authorizes requests by verifying the scp field of the JWT token. For Azure Entra authentication, the used approach is to create roles for applications and later assign roles to users and applications.
Would there be any push back if I created a PR to extract the scopes from a different claim and in this case authorize using roles, instead of the scope. My proposed solution would include an additional config parameter under authorization -> directives: `scopes_claim`, defaulting to the original approach of `map.get("scope").or_else(|| map.get("scopes"))` when not set.

rafagsiqueira:
Closes #1361

## Summary

- Adds an optional `jwt.scopes_claim` config option. When set, the `@requiresScopes` directive reads authorization data from that JWT claim instead of the hardcoded `scope`/`scopes` claim.
- Motivation: identity providers such as Microsoft Entra ID grant authorization data via application roles issued under a `roles` claim, not `scope`. This lets `@requiresScopes` work with such providers without requiring the IdP to be reconfigured to emit a `scope` claim.
- When `jwt.scopes_claim` is unset, behavior is unchanged (still reads `scope`, falling back to `scopes`). When set, only the configured claim is consulted — the default `scope`/`scopes` claim is no longer read, so there's no ambiguity about which claim authorizes the request.

## Changes

- `lib/router-config/src/jwt_auth.rs`: new `scopes_claim: Option<String>` field on `JwtAuthConfig`.
- `bin/router/src/jwt/context.rs`: `JwtRequestContext::extract_scopes()` reads the configured claim name when present.
- `bin/router/src/jwt/mod.rs`: threads the configured claim name into `JwtRequestContext` at construction.
- `docs/README.md`: regenerated from the config schema to document the new field.
- `.changeset/jwt_scopes_claim_config.md`: minor version bump for `hive-router-config` and `hive-router`.

## Test plan

- [x] Unit tests in `lib/router-config/src/jwt_auth.rs` covering default (`None`), omitted-field, and explicit `scopes_claim` deserialization.
- [x] Unit tests in `bin/router/src/jwt/context.rs` covering default claim extraction, custom claim extraction, and that the default claim is ignored once a custom claim is configured.
- [x] New e2e tests in `e2e/src/authorization_directives_filter.rs` (new fixture `e2e/configs/jwt_auth_custom_scopes_claim.router.yaml` with `jwt.scopes_claim: roles`):
  - a token with a `roles` claim satisfies `@requiresScopes(["read:birthday"])`
  - a token with the default `scope` claim (but not `roles`) is unauthorized once `scopes_claim` is configured, confirming there's no silent fallback
- [x] Full workspace build, `cargo fmt`/`cargo clippy` clean on touched files, all existing authorization (45) and JWT (12) e2e tests still pass, no regressions.

### DIFF
--- a/.changeset/jwt_scopes_claim_config.md
+++ b/.changeset/jwt_scopes_claim_config.md
@@ -1,0 +1,10 @@
+---
+hive-router-config: minor
+hive-router: minor
+---
+
+# Add `jwt.scopes_claim`
+
+Adds a new `jwt.scopes_claim` configuration option that lets you specify which JWT claim the `@requiresScopes` directive should read authorization data from, instead of the hardcoded `scope`/`scopes` claim.
+
+This is useful for identity providers that grant authorization data under a different claim name — for example, Microsoft Entra ID, which issues app roles under a `roles` claim rather than `scope`. Setting `jwt.scopes_claim: roles` allows `@requiresScopes` to authorize requests using Entra app roles without requiring any changes to how the claim is issued.
--- a/bin/router/src/jwt/context.rs
+++ b/bin/router/src/jwt/context.rs
@@ -11,6 +11,8 @@ pub struct JwtRequestContext {
     pub token_prefix: Option<String>,
     pub token_raw: String,
     pub token_payload: Arc<JwtTokenPayload>,
+    /// The name of the claim to read authorization scopes from. Defaults to "scope"/"scopes" when `None`.
+    pub scopes_claim: Option<String>,
 }
 
 impl JwtRequestContext {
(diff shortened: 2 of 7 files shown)


dotansimha (approved):
Thank you! I'll be waitinf for CI to clear and then will proceed to merge. This will also be part of the next release

PR #1366, opened 2026-08-04
fix(router): register the state the subscription callback handler extracts

gagalago:
## Summary

The subscription callback handler cannot serve any request. Every call fails the `ntex` `State` extractor with:

```
App state is not configured, to configure use App::state()
```

and returns 500 to the subgraph. Since callbacks are how a subgraph delivers events back to the router, subscriptions using the callback transport silently stop producing messages.

Introduced in #1261 (subscription metrics, released in `0.0.83`) and still present on `main` (`ae2d2d2`).

## Cause

`handle_callback` (`bin/router/src/pipeline/http_callback.rs`) extracts two pieces of state:

```rust
callback_subscriptions: web::types::State<CallbackSubscriptionsMap>,
telemetry_context:      web::types::State<std::sync::Arc<TelemetryContext>>,
```

Neither listener in `bin/router/src/lib.rs` registers a matching pair:

| Listener | `CallbackSubscriptionsMap` | `Arc<TelemetryContext>` |
| --- | --- | --- |
| Dedicated (`subscriptions.callback.listen` set) | ✅ `.state(cb_subs)` | ❌ `.state(telemetry.context.clone())` — `Telemetry::context` is a bare `TelemetryContext` (`bin/router/src/telemetry.rs`), not an `Arc` |
| Main (no `listen`; callback mounted on the main app) | ❌ never registered | ✅ `shared_state.telemetry_context` |

`State<T>` is generic, so both mismatches compile cleanly and only fail when the extractor runs. The result is that the callback protocol is broken in **both** supported configurations, and there is no config-level workaround.

## Fix

```rust
// dedicated listener
- let cb_telemetry_context = telemetry.context.clone();
+ let cb_telemetry_context = std::sync::Arc::new(telemetry.context.clone());

// main listener
  .state(shared_state.telemetry_context.clone())
+ .state(callback_subscriptions_for_handler.clone())
```

## Reproducing

With `subscriptions.callback.listen` configured, any callback request — including a `check` for an id that does not exist, which should be a 404 — returns the state error instead:

```
curl -X POST http://<router>:4001/callback \
  -H 'content-type: application/json' \
  -d '{"kind":"subscription","action":"check","id":"<any>","verifier":"<any>"}'
```

We hit this in production after upgrading `0.0.49 → 0.0.85`: callback checks went to 100% 500s across both environments, and we rolled back. `0.0.49` is unaffected — it has a single extractor, correctly registered.

## Notes for reviewers

- Changeset added as `hive-router: patch`.
- I have **not** been able to run the repo's own test suite against this change; the fix is derived from reading the handler signature against both registration sites. If there is an integration test that drives a real callback, it would be worth pointing it at both listener configurations — the two code paths fail for different reasons, so a test covering only one would still pass with the other broken.
- Happy to adjust the approach if you would rather change `Telemetry::context` to hold an `Arc` directly, or narrow the handler to `State<TelemetryContext>` — this keeps the handler signature untouched as the smaller change.

### DIFF
--- a/.changeset/register_callback_subscription_state.md
+++ b/.changeset/register_callback_subscription_state.md
@@ -1,0 +1,23 @@
+---
+hive-router: patch
+---
+
+# Register the state the subscription callback handler extracts
+
+The subscription callback handler could not serve a single request: every call failed the `ntex` `State` extractor with `App state is not configured, to configure use App::state()`, returning a 500 to the subgraph. Because callbacks are how the subgraph delivers events back to the router, any subscription using the callback transport silently stopped producing messages.
+
+`handle_callback` extracts two pieces of state:
+
+```rust
+callback_subscriptions: web::types::State<CallbackSubscriptionsMap>,
+telemetry_context:      web::types::State<std::sync::Arc<TelemetryContext>>,
+```
+
+Neither listener registered a matching pair:
+
+- the **dedicated** callback server (`subscriptions.callback.listen` set) registered the map, but passed `telemetry.context` — a bare `TelemetryContext`, since `Telemetry::context` is not an `Arc` — where the handler asks for `State<Arc<TelemetryContext>>`;
+- the **main** server (no `listen`, callback mounted on the main app) registered `Arc<TelemetryContext>` via `shared_state.telemetry_context`, but never registered `CallbackSubscriptionsMap` at all.
+
+`State<T>` is generic, so both mismatches compile; they only fail when the extractor runs. That makes the callback protocol broken in **both** configurations, with no config-level workaround.
+
+This registers an `Arc` on the dedicated server and adds the missing map on the main one.
--- a/.changeset/register_callback_subscription_state.md
+++ b/.changeset/register_callback_subscription_state.md
@@ -20,4 +20,6 @@ Neither listener registered a matching pair:
 
 `State<T>` is generic, so both mismatches compile; they only fail when the extractor runs. That makes the callback protocol broken in **both** configurations, with no config-level workaround.
 
-This registers an `Arc` on the dedicated server and adds the missing map on the main one.
+The dedicated server now takes its telemetry context from `shared_state.telemetry_context`, the same `Arc` the main server already registers, and the main server gains the missing map.
+
+The e2e suite does not catch this: `TestRouter` builds its own `ntex` `App` rather than reusing the one `run_router` assembles, and its copy happens to register both states correctly — it reads the telemetry context from `shared_state` (already an `Arc`) and registers `callback_subscriptions` on the main app. `listen_on_different_port` therefore passes against a build whose callback endpoint 500s on every request.
--- a/bin/router/src/lib.rs
+++ b/bin/router/src/lib.rs
@@ -374,7 +374,7 @@ pub async fn router_entrypoint(plugin_registry: PluginRegistry) -> Result<(), Ro
             let cb_path = path.to_string();
             let cb_addr = listen.to_string();
             let cb_subs = callback_subscriptions_for_handler.clone();
-            let cb_telemetry_context = std::sync::Arc::new(telemetry.context.clone());
+            let cb_telemetry_context = shared_state.telemetry_context.clone();
             let mut cb_server_builder = web::HttpServer::new(async move || {
                 let cb_subs = cb_subs.clone();
                 let cb_path = cb_path.clone();
--- a/bin/router/src/lib.rs
+++ b/bin/router/src/lib.rs
@@ -374,7 +374,7 @@ pub async fn router_entrypoint(plugin_registry: PluginRegistry) -> Result<(), Ro
             let cb_path = path.to_string();
             let cb_addr = listen.to_string();
             let cb_subs = callback_subscriptions_for_handler.clone();
-            let cb_telemetry_context = telemetry.context.clone();
+            let cb_telemetry_context = std::sync::Arc::new(telemetry.context.clone());
             let mut cb_server_builder = web::HttpServer::new(async move || {
                 let cb_subs = cb_subs.clone();
                 let cb_path = cb_path.clone();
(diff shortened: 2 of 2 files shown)


gagalago:
Update: I have now built and run the router crate tests against this branch, so the caveat in the description about not having run the suite no longer applies.

\`cargo test -p hive-router\` on \`b7a2175\` (this PR head), in \`rust:1.90-bookworm\`:

```
test result: ok. 168 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

No build errors or new warnings. Confirmed the built tree contains both changes (`lib.rs:377` and `lib.rs:431`).

Two things that remain true and are worth a maintainer's eye:

1. **No test exercises the regression.** The 168 passing tests pass equally without the fix — nothing drives a callback request, which is exactly why a mismatched `State<T>` shipped in the first place. If you would like, I am happy to add one, but it needs a decision from you on shape: the two listener configurations (`subscriptions.callback.listen` set vs unset) fail for different reasons, so a single test covering one would still go green with the other broken.

2. **I have not reproduced the fix at runtime.** A synthetic `POST /callback` for an unregistered id returns 404 locally before the extractors run, on both the patched and unpatched builds, so it proves nothing either way. Reaching the extractors needs a registered subscription id and verifier. Our own evidence is production behaviour: 100% of callback checks returned the state error on 0.0.85, and none on 0.0.49.

So: the change compiles and does not regress the existing suite, and the reasoning is from the handler signature against both registration sites — but it is not covered by a test, and I want to be straightforward about that rather than let the green run imply more than it does.

gagalago:
Correction to my previous comment: I said reaching the extractors requires a registered subscription id. That is wrong, and I want to fix the record.

The callback route takes the id as a **path segment** — `POST /callback/{subscription_id}` — as `e2e/src/http_callback.rs` in this repo already does. My earlier probe posted to `/callback` with the id in the body, which simply does not match the route, so the 404 I saw was a malformed request rather than anything about the protocol. An unknown id on the correct path should still route to the handler, and the state extractors run before the map lookup.

That means this **is** reachable from the existing e2e harness, and the interesting question becomes why the current coverage did not catch it. `http_callback.rs` has `listen_on_different_port()`, which exercises the dedicated listener — the configuration where the map *is* registered and only the telemetry state is mismatched. I have not worked out why that case passes; the type mismatch should fail there too, so either the assertion does not reach the failing path or something about the test config differs.

Worth understanding before adding tests, because it decides what to add. Two candidate gaps:

1. the **main-listener** configuration (no `subscriptions.callback.listen`), where `CallbackSubscriptionsMap` is absent entirely — I see no e2e case for it;
2. whichever property makes `listen_on_different_port()` pass today despite the telemetry mismatch.

Happy to dig into that and add coverage for both configurations if useful — I did not want to guess at it and submit a test that passes for the wrong reason.

gagalago:
Two updates: runtime evidence for the bug, and a simplification of the patch.

## Runtime reproduction

I built `main` (`ae2d2d2`, unmodified) and pointed a bare `POST` at the callback endpoint with `subscriptions.callback.listen` configured:

```
$ curl -i -X POST http://127.0.0.1:4011/callback/probe-id \
    -H 'Content-Type: application/json' -d '{}'

HTTP/1.1 500 Internal Server Error
App state is not configured, to configure use App::state()
```

Same result with a well-formed `check` payload. The `State<T>` extractors are handler arguments, so they resolve before the body — no registered subscription is needed to hit it, and no request can get past it.

With the patch applied, both listener configurations reach the handler and return the protocol-level rejection you'd expect:

```
=== dedicated listener (callback.listen set) ===
HTTP/1.1 400 Bad Request
subscription-protocol: callback/1.0

=== main listener (no callback.listen) ===
HTTP/1.1 400 Bad Request
subscription-protocol: callback/1.0
```

(The unpatched *main-listener* case is the missing `CallbackSubscriptionsMap`; I reasoned it from the types rather than measuring it — I lost the build environment before running that fourth probe. The dedicated-listener numbers above are measured.)

## Patch simplified

I'd written `std::sync::Arc::new(telemetry.context.clone())`. It's now `shared_state.telemetry_context.clone()`, which is what the main server already registers one screen down at `lib.rs:430`, and `shared_state` is in scope from `lib.rs:355`. Same type, one shared `Arc` instead of a second one, and the two listeners now read identically.

## Why CI is green on a broken callback protocol

`listen_on_different_port()` passes against unmodified `main`:

```
test http_callback::http_callback_e2e_tests::listen_on_different_port ... ok
test result: ok. 1 passed; 0 failed; 624 filtered out
```

That test drives a real SSE subscription with `callback.listen` set and asserts events arrive — it *should* be the test that catches this. It doesn't, because `TestRouter::start()` (`e2e/src/testkit/mod.rs:823-950`) builds its own `ntex` `App` rather than reusing the one `run_router` assembles, and its copy differs from `lib.rs` in exactly the two places that are broken:

| | `bin/router/src/lib.rs` | `e2e/src/testkit/mod.rs` |
|---|---|---|
| dedicated listener | `.state(telemetry.context.clone())` — bare `TelemetryContext` (`telemetry.rs:70`) | `.state(shared_state.telemetry_context.clone())` — already `Arc<TelemetryContext>` ✅ |
| main listener | `CallbackSubscriptionsMap` never registered | `.state(callback_subs)` (line 929) ✅ |

The harness independently registers both states correctly, so no e2e test can observe a state-registration mistake in `lib.rs` — the whole class is invisible, not just this instance. Worth noting the main-listener configuration has no e2e case at all.

The durable fix is to have `lib.rs` and the testkit share one app-assembly function, at which point `listen_on_different_port` covers this for free. Happy to do that here or in a follow-up — your call on scope.

If a cheaper guard is useful in the meantime: `POST /callback/{anything}` with no `subscription-protocol` header must return **400** with that header on the response. A 500 is the state bug, a bare 404 means the route moved. That's the check I added downstream and it distinguishes all three cases without needing a live subscription.

enisdenjo (approved):
thank you for the contribution! this is looking good

dotansimha:
Available in `hive-router 0.0.87`

PR #1377, opened 2026-08-06
fix(router): include persisted document hash in usage reports

### LINKED ISSUE #1343: The router never sends persisted_document_hash, so app deployments show no usage
I started to integrate hive-router into a project this week, so I am new to this code. I use
persisted documents and usage reporting together. The app deployments in Hive Console show no
usage. I think I found the cause, but please correct me if I missed a configuration step.

The router resolves the document id correctly. Then it removes the id before it sends the usage
report. Hive Console needs this hash to match a request to an app deployment. Without the hash,
the "Last used" column stays empty. The requests themselves work correctly.

`collect_usage_report` always sets the field to `None`:

https://github.com/graphql-hive/router/blob/hive-router/v0.0.86/bin/router/src/pipeline/usage_reporting.rs#L141

This function is the only place where the router makes an `ExecutionReport`. Both callers use this
function: `pipeline/mod.rs:481` and `pipeline/websocket_server.rs:625`. Therefore the problem
applies to HTTP and to WebSocket. The function also has no parameter for the hash.
`PersistedDocumentsContext` keeps only `skip_enforcement`. Thus the router loses the resolved id
before it reports the usage.

Two other components support this field:

- `hive-console-sdk` has `persisted_document_hash` in `ExecutionReport`. The agent sends the field
  in the report payload. See `lib/hive-console-sdk/src/agent/usage_agent.rs#L51`.
- The Apollo Router integration in this repository reads the hash from the context and sets the
  field. See `apollo-router-workspace/bin/router/src/usage.rs#L330`. That runtime reports the hash.
  hive-router does not report it.

Hive Console writes the `app_deployment_usage` table only when an operation includes
`persistedDocumentHash`. This table supplies the `lastUsed` value. The `graphql-client-name` and
`graphql-client-version` headers go to `metadata.client`. Insights reads `metadata.client`, but
this table does not. Therefore you cannot fill the column with the headers only.

`require_id` does not change the result. The router reports the usage after the enforcement step.

### Steps to reproduce

1. Publish an app deployment with a document manifest. Activate the deployment.
2. Set `persisted_documents.storage.type` to `hive`.
3. Enable `telemetry.hive.usage_reporting`.
4. Send an operation by its id. Use the format `<appName>~<appVersion>~<hash>`.
5. Open Hive Console. Insights shows the client name and the version. The app deployment shows
   "No data" for Last used.

My project uses version 0.0.85. Line 141 is the same on main.

### Expected result

The usage report includes the resolved document id. Then Hive Console matches the request to the
correct app deployment.

I think this function was not in the original plan. RFC #311 describes the store, the wire formats
and the policy for arbitrary operations. It does not mention usage reporting.

I can send a pull request, if you agree with the approach:

- Keep the resolved id in `PersistedDocumentsContext`.
- Add a parameter for the hash to `collect_usage_report`.
- Set the parameter at the two call sites.

azproduction:
Closes #1343

## What

`collect_usage_report` always set `persisted_document_hash: None`, so Hive Console never received the resolved document id and app deployments showed no usage ("Last used" stayed empty) even though requests were served correctly.

This follows the approach agreed in the issue:

- `collect_usage_report` gains a `persisted_document_hash: Option<&str>` parameter and forwards it into the `ExecutionReport` (the SDK already supports the field and sends it in the report payload).
- The HTTP pipeline passes `prepared_operation.resolved_document_id`. Thanks to `enforce_require_id_policy`, that field is `Some` only when the executed operation was actually resolved from a persisted document (it is cleared when the feature is disabled or when a plain `query` took precedence), so the report carries the id exactly when it should. The full document id is reported, matching the Apollo Router integration in this repo.
- The WebSocket call site passes `None` explicitly: the graphql-transport-ws `Subscribe` payload has no `documentId` field and that path never invokes the document id resolver, so there is no id to report there today. (One correction to the issue text, where I assumed both paths lose a resolved id.)

## Tests

New e2e test `usage_reporting_includes_persisted_document_hash` combining file-storage persisted documents with the mock usage endpoint:

- an operation sent by `documentId` produces a report whose operation carries `persistedDocumentHash` equal to the document id;
- a plain `query` request produces a report without the field.

`cargo fmt --check`, `cargo clippy -p hive-router`, `cargo test -p hive-router` (171 passed) and `cargo test -p e2e telemetry::usage_reporting` (15 passed) are green. A changeset (`hive-router: patch`) is included.

### DIFF
--- a/.changeset/usage-report-persisted-document-hash.md
+++ b/.changeset/usage-report-persisted-document-hash.md
@@ -1,0 +1,11 @@
+---
+hive-router: patch
+---
+
+# Report `persistedDocumentHash` in usage reports
+
+Usage reports now include the resolved persisted document id, so Hive Console can match
+requests to app deployments and populate their "Last used" data. Previously the router
+resolved the document id but always omitted it from the usage report.
+
+Closes https://github.com/graphql-hive/router/issues/1343
--- a/bin/router/src/pipeline/mod.rs
+++ b/bin/router/src/pipeline/mod.rs
@@ -509,6 +509,7 @@ pub async fn graphql_request_handler(
                 hive_usage_agent,
                 shared_response.error_count(),
                 Some(usage_reporting::request_details_from_ntex_request(req)),
+                prepared_operation.resolved_document_id.as_deref(),
             )
             .await;
         }
--- a/bin/router/src/pipeline/usage_reporting.rs
+++ b/bin/router/src/pipeline/usage_reporting.rs
@@ -118,6 +118,7 @@ pub async fn collect_usage_report<'a>(
     hive_usage_agent: &UsageAgent,
     error_count: usize,
     request_details: Option<RequestDetails>,
+    persisted_document_hash: Option<&str>,
 ) {
     let timestamp = SystemTime::now()
         .duration_since(UNIX_EPOCH)
@@ -138,7 +139,7 @@ pub async fn collect_usage_report<'a>(
             OperationKind::Subscription => OperationType::Subscription,
         }),
         operation_name: operation_name.map(|s| s.to_owned()),
-        persisted_document_hash: None,
+        persisted_document_hash: persisted_document_hash.map(|hash| hash.to_owned()),
     };
 
     if let Err(err) = hive_usage_agent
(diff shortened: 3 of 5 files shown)


enisdenjo (approved):

dotansimha:
thanks @azproduction ! This looks great. 

Regarding websocket, @enisdenjo was working on fixing that on another PR :)

dotansimha (approved):

azproduction:
@dotansimha thanks for the swift iteration on this fix 🙌

PR #1379, opened 2026-08-06
feat(router): allow to customize access log message by plugin system

### LINKED ISSUE #1378: Allow to customize summary (access log) message

dotansimha:
Closes https://github.com/graphql-hive/router/issues/1378 

* Introduces `set_summary_message` (and `get_current_summary`) to allow plugins to customize the summary message based on request outcome data.
* Updates request summary logging to carry a customizable message field
* Also, fixed a tiny syncing issue that prevented `on_end` of plugin system to have access to the summary. This also makes calculation of duration more accurate (we drop the guard only when `Response` finishes and released from memory)

### DIFF
--- a/bin/router/src/lib.rs
+++ b/bin/router/src/lib.rs
@@ -143,19 +143,22 @@ pub fn set_summary_attribute(key: impl Into<String>, value: impl Into<sonic_rs::
     summary::record(|s| s.set_custom(key, value));
 }
 
+/// Returns the current request log summary for the current request, if one exists.
 pub fn get_current_summary() -> Option<Arc<summary::RequestSummary>> {
     summary::current_summary()
 }
 
 /// Lets plugins attach a custom correlation to every log line of the current request (not
-/// just the summary), e.g. a tenant or project id extracted from the URL. Setting the same
-/// key again overwrites the previous value. A no-op outside a request.
+/// just the summary), e.g. a tenant or project id extracted from the URL.
+/// Setting the same key again overwrites the previous value. A no-op outside a request.
 pub fn set_log_correlation(key: impl Into<String>, value: impl std::fmt::Display) {
     request_id::set_correlation(key, value);
 }
 
-/// Lets plugins override the request summary log line's message. The first call wins;
-/// later calls for the same request are no-ops.
+/// Lets plugins override the request summary log line's message.
+/// This can be called only once per request, and only during the request's lifetime.
+/// Calling it more than once, for the same request is a no-op.
+/// Calling it outside of a request is a no-op.
 pub fn set_summary_message(message: impl Into<std::borrow::Cow<'static, str>>) {
     summary::record(|s| s.set_message(message));
 }
--- a/bin/router/src/lib.rs
+++ b/bin/router/src/lib.rs
@@ -191,7 +191,7 @@ async fn graphql_endpoint_handler(
         .capture_request(&request);
 
     let started_at = std::time::Instant::now();
-    let (mut response, summary_guard) = async {
+    let (response_mode, mut response, summary_guard) = async {
         let summary_guard = summary::SummaryOnDrop::new(started_at);
         debug!(
             target: targets::HTTP_SERVER,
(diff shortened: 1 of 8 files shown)


lib/internal/src/telemetry/logging/summary.rs
```diff
@@ -94,6 +96,11 @@ impl RequestSummary {
         }
     }
 
+    /// Overrides the summary log line's message. First call wins; later calls are no-ops.
```
enisdenjo, line 99: I had to read all of the code to understand why are other call no-ops - maybe help future viewers by explaining why right here in the comment?

bin/router/src/lib.rs
```diff
@@ -221,16 +231,11 @@ async fn graphql_endpoint_handler(
...
     }
     .await;
 
-    // for streamed responses the summary must be emitted when the stream ends, not now
-    // we do that by attaching the summary guard to the response body, so it will be emitted
-    // when the stream terminates (or the client disconnects)
-    if response_mode.can_stream() {
-        response = summary_guard.attach_to_response(response);
-    }
+    response = summary_guard.attach_to_response(response);
```
dotansimha: We need to attach all kind of requests, not just streams. This also makes the `duration_ms` more accurate, because it `emit` when the request is really full dropped.

enisdenjo (approved):
one nit otherwise great

bin/router/src/lib.rs
```diff
@@ -178,7 +188,7 @@ async fn graphql_endpoint_handler(
         .capture_request(&request);
 
     let started_at = std::time::Instant::now();
-    let (response_mode, mut response, summary_guard) = async {
+    let (mut response, summary_guard) = async {
```
dotansimha: we just no longer needs `respose_mode` here at all

bin/router/src/lib.rs
```diff
@@ -143,13 +143,23 @@ pub fn set_summary_attribute(key: impl Into<String>, value: impl Into<sonic_rs::
...
 /// Lets plugins attach a custom correlation to every log line of the current request (not
 /// just the summary), e.g. a tenant or project id extracted from the URL. Setting the same
 /// key again overwrites the previous value. A no-op outside a request.
 pub fn set_log_correlation(key: impl Into<String>, value: impl std::fmt::Display) {
     request_id::set_correlation(key, value);
 }
 
+/// Lets plugins override the request summary log line's message. The first call wins;
+/// later calls for the same request are no-ops.
+pub fn set_summary_message(message: impl Into<std::borrow::Cow<'static, str>>) {
```
dotansimha, line 162: the actual front-facing "api" for plugin system

lib/internal/src/telemetry/logging/summary.rs
```diff
@@ -242,10 +254,15 @@ impl Drop for SummaryOnDrop {
...
 
-        match self.request_ids.take() {
-            Some(ids) => REQUEST_IDENTIFIERS.sync_scope(ids, || summary.emit()),
-            None => summary.emit(),
-        }
+        // Re-enter both task-locals before emitting: by now (especially for responses whose
+        // body outlives the original request future) they may no longer be ambiently scoped,
+        // but the formatters look up `custom`/`correlations` independently via their own
+        // `try_with` at format time, so both must be active for that lookup to succeed.
+        let request_ids = self.request_ids.take();
```
dotansimha, line 261: This fix the sync issue and ensures `on_end` of `on_http_request` has access to both the summary and log correlations

bin/router/src/lib.rs
enisdenjo: what if you have multiple plugins setting the summary message? the "why" does only the first call win will be questioned - like with https://github.com/graphql-hive/router/pull/1379#discussion_r3729851045 - maybe explain why here?

dotansimha: first call wins, otherwise we have to make the mechanism more complex with locks. I assumed users will implement their own plugin in order to override, so race conditions / conflicts are not expected.

lib/internal/src/telemetry/logging/summary.rs
dotansimha: fixed, thanks

PR #1392, opened 2026-08-11
feat(plugin-system): allow `on_http_request` to be `async` and return a `Future` 

### LINKED ISSUE #1391: Consider to make `on_http_req` plugin system hook `async`
This will allow developers to make HTTP calls without blocking the thread, or to implement custom async flows, before processing starts.

dotansimha:
Closes https://github.com/graphql-hive/router/issues/1391 

`RouterPlugin::on_http_request` now returns `OnHttpRequestHookFuture<'req>` instead of `OnHttpRequestHookResult<'req>` directly, so plugins can `await` inside - for example calling an upstream/external service, before the rest of the request pipeline runs.

The hook still runs on the same worker thread it started on and is never moved across threads, so the returned future does not need to be `Send`, by using `LocalBoxFuture`. 

```rust
// Before
fn on_http_request<'req>(
    &self,
    payload: OnHttpRequestHookPayload<'req>,
) -> OnHttpRequestHookResult<'req> {
    self.record_request(&payload);
    payload.proceed()
}

// After
fn on_http_request<'req>(
    &'req self,
    payload: OnHttpRequestHookPayload<'req>,
) -> OnHttpRequestHookFuture<'req> {
    Box::pin(async move {
        self.record_request(&payload);
        payload.proceed()
    })
}
```

If the body reads `self` inside the `async move` block, `&self` needs to become `&'req self` - the future now holds that borrow for its own lifetime `'req`, so the borrow has to be able to live at least that long.

### TODO

- [x] try `async_trait` with `?Send` (update: can't, it will force change the whole trait signature and all fns) 
- [x] try `OnHttpRequestHookResult` - ok this works
- [x] update all existing examples
- [x] add example with mock test
- [x] figure out e2e failures
- [ ] docs

### DIFF
--- a/bin/router/src/plugins/plugins_service.rs
+++ b/bin/router/src/plugins/plugins_service.rs
@@ -110,7 +110,7 @@ where
             let mut on_end_callbacks = Vec::with_capacity(plugins.len());
 
             for plugin in plugins.as_ref() {
-                let result = plugin.on_http_request(start_payload);
+                let result = plugin.on_http_request(start_payload).await;
                 start_payload = result.payload;
                 match result.control_flow {
                     StartControlFlow::Proceed => {
--- a/lib/executor/src/plugins/hooks/on_http_request.rs
+++ b/lib/executor/src/plugins/hooks/on_http_request.rs
@@ -1,3 +1,4 @@
+use futures::future::LocalBoxFuture;
 use ntex::{
     http::Response,
     web::{self, DefaultError, WebRequest},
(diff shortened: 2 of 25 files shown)


bin/router/src/plugins/plugins_service.rs
```diff
@@ -110,7 +110,7 @@ where
             let mut on_end_callbacks = Vec::with_capacity(plugins.len());
 
             for plugin in plugins.as_ref() {
-                let result = plugin.on_http_request(start_payload);
+                let result = plugin.on_http_request(start_payload).await;
```
dotansimha, line 113: we now `await` obviously

lib/executor/src/plugins/hooks/on_http_request.rs
```diff
@@ -111,6 +116,9 @@ pub type OnHttpRequestHookResult<'req> = StartHookResult<
     Response,
 >;
 
+/// A local (non-`Send`) boxed future - always polled on its own worker thread
+pub type OnHttpRequestHookFuture<'req> = LocalBoxFuture<'req, OnHttpRequestHookResult<'req>>;
```
dotansimha, line 120: This is the main change

PR #1397, opened 2026-08-12
fix(router): use route template as `http.route` in metrics and traces

praguevara:
HTTP server spans and metrics set `http.route` to the concrete request path. `http.route` is required to be low-cardinality: backends commonly derive span names and span metrics from it, and it is also a metric attribute. Whenever `http.graphql_endpoint` contains path parameters, every distinct path became its own span name and its own metric time series.

Report the configured GraphQL route template instead, which is the pattern the endpoint is registered with. The concrete path stays on `url.path`/`url.full`, which are not constrained to be low-cardinality. Persisted-document subpaths (`/graphql/sha256:<hash>`) collapse to the same template rather than contributing one route value per document.

For the default `/graphql` endpoint the template equals the request path, so the reported value is unchanged.

### DIFF
--- a/.changeset/fix_httproute_lable_value_in_metrics.md
+++ b/.changeset/fix_httproute_lable_value_in_metrics.md
@@ -1,0 +1,13 @@
+---
+hive-router-internal: patch
+hive-router: patch
+hive-router-plan-executor: patch
+---
+
+# Fix `http.route` lable value in metrics
+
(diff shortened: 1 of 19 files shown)


dotansimha (approved):
I've added some e2e tests, and changed the way the route is injected. Also did some internal refactor to `HiveRouterConfig` in order to use it as static instead of cloning.

PR #1434, opened 2026-08-21
feat(http): allow overriding the HTTP server keep-alive timeout, upgrade `ntex` to latest

Alastair-Spencer:
In addition, replaces https://github.com/graphql-hive/router/pull/1408

## Problem

The router never calls ntex's `set_keepalive`, so inbound HTTP/1 keep-alive is ntex's hardcoded 5 second default. There is no Hive Router setting to change it.

Behind a reverse proxy whose idle timeout is longer than 5 seconds, the router closes the socket first. The proxy then reuses a connection the server has already dropped, which fails the next request on that socket. This is independent of `traffic_shaping.all.pool_idle_timeout`, which only covers outbound subgraph connections.

Related: https://github.com/graphql-hive/router/pull/1433 (graceful shutdown timeout). Same class of gap: an ntex HTTP-server default that Hive did not expose.

## Change

Adds `traffic_shaping.router.keep_alive`, a human-readable duration passed to ntex `HttpServiceConfig::set_keepalive` (not `set_keepalive_timeout`, which is a second setter on the same field):

```yaml
traffic_shaping:
  router:
    keep_alive: 80s
```

- Defaults to 5 seconds, so behaviour is unchanged for configurations that omit it.
- Applied to both the main server and a separately bound callback server.
- Overridable via `ROUTER_HTTP_KEEP_ALIVE`.
- Uses the repository's `humantime_serde` plus `schemars(with = "String")` convention.
- `0s` disables keep-alive (`KeepAlive::Disabled`).

When the router sits behind a reverse proxy, set this above that proxy's idle timeout so the proxy closes first.

ntex takes whole seconds, so partial seconds round up and oversized values clamp rather than wrap.

## Tests

Unit tests cover the 5 second default, parsing `80s`, the serialised duration-string shape, both environment override paths including precedence over a config file value, and the `Duration` to ntex `KeepAlive` conversion boundaries.

```
cargo test -p hive-router --lib -- keep_alive to_ntex_keep_alive config::
```

### DIFF
--- a/.changeset/configurable-http-keep-alive.md
+++ b/.changeset/configurable-http-keep-alive.md
@@ -6,7 +6,7 @@ Allow overriding the HTTP server keep-alive timeout
 
 Adds a new `http.keep_alive` configuration option (and `ROUTER_HTTP_KEEP_ALIVE` environment variable) to control how long the HTTP server waits for a follow-up request on an idle keep-alive connection before closing it.
 
-This was previously fixed at ntex's 5 second default, with no Hive Router setting to change it. Behind a reverse proxy whose idle timeout is longer than 5 seconds, the router would close the socket first. The proxy then reused a connection the server had already dropped. The default stays at 5 seconds, so existing behaviour is unchanged.
+This was previously fixed at ntex's 5 second default, with no Hive Router setting to change it. Behind a reverse proxy whose idle timeout is longer than 5 seconds, the router would close the socket first. The proxy then reused a connection the server had already dropped.
 
 ```yaml
 http:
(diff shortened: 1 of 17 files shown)


dotansimha:
Thanks @Alastair-Spencer ! 
I'm going to take over this PR, to also upgrade `ntex`'s patch version since it includes a crucial fix for `keep-alive` handling. 
Will figure this out and also add e2e to confirm this behavior.

dotansimha (approved):

dotansimha:
I've also moved this to `traffic_shaping.router.` and upgraded to latest `ntex`.

PR #1445, opened 2026-08-25
fix(http): restore graceful SIGTERM drain after `ntex-server` 3.11

Alastair-Spencer:
## Problem

`SIGTERM` force-drops in-flight requests, and `http.shutdown_timeout` (#1433) has no effect.

`ntex-server` 3.11.0 ([ntex-rs/ntex#964](https://github.com/ntex-rs/ntex/pull/964), `0f0fb48`) made graceful signal handling opt-in:

- added `WorkerPool.graceful_shutdown`, defaulting to `false`
- changed `Signal::Term` from `state.stop(true, ...)` to `state.stop(cfg.graceful_shutdown, ...)`
- exposed `.graceful_shutdown()` on `ntex_server::ServerBuilder` only

`ntex::web::HttpServer` has no `graceful_shutdown()` passthrough, so the router cannot opt in. With the flag `false`, `stop()` takes the non-graceful branch and calls `worker.stop(Millis::ZERO)`, a forced shutdown that ignores `shutdown_timeout` entirely.

This regressed when #1434 upgraded `ntex` to pull in `ntex-server` 3.11.1. It also contradicts the documented contract. [ntex's server guide](https://ntex.rs/docs/server) still states "SIGTERM - Graceful shutdown workers", and `HttpServer::shutdown_timeout` is still documented as the graceful drain window.

Measured against a subgraph that answers at T+6s, with `SIGTERM` sent at T+2s:

| | in-flight request | process exit |
|---|---|---|
| `ntex-server` 3.10.4 | completes, `200` at T+6.05s | 4.4s after `SIGTERM` |
| `ntex-server` 3.11.1 | `curl (52) Empty reply from server` at T+2.39s | 0.56s after `SIGTERM` |

## Change

Hold `ntex-server` at `=3.10.4`, and `ntex-net` at `=3.14.1` and `ntex-rt` at `=3.16.1` alongside it.

All three have to move together. `ntex-server` 3.10.4 does not build against `ntex-rt` 3.17, which added a `Signal::Panic(PanicSource)` variant and dropped `Copy` from `Signal`; 3.10.4's signal match then fails with `E0004` (non-exhaustive patterns) and `E0507` (cannot move out of `*sig`). `ntex-net` 3.15.0 in turn requires `ntex-rt >=3.17.0`, so it comes back to 3.14.1, the last release that pairs with `ntex-rt` 3.16.1.

`ntex` itself stays on 3.12.3, which accepts `ntex-net ^3.14.0`, `ntex-rt ^3.16.1` and `ntex-server ^3.10.4`, so there is no wider `ntex` downgrade.

The cost is `ntex`'s new worker panic supervision (`ntex-rt` 3.17's panic handling and `Arbiter::on_close`, `ntex-net` 3.15's "do not unwind reactor panics"). The router does not opt into any of it today, so nothing in-tree regresses, but it is a real capability held back and I'd rather not sit on it long.

## Preferred fix

This is a workaround, not where I think this should land. `web::HttpServer::graceful_shutdown()` already exists upstream — it was added in `ntex` [4.0.0-beta.0](https://docs.rs/ntex/4.0.0-beta.0/ntex/web/struct.HttpServer.html). If that passthrough were backported to a 3.x release, all three pins here collapse back to a plain version bump and the panic supervision comes with it.

Two options from my side, happy either way:

- I raise the backport on `ntex` and we hold this PR until a 3.x release carries it.
- We take these pins now to stop the regression bleeding, and revert them when the backport lands.

Alternatively the router could stop using `web::HttpServer` and drive `ntex::server::build().graceful_shutdown()` directly, replicating what `HttpServer::bind`/`listen` do. That keeps every version current but is a much larger change to the startup path, and it seemed wrong to reach for it while a five-line upstream passthrough exists.

## Tests

End-to-end behaviour was verified with the same harness as the table above: `SIGTERM` at T+2s during a 6s request returns `200` with the complete body, and the process exits about 4.4s after the signal.

I did not add an e2e test for this. As noted in #1433 I wasn't sure the existing testkit can deterministically sequence a stop with a request in flight, but happy to add one if you can point me at the right approach.

### DIFF
--- a/.changeset/restore-graceful-sigterm-drain.md
+++ b/.changeset/restore-graceful-sigterm-drain.md
@@ -1,0 +1,12 @@
+---
+hive-router: patch
+---
+
+Restore graceful HTTP server shutdown after the ntex-server 3.11 upgrade.
+
+ntex-server 3.11 made graceful signal handling opt-in, while
+`ntex::web::HttpServer` does not expose the new setting. This caused
+`SIGTERM` to force-drop workers immediately and bypass the configured
+`http.shutdown_timeout`. Pin `ntex-server` to 3.10.4 so SIGTERM drains
+in-flight requests until ntex provides a supported opt-in on the public
+HTTP server API.
--- a/.changeset/restore-graceful-sigterm-drain.md
+++ b/.changeset/restore-graceful-sigterm-drain.md
@@ -4,9 +4,11 @@ hive-router: patch
 
 Restore graceful HTTP server shutdown after the ntex-server 3.11 upgrade.
 
-ntex-server 3.11 made graceful signal handling opt-in, while
-`ntex::web::HttpServer` does not expose the new setting. This caused
-`SIGTERM` to force-drop workers immediately and bypass the configured
-`http.shutdown_timeout`. Pin `ntex-server` to 3.10.4 so SIGTERM drains
-in-flight requests until ntex provides a supported opt-in on the public
-HTTP server API.
+ntex-server 3.11 made graceful signal handling opt-in and defaults it to false,
+while `ntex::web::HttpServer` exposes no way to opt in. This caused `SIGTERM` to
+force-drop workers immediately and bypass the configured `http.shutdown_timeout`.
+
+Hold `ntex-server` at 3.10.4, together with `ntex-net` 3.14.1 and `ntex-rt`
+3.16.1, so that `SIGTERM` drains in-flight requests again. All three move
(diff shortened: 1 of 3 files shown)


dotansimha (approved):

dotansimha (approved):

dotansimha:
Fully fixed with latest ntex in hive-router 0.2.4

Issue #1475, opened 2026-09-02
jwt: allow custom headers on remote jwks_providers (e.g. Host, for in-cluster IdPs that route by Host header)


peter-svensson:
### Problem

`jwks_providers` with `source: remote` fetches the key set with a bare GET and no way to set request headers:

```rust
// bin/router/src/config/jwt_auth.rs
Remote {
    url: String,
    polling_interval: Option<Duration>,
    prefetch: Option<bool>,
},
```

```rust
// bin/router/src/jwt/jwks_manager.rs
let client = reqwest::Client::new();
let response_text = client.get(url).send().await?...
```

That makes it impossible to reach an identity provider that selects its tenant/instance from the **`Host` header** when the router addresses it by an internal address.

### Concrete case

Zitadel resolves which instance a request belongs to by matching the request `Host` against its registered instance domains. Requests arriving with any other Host get:

```
unable to set instance using origin ... Message=Instance not found
```

Deployed in Kubernetes, that leaves two options for a router running in the same cluster as the IdP:

1. Point `url` at the in-cluster Service (`http://idp.<ns>.svc.cluster.local/oauth/v2/keys`). The connection succeeds, but the `Host` is now the internal name, the IdP has no instance registered for it, and the fetch 404s.
2. Point `url` at the public URL. This works — and is what we do today — but every JWKS refresh leaves the cluster, traverses the external load balancer and ingress, and comes back to a pod that is one ClusterIP hop away. Token validation then depends on public DNS, the external LB, and a public TLS certificate.

Neither is good. Option 2 in particular turns an internal dependency into an external one on the auth hot path.

Other IdPs and multi-tenant setups behave the same way, and the pattern is not unusual: any virtual-host-routed service behind a Kubernetes Service has this shape.

### Request

Allow headers on the remote provider:

```yaml
jwt:
  jwks_providers:
    - source: remote
      url: http://idp.identity.svc.cluster.local/oauth/v2/keys
      headers:
        Host: auth.example.com
      polling_interval: "15m"
      prefetch: true
```

A general `headers` map is the most useful shape — it also covers JWKS endpoints behind an API gateway that expects an API key or a custom auth header, which is a separate ask this would satisfy at the same time.

Note `reqwest` needs `Host` set specifically: a plain header insert is respected on the request, so `RequestBuilder::header(HOST, ...)` works, but it is worth an explicit test since some clients override it from the URL authority.

### Precedent elsewhere in the config

[...]

PR #1476, opened 2026-09-02
feat(jwt): add `headers` to remote JWKS providers

peter-svensson:
Related #1475. Implements the shape agreed in [that thread](https://github.com/graphql-hive/router/issues/1475#issuecomment-5508595223).

```yaml
jwt:
  jwks_providers:
    - source: remote
      url: http://idp.identity.svc.cluster.local/oauth/v2/keys
      headers:
        Host: auth.example.com
      polling_interval: "15m"
```

## Why

Two setups that previously could not be expressed.

**An IdP that resolves its tenant from the `Host` header.** Addressing it by an internal URL fails: the connection succeeds, but the provider sees the internal authority, finds no tenant registered for it, and the fetch errors. Zitadel matches the request `Host` against its registered instance domains and otherwise answers `Instance not found`. Without this, the only option is fetching keys over the public URL — so in a Kubernetes deployment every refresh leaves the cluster, traverses the external load balancer, and comes back to a pod one ClusterIP hop away, putting public DNS, the LB, and a public certificate on the token-validation path.

**A JWKS endpoint behind a gateway** expecting an API key or custom auth header, now reachable without a proxy in front of the router.

## Implementation

Follows `preflight_response_headers` in `cors.rs`: `http_serde::header_map` for (de)serialization, `schemars(with = "HashMap<String, String>")` for the JSON schema. No new dependencies.

The field defaults to empty and `.headers()` with an empty map is a no-op, so existing configurations are byte-for-byte unaffected.

## On `Host` specifically

Worth calling out, since it is the motivating case and could plausibly not work: `reqwest` might have derived `Host` from the URL authority and ignored the override. **It does not.** Verified against a raw listener before writing the tests:

```
GET /jwks.json HTTP/1.1
host: auth.example.com      <- the override, not 127.0.0.1:<port>
x-api-key: secret
accept: */*
```

The override reaches the wire while the request still goes to the URL's address — exactly the behaviour this feature needs. That assertion is now a test rather than a note, so a future `reqwest` change that breaks it fails CI instead of silently disabling the feature.

## Tests

`cargo test -p hive-router --lib config::jwt_auth jwt::jwks_manager` — 8 pass.

- **config**: headers absent → empty; headers parsed and normalized; an invalid header name (`"Invalid Header"`) is rejected at parse time
- **fetch** (`mockito`, already a dev-dependency): configured headers reach the server, asserted with `.match_header("host", "auth.example.com")`; and the no-headers path is unchanged

`cargo fmt --check` clean. `cargo clippy` reports nothing on the two changed files — note it reports 38 pre-existing errors on unmodified `main` for me (`websocket_server.rs` and the benches), presumably a newer clippy than CI pins; I have left those alone as out of scope.

## Not included

The docs site lives in another repo, so the `jwks_providers` reference there will need the new field. Happy to open that PR too — point me at the right place.

### DIFF
--- a/.changeset/jwks_remote_headers.md
+++ b/.changeset/jwks_remote_headers.md
@@ -1,0 +1,27 @@
+---
+hive-router: minor
+---
+
+# Add `headers` to remote JWKS providers
+
+`jwt.jwks_providers` with `source: remote` now accepts a `headers` map, applied to the JWKS fetch.
+
+```yaml
+jwt:
+  jwks_providers:
+    - source: remote
+      url: http://idp.identity.svc.cluster.local/oauth/v2/keys
+      headers:
+        Host: auth.example.com
+      polling_interval: "15m"
+```
+
+This makes two previously unreachable setups work.
+
+An identity provider that resolves its tenant or instance from the `Host` header cannot be addressed by an internal URL: the connection succeeds, but the provider sees the internal authority, finds no tenant registered for it, and the fetch fails. Zitadel is one such provider — it matches the request `Host` against its registered instance domains and otherwise answers `Instance not found`. Setting `Host` here lets the router reach the provider over a cluster-internal address while still identifying the tenant, instead of routing every key refresh out through a public load balancer and back.
+
+A JWKS endpoint behind a gateway that expects an API key or a custom auth header is now also reachable, without a proxy in front of the router.
+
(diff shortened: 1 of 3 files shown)


dotansimha:
Thanks @peter-svensson ! I'm going to review and get this in for the upcoming release

dotansimha:
Updated the docs in https://github.com/the-guild-org/website/pull/1944 
Available in 0.2.7 

Thanks @peter-svensson !

Issue #315, opened 2025-08-07
RFC: HTTP Compression 


dotansimha:
If you are looking for `Traffic Shaping` categorization, see the note [here](https://github.com/graphql-hive/router/issues/317)

# Introduction
HTTP Compression is a technique in HTTP connections to pass over the data compressed in a specific algorithm or algorithms.
There are two headers used for the negotiation between the client and the server.

`Accept-Encoding` which contains the list of algorithms that are supported by the peer(client/server)
`Content-Encoding` which contains the list of algorithms(usually a single one) applied to the data. So the consumer peer should decompress the body using that or those algorithms before processing it.

Here are some problems we need to keep on mind before continue;
When an initial request is sent to a server, you don't know which encoding types are supported by the server. So the algorithm you set in `Content-Encoding` in your request headers might not be supported. So you have to be careful on that.
An initial request should always be sent uncompressed, and `Accept-Encoding` should be waited to know which algorithms are supported.
Also while sending the request, you should set `Accept-Encoding` with a list of algorithms you support such as `gzip, br, deflate`. The order is important because the peer will pick the first compliant one from left to right.

# Gateway's use case
In a GraphQL gateway, there are 4 different ways of HTTP communication when the compression can take place.

## From the gateway to the client
When the client sends a request to the gateway, it can specify the compression algorithm that it supports using the Accept-Encoding header. Then the gateway can compress the response using the specified algorithm before sending it back to the client with the Content-Encoding header, so that the client can decompress it.

```js
const res = await fetch('http://localhost:4000/graphql', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
    'Accept-Encoding': 'gzip'
  },
  body: JSON.stringify({
    query: `
      query {
        hello
      }
    `
  })
})
console.assert(res.headers.get('Content-Encoding') === 'gzip', 'Response is compressed')
```

When you send this request, you tell the server that the client supports `gzip`. So the server CAN send it as gzip. But you should always check the header because it is not guaranteed that the server sends it `gzip`.

## From the client to the gateway
When the client sends a request to the gateway, it can be compressed and the algorithm MUST be defined in `Content-Encoding` so the gateway respects it and decompresses it before processing it. But it is risky because the gateway might not support that algorithm and return an HTTP error because there is no way to check if the server supports compression from the consumer side. Before configuring this feature on the client side, make sure that the gateway supports the compression algorithm that the client supports.
[...]

PR #1477, opened 2026-09-02
feat(router): `traffic_shaping.router.compression` 

dotansimha:
Related https://github.com/graphql-hive/router/issues/315

HTTP compression between the client and the router. Supports gzip, deflate, br, and zstd in both directions.

Default config:

```yaml
traffic_shaping:
  router:
    compression:
      response:
        enabled: true
        algorithms:
          - kind: gzip
          - kind: zstd
            level: 3
          - kind: br
            quality: 5
          - kind: deflate
        min_size: 1KiB
      request:
        enabled: true
        algorithms: [gzip, zstd, br, deflate]
```

## Response Compression

- Negotiated against the client's `Accept-Encoding` (`q`-value aware, `*`-aware), restricted to a configurable algorithms allow-list — br/zstd carry their own inline tuning (quality/level) co-located with their entry in that list.
-  Responses below `min_size` (default 1KiB) are left uncompressed; Also streaming responses (SSE/multipart subscriptions) are never compressed.
- gzip/deflate reuse `ntex`'s built-in encoder; br/zstd (unsupported by `ntex`) compress on the blocking pool. 
- Best-effort: a compression failure falls back to the original uncompressed body rather than erroring.
- Sets `Vary: Accept-Encoding` on every response while compression is enabled, appending to any existing Vary (e.g. from CORS) rather than clobbering it — needed for downstream cache layer

## Request Decompression

- Applied based on the client's `Content-Encoding`, validated against `compression.request` (enabled + algorithm allow-list). Unsupported, disabled, or stacked/multi-valued (`gzip, br`) encodings are rejected with HTTP 415.
- Decompresses incrementally as chunks arrive, checking the decompressed size against `limits.max_request_body_size` after each chunk — this is what actually guards against decompression bombs.
- A body that doesn't actually match its claimed `Content-Encoding` fails closed with HTTP 400, for all four algorithms.

## Other stuff 
- Fixed a pre-existing bug: `content-encoding` was missing from the router's hop-by-hop header deny-list. Now it's more crucial so fixes.

### DIFF
--- a/bin/router/src/config/traffic_shaping.rs
+++ b/bin/router/src/config/traffic_shaping.rs
@@ -1,6 +1,7 @@
 use std::{borrow::Cow, collections::HashMap, fmt, time::Duration};
 
 use http::StatusCode;
+use human_size::Size;
 use schemars::{json_schema, JsonSchema, Schema, SchemaGenerator};
 use serde::{de, Deserialize, Serialize};
 
@@ -492,6 +493,14 @@ pub struct TrafficShapingRouterConfig {
     )]
     #[schemars(with = "String")]
     pub keep_alive: Duration,
+
+    /// HTTP compression between the client and the router: compressing responses sent to the
+    /// client, and decompressing requests received from the client.
+    ///
+    /// This is unrelated to `traffic_shaping.all`/`traffic_shaping.subgraphs`, which control
+    /// compression between the router and subgraphs.
+    #[serde(default)]
+    pub compression: TrafficShapingRouterCompressionConfig,
 }
 
 #[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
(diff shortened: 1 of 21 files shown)


e2e/src/compression.rs
```diff
@@ -0,0 +1,1174 @@
...
+        }
+    }
+
+    // confirm something similar to Apollo Router's CVE-2024-28101
+    //
+    // the request body size limit was only enforced *after* the
+    // compressed body had been fully decompressed into memory, so a small, highly
+    // compressible payload could exhaust memory before the check ever ran
+    #[ntex::test]
+    async fn should_reject_single_chunk_decompression_bomb_promptly() {
```
kamilkisiela, line 702: it rejects the request based on `Content-Length` here, but what if the `Content-Length` is below the limit, but decompresses to 2KiB?

yanns:
> br/zstd (unsupported by ntex) compress on the blocking pool

It'd be nice to add native support to ntex for those algos.

e2e/src/compression.rs
dotansimha: yeah, the `Content-Length` one was helping the test a bit, it was taking the fast path to the error. 

We do have a more in-depth check based on the actual data, so i adjust the test to passthrough the content-length limit and confirmed that it fails because the (real) body size is too big https://github.com/graphql-hive/router/pull/1477/commits/40549e68ff02962725b14e6e69eed0e143712bab

dotansimha:
> zstd

is not supported by ntex at the moment, created an issue: https://github.com/ntex-rs/ntex/issues/1006

> br

Yes, not sure why, but I saw many issues and PRs in `ntex` and it was eventually dropped. I think it's because `brotli2` crate was unmaintained.

bin/router/src/http_utils/body.rs
```diff
@@ -129,24 +137,115 @@ pub async fn read_body_stream<R: RequestLike>(
...
+                    .map_err(|_| std::io::Error::other("brotli stream did not finish cleanly"))?
+            }
+            Self::Zstd(mut d) => {
+                d.flush()?;
+                d.into_inner()
+            }
+        };
+        Ok(Bytes::from(bytes))
+    }
 }
```
enisdenjo, line 318: I love rust

bin/router/src/http_utils/body.rs
```diff
@@ -129,24 +137,115 @@ pub async fn read_body_stream<R: RequestLike>(
...
-        // limit max size of in-memory payload
-        if chunk.len() > max_size.saturating_sub(body.len()) {
-            write_request_body_size(req, (body.len() + chunk.len()) as u64);
+        decoder
+            .write_all(&chunk)
+            .map_err(|err| ReadBodyStreamError::DecompressionFailed(err.to_string()))?;
+
+        if decoder.len() > max_size {
+            write_request_body_size(req, decoder.len() as u64);
+            drain_body_stream(&mut body_stream).await;
```
enisdenjo: decompressed limit is checked only after `write_all` returns. a single compressed chunk can therefore inflate far beyond `max_size` into the backing `Vec` before line 170 checks it. there is decompression size limit test, but it only confirms a timely response, yet still permits the full 25 MB allocation.

lets instead try a size-limited sink that returns an error as soon as output exceeds `max_size`?

bin/router/src/pipeline/response_compression.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    let original = match drain_body(body).await {
+        Ok(bytes) => bytes,
+        Err(err) => {
+            error!(
+                target: targets::HTTP_SERVER,
+                error = %err,
+                "failed to read response body while negotiating compression, sending empty body"
+            );
+            return web::WebResponse::new(head.set_body(Body::Empty), request);
```
enisdenjo: hmm this converts a response body read failure into a successful response with an empty body, silently corrupting the response into a invalid graphql over http response.

it also conflicts with the function documentation saying the original response is returned on failure

```rs
/// Compression is best-effort: if it fails, we log the failure/error and return the original response
```

bin/router/src/pipeline/response_compression.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        Ok(response)
+    }
+}
+
+/// Picks the algorithm from `algorithms` that best matches the client's `Accept-Encoding` preference
+fn negotiate<'cfg>(
+    accept_encoding: Option<&HeaderValue>,
+    algorithms: &'cfg [ResponseCompressionAlgorithmConfig],
+) -> Option<&'cfg ResponseCompressionAlgorithmConfig> {
```
enisdenjo, line 125: `Accept-Encoding` headers  belong to the `Accept` family, they're structuraly the same but negotiate different things. I think you can use the `header_accept` crate (that we use for content types) to negotiate and calculate the `q` weights without a hassle

bin/router/src/config/traffic_shaping.rs
```diff
@@ -585,10 +594,226 @@ impl Default for TrafficShapingRouterConfig {
...
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, Copy, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct BrotliCompressionConfig {
+    /// Brotli quality level, from `0` (fastest, worst ratio) to `11` (slowest, best ratio).
+    ///
+    /// Defaults well below the maximum: quality `11` is slow enough to become a bottleneck
+    /// for dynamically generated GraphQL responses, where CPU time usually matters more
+    /// than squeezing out the last few percent of size.
+    #[serde(default = "default_brotli_quality")]
+    pub quality: u8,
```
enisdenjo, line 765: should we validate 0..11?

bin/router/src/config/traffic_shaping.rs
```diff
@@ -585,10 +594,226 @@ impl Default for TrafficShapingRouterConfig {
...
+
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone, Copy, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct ZstdCompressionConfig {
+    /// Zstandard compression level, from `1` (fastest, worst ratio) to `22` (slowest, best ratio).
+    ///
+    /// Defaults to zstd's own standard default, which already gives a good balance of
+    /// speed and ratio for dynamically generated GraphQL responses.
+    #[serde(default = "default_zstd_level")]
+    pub level: i32,
```
enisdenjo, line 801: also, as with brotli - should we validate 1..22?

enisdenjo (changes requested):
a few remarks. I'd add more tests around request size limiting and decompression bombs, just to make sure our security is up to a standard.

bin/router/src/config/traffic_shaping.rs
dotansimha: yes! done in https://github.com/graphql-hive/router/pull/1477/commits/5bc778b4c7da2fabc6c65a55140ce6616d8a7609

bin/router/src/config/traffic_shaping.rs
dotansimha: yes! done in https://github.com/graphql-hive/router/pull/1477/commits/5bc778b4c7da2fabc6c65a55140ce6616d8a7609

bin/router/src/pipeline/response_compression.rs
dotansimha: I thought the same 😆 but it doesn't, it expects some hardcoded strings so it fails to parse the `q` (and even the basic value of the algo name)

dotansimha: I also found this https://crates.io/crates/accept-encoding but it's outdated and depends on a super old version of `http` (last release was 8y ago)

bin/router/src/pipeline/response_compression.rs
dotansimha: Yes. You are correct on this one. 
It should be 500, as mid-flight-read-error is not something we can recover from. In that specific case, we can't really fall back to the original response, since we don't really have one (we failed to read it...)

dotansimha: I change it to 500 and improved the log/comments around it in https://github.com/graphql-hive/router/pull/1477/commits/891515444d2ce26bcebcc817015db48e5a585548

bin/router/src/http_utils/body.rs
dotansimha: yes. good point. Done in https://github.com/graphql-hive/router/pull/1477/commits/c2ca9184c977ea67fde74757e48bb3a90f45eac1

bin/router/src/pipeline/response_compression.rs
enisdenjo: 🤦‍♂️

bin/router/src/http_utils/body.rs
```diff
@@ -129,24 +137,184 @@ pub async fn read_body_stream<R: RequestLike>(
...
+        } else {
+            BytesMut::new()
+        };
+
+        while let Some(chunk) = body_stream.try_next().await? {
+            // limit max size of in-memory payload
+            if chunk.len() > max_size.saturating_sub(body.len()) {
+                write_request_body_size(req, (body.len() + chunk.len()) as u64);
+                return Err(ReadBodyStreamError::PayloadTooLargeBodyStream);
+            }
```
enisdenjo, line 152: excellent

enisdenjo (approved):

Issue #1488, opened 2026-09-04
Missing `processVariables` for Hive usage reporting


dotansimha:
Same as the JS implementation: https://the-guild.dev/graphql/hive/docs/api-reference/client#processvariables

PR #1503, opened 2026-09-07
feat(console-sdk, router, apollo-router-fork): `process_variables` for usage-reporting 

dotansimha:
## Add `process_variables` to Hive Console usage reporting

Related https://github.com/graphql-hive/router/issues/1488 

The router can now inspect the actual `variables` payload of each operation and report only the schema coordinates that were genuinely used, instead of every coordinate the operation's input variables *could* represent. Enable it with `telemetry.hive.usage_reporting.process_variables: true` (default: `false`).

No content of the variables object is ever sent to Hive Console — only the derived schema coordinates.

Added this both to Hive Router and Apollo-Router fork 

### Improve the usage-report operation cache key

`OperationProcessor` now keys its cache with an xxh3 hash of the operation body and, when `process_variables` is enabled, the shape of the variables payload: variable names, object keys, nesting and list lengths, never values.

### Stop reporting directive arguments as schema coordinates in usage reports

Arguments of directives applied to fields, such as `users @include(if: $flag)`, were collected as field-argument coordinates (`Query.users.if`).

Directive arguments are not schema coordinates, so usage reports now skip them as well.

(this aligns to the JS SDK)


lib/hive-console-sdk/src/agent/utils.rs
```diff
@@ -290,6 +318,15 @@ impl<'a> OperationVisitor<'a, SchemaCoordinatesContext<'a>> for SchemaCoordinate
...
         let type_name = Self::resolve_type_name(&var.var_type);
 
+        if self.process_variables {
+            ctx.plan_entries.push(VariablePlanEntry {
+                var_name: var.name.clone(),
+                type_name: type_name.to_string(),
+            });
+
+            return;
+        }
```
enisdenjo, line 328: this returns before processing the variable default. if `$filter: Filter = { active: true }`, extraction reports the bare type and misses `Filter.active!` and the argument `!` marker of course

lib/hive-console-sdk/src/agent/variables_plan.rs
```diff
@@ -0,0 +1,217 @@
...
+pub struct VariablesExtractionPlan {
+    pub entries: Vec<VariablePlanEntry>,
+    pub markers: Vec<VariableUsageMarker>,
+}
+
+impl VariablesExtractionPlan {
+    pub fn is_empty(&self) -> bool {
+        self.entries.is_empty() && self.markers.is_empty()
+    }
+
```
enisdenjo, line 43: unused

```suggestion
```

apollo-router-workspace/bin/router/src/usage.rs
```diff
@@ -290,26 +371,38 @@ impl Plugin for UsagePlugin {
...
+                                        .filter_map(|(k, v)| {
+                                            v.to_str()
+                                                .ok()
+                                                .map(|value| (k.to_string(), value.to_string()))
+                                        })
+                                        .collect(),
+                                };
+
+                                let input_variable_coordinates =
+                                    Self::collect_input_variable_coordinates(&agent, &schema, req);
```
enisdenjo, line 396: with a 1% sample rate, variable extraction still runs for 100% of requests while 99% of the results are discarded. maybe return `dropped` from `populate_context` and then

```suggestion
                                let input_variable_coordinates = if dropped {
                                    Vec::new()
                                } else {
                                    Self::collect_input_variable_coordinates(&agent, &schema, req)
                                };
```

lib/hive-console-sdk/src/agent/usage_agent.rs
```diff
@@ -167,6 +188,27 @@ pub trait UsageAgentExt {
         execution_report: ExecutionReport,
         request: Option<RequestDetails>,
     ) -> Result<(), AgentError>;
+
+    /// Whether granular `processVariables` reporting is enabled for this agent.
+    fn process_variables_enabled(&self) -> bool;
```
enisdenjo, line 193: `extract_variable_coordinates` already returns an empty vector when processing is disabled, so the Apollo guard (in `collect_input_variable_coordinates`) repeats the same check. 

maybe remove `process_variables_enabled` altogether and call extraction directly?

lib/hive-console-sdk/src/agent/variables_plan.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn visit_items(self, visit: &mut dyn FnMut(Self)) {
+        if let Some(items) = self.0.as_array() {
+            for value in items {
+                visit(SerdeJsonView(value));
+            }
+        }
+    }
+}
```
enisdenjo, line 158: this newtype and `SonicJsonView` below only wrap references and forward operations. could we implement it directly for `&serde_json::Value` and `&sonic_rs::Value`? `VariablesPayload` can then return references directly, removing the wrapper structs and all .map(...JsonView) calls

apollo-router-workspace/bin/router/src/usage.rs
```diff
@@ -179,6 +187,75 @@ impl UsagePlugin {
...
+    where
+        Self: 'a;
+
+    fn get(&self, key: &str) -> Option<Self::Value<'_>> {
+        self.0.get(key).map(NativeJsonView)
+    }
+}
+
+#[derive(Clone, Copy)]
+struct NativeJsonView<'a>(&'a serde_json_bytes::Value);
```
enisdenjo, line 229: same as the other `*JsonView`. `JsonView` can be implemented directly for `&serde_json_bytes::Value`. keeping the necessary map adapter while removing the value wrapper. what do oyu think?

enisdenjo (changes requested):

lib/hive-console-sdk/src/agent/utils.rs
dotansimha: Fixed, see new impl

kamilkisiela (approved):

Issue #1494, opened 2026-09-04
Quiet upstream subscription streams leak: subgraph connection survives its last consumer disconnecting


maxbol:
Disclosure: This issue, and the technical investigation behind it, is co-authored with an LLM (Claude Fable 5)

## Description

When the last consumer of an upstream subscription stream disconnects, the router only tears down the subgraph connection once the **next subscription event** arrives on that stream. For a stream that emits no events (or emits rarely), the connection stays open indefinitely — until an event happens to arrive, the supergraph reloads, or a process restarts.

Note the scope: receiver liveness is per upstream stream (one broadcast channel per `ActiveSubscriptions::register`), not per subscription field. With request deduplication disabled — the default — every client subscribe opens its own dedicated upstream stream with a single receiver, so **every individual client disconnect orphans one upstream connection**, regardless of how many other clients are subscribed. With dedupe enabled the leak surfaces when the last client sharing a fingerprint leaves; since auth headers are part of the fingerprint in any per-user-auth setup, that is still effectively per-client.

Observed on `hive-router 0.0.87`; the relevant code is unchanged on `main` as of 2026-09-04.

## Mechanism

Receiver liveness is detected only via a failed send, at two independent layers:

1. **The event pump** (`bin/router/src/pipeline/mod.rs`, `execute_planned_request`, `QueryPlanExecutionResult::Stream` branch): the spawned pump loop selects on `body_stream.next()` and `supergraph.snapshot.retired()`. The only way it learns that all broadcast receivers are gone is `producer_handle.send(...)` returning `false` — which requires an upstream event.

2. **The subscription buffer drainer** (`lib/executor/src/executors/subscription_buffer.rs`, `drain_into`): the drainer owns the actual subgraph response body and forwards parsed events into a bounded mpsc channel. It only learns the consumer is gone when `try_send` returns `Closed` — which, again, requires an upstream event.

Crucially, SSE keep-alive pings/comments never reach either send: the SSE parser (`lib/executor/src/executors/sse.rs`, `parse_to_stream`) filters everything that is not an `event: next` / `event: complete` (`_ => { // ping }`). So a subgraph dutifully sending keep-alives keeps the TCP connection healthy while neither layer ever gets a chance to notice the departed consumers.

The WebSocket-server side cleans up correctly on client disconnect (`state.subscriptions.clear()` on shutdown drops the per-client forward tasks and their broadcast receivers) — the leak is purely on the subgraph side of the fan-out.

[...]

PR #1504, opened 2026-09-07
Close quiet upstream subscriptions after client disconnects

enisdenjo:
Related #1494

The router now closes an upstream subscription as soon as its final downstream client disconnects, even when the upstream emits no further events.


.changeset/close_quiet_upstream_subs.md
```diff
@@ -0,0 +1,10 @@
+---
+hive-router: patch
+node-addon: patch
```
dotansimha: I think `node-addon` doesn't really need a release here?

PR #1521, opened 2026-09-09
feat(router): native `datadog` tracing

enisdenjo:
Docs https://github.com/the-guild-org/website/pull/1967

Hive Router can export traces through generic OpenTelemetry exporters or through native Datadog tracing. Generic OTLP sends OpenTelemetry payloads to an OTLP collector or endpoint, usually on port `4317`. Native Datadog tracing uses the Datadog Rust tracer and sends traces to a Datadog Agent trace endpoint, normally on port `8126`.

Native tracing requires a reachable Datadog Agent while the Router is running. Sampling happens inside the Router and is not delegated to the Agent. Agent unavailability or temporary Agent errors do not block Router startup. Export errors are reported by Datadog logging.

## Configuration

```yaml
telemetry:
  tracing:
    collect:
      sampling: 0.01
    propagation:
      trace_context: true
    exporters:
      - kind: datadog
        enabled: true
        endpoint: http://datadog-agent.datadog-operator:8126
```

`enabled` defaults to `true`. To disable only native Datadog export, set this field to `false`. Do not use `DD_TRACE_ENABLED=false` for that purpose: because the Router uses one shared provider, that environment variable disables tracing for every OTLP, stdout, and Hive processor attached to the Datadog-backed provider.

`endpoint` is optional. If omitted, Datadog uses its native configuration, including `DD_TRACE_AGENT_URL`, `DD_AGENT_HOST`, and `DD_TRACE_AGENT_PORT`. A configured endpoint may contain a Router expression. The Router resolves the expression and validates the URL and Agent URL scheme during startup. Unresolved expressions, malformed URLs, unsupported schemes, and multiple enabled Datadog exporters fail startup. The Router does not probe Agent connectivity.

The Datadog trace endpoint and an OTLP endpoint are different protocols. Use the Agent trace port `8126`, not the common OTLP gRPC port `4317`, unless the relevant service is explicitly configured otherwise.

## Sampling and statistics

The Router owns one `SdkTracerProvider`. When native Datadog tracing is enabled, Datadog therefore owns provider-wide, parent-aware sampling. `parent_based_sampler` only configures the generic OpenTelemetry sampler and has no effect on the Datadog-backed provider.

The Router applies `collect.sampling` as Datadog's programmatic catch-all trace sample rate. That value overrides `DD_TRACE_SAMPLE_RATE`. More specific `DD_TRACE_SAMPLING_RULES` can run before the catch-all rate, and Datadog remote configuration can update native sampling behavior when remote configuration is enabled.

When explicit sampling is active, Datadog defaults to retaining at most 100 detailed traces per second. `DD_TRACE_RATE_LIMIT` changes that ceiling. At high request rates, the retained percentage can therefore be lower than `collect.sampling`. Datadog still records non-retained requests as `RecordOnly`, allowing native request, error, and latency statistics to cover all recorded requests. This is why `sampling: 0.0` keeps Router tracing instrumentation active when Datadog is enabled: detailed traces are not retained, but all-request APM statistics can still be computed.

OTLP and stdout processors attached to the shared Datadog-backed provider export only sampled spans, so they ignore Datadog `RecordOnly` spans. Hive tracing applies the same sampled-span guard. Sampled spans retain the Router's existing semantic-convention compatibility and redaction behavior.

## Resources and sensitive GraphQL data

The Router passes its OpenTelemetry resource to Datadog. Datadog's native resource merging rules determine precedence between Router attributes and equivalent `DD_SERVICE`, `DD_ENV`, and `DD_VERSION` values.

Agent-side scrubbing is not the default safety boundary. See the [Datadog data security guidance](https://docs.datadoghq.com/tracing/configure_data_security/) and the [OpenTelemetry GraphQL semantic conventions](https://opentelemetry.io/docs/specs/semconv/graphql/graphql-spans/). Having said that, Hive Router will omit the `graphql.document` from the traces because it may contain sensitive information (for example, when using inline variables).

## Propagation and lifecycle

The Router continues to install its configured OpenTelemetry propagators, including W3C Trace Context. It initializes Datadog locally and intentionally does not install the Datadog propagator. Native Datadog propagation requires future integration with the Router's `TelemetryContext` rather than replacing existing propagation.

The Datadog-backed provider is retained in the Router's existing telemetry state. Existing force-flush and shutdown handling therefore applies to Datadog and every attached processor. Datadog initialization can internally recover with a no-op provider and report errors through logging; only deterministic Router-owned configuration failures are fatal.


bin/router/src/config/telemetry/tracing.rs
```diff
@@ -218,6 +240,62 @@ pub enum TracingExporterConfig {
...
+///
+/// Enabling this exporter does not change trace propagation. The Router keeps
+/// the propagators configured under `telemetry.tracing.propagation`, including
+/// W3C Trace Context. Datadog supports W3C `traceparent` and `tracestate` headers,
+/// so Datadog-specific headers are not required when connected services also
+/// support W3C propagation.
+#[derive(Debug, Deserialize, Serialize, JsonSchema, Clone)]
+#[serde(deny_unknown_fields)]
+#[non_exhaustive]
+pub struct DatadogExporterConfig {
```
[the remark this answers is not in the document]
enisdenjo: got rid of it

bin/router/src/telemetry/traces/control.rs
```diff
@@ -1,9 +1,13 @@
...
+use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
 use tracing::{Level, Span};
 
 // Atomic representation of the max enabled tracing level.
 // 0: Off, 1: Error, 2: Warn, 3: Info, 4: Debug, 5: Trace
 static MAX_LEVEL: AtomicU8 = AtomicU8::new(3); // Default to Info
+                                               // datadog's private processor bypasses exporter wrappers, so this shared gate
+                                               // keeps graphql documents and their potentially sensitive literals out before
+                                               // spans reach any processor; provider rebuilds update it during config reloads
+static GRAPHQL_DOCUMENT_RECORDING_ENABLED: AtomicBool = AtomicBool::new(true);
```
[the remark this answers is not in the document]
enisdenjo: scoped to subscribers in https://github.com/graphql-hive/router/pull/1521/commits/ba610537eb52f71aec15f3c1b932112ecb4706c7

bin/router/src/telemetry/traces/mod.rs
```diff
@@ -63,31 +117,67 @@ pub(super) fn build_trace_provider<I>(
...
+        ));
+    }
 
-    if config.tracing.collect.parent_based_sampler {
-        builder = builder.with_sampler(Sampler::ParentBased(Box::new(base_sampler)));
+    let builder = if let Some(datadog) = datadog_config {
+        let mut datadog_config = DatadogConfig::builder();
+        if let Some(endpoint) = &datadog.endpoint {
+            let endpoint = resolve_value_or_expression(endpoint, "Datadog Agent endpoint")?;
+            datadog_config.set_trace_agent_url(endpoint);
```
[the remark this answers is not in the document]
enisdenjo: https://github.com/graphql-hive/router/pull/1521/commits/de176f1b1e0f05d795e4585b30d7c0348da5254c

bin/router/src/telemetry/traces/spans/graphql.rs
```diff
@@ -391,10 +391,12 @@ impl GraphQLOperationSpan {
         if self.span.is_disabled() {
             return;
         }
-
+        // this avoids recording sensitive documents when an exporter cannot redact them
```
kamilkisiela: do I understand correctly that enabling DataDog exporter disables recording the graphql document? If so then, when OTLP exporter is configured as well, it'd not receive the document, no?

bin/router/src/config/telemetry/tracing.rs
```diff
@@ -85,9 +89,27 @@ pub struct TracingCollectConfig {
     pub max_attributes_per_event: u32,
     #[serde(default = "default_max_attributes_per_link")]
     pub max_attributes_per_link: u32,
-    /// Can also be set via the `TELEMETRY_TRACING_SAMPLING_RATE` environment variable.
+    /// Fraction of traces to retain, from `0.0` to `1.0`.
```
kamilkisiela: Let's simplify comments a bit, they are very technical and hard to understand quickly

bin/router/src/telemetry/traces/standard_pipeline_exporter.rs
```diff
@@ -56,8 +56,11 @@ impl<E: SpanExporter> StandardPipelineExporter<E> {
 
 impl<E: SpanExporter> SpanExporter for StandardPipelineExporter<E> {
     async fn export(&self, batch: Vec<SpanData>) -> OTelSdkResult {
+        // datadog uses record-only spans for statistics, but standard
+        // exporters only get retained traces
         let processed_batch = batch
             .into_iter()
+            .filter(|span| span.span_context.is_sampled())
```
kamilkisiela: won't that change behavior for non-dd exporters?

kamilkisiela:
maybe worth adding a few tests where DD exporter is mixed with stdout or otlp to check if one does not affect the other when it comes to sampling, span attributes etc

bin/router/src/config/telemetry/tracing.rs
enisdenjo: https://github.com/graphql-hive/router/pull/1521/commits/fe23aefe71ff2cd395e68d9458d4ea5b9e696bed

bin/router/src/telemetry/traces/spans/graphql.rs
enisdenjo: yeah, enabling Datadog prevents `graphql.document` from being recorded on spans for that subscriber/provider; however, for OTLP this does not change exported data: `StandardPipelineExporter` already strips graphql.document before sending spans.

the difference is only _where_ redaction happens:

- Without Datadog: recorded on the span, then stripped from OTLP/stdout.
- With Datadog: never recorded, because Datadog’s native processor bypasses that exporter wrapper.

so what is received by the enedpoints is the same in both cases. yeaj one caveat is indeed that a Hive exporter sharing the Datadog-backed provider also loses the document before the exporter is reached.

bin/router/src/telemetry/traces/standard_pipeline_exporter.rs
enisdenjo: right right, I added it for mixed Datadog + standard exporter setups, to prevent Datadog `RecordOnly` spans from being exported through OTLP/stdout.

but checking the actual processor flow, our async batch processor already drops unsampled spans, and Hive has the same guard making this filter redundant.

removed in https://github.com/graphql-hive/router/pull/1521/commits/c2eeb62b0fc4e2855160a769970ca2d2aa6d6139

enisdenjo:
added some more tests in https://github.com/graphql-hive/router/pull/1521/commits/c214f11f3dce4f892df8a5148b5d0cacdd30312d

dotansimha (approved):

bin/router/src/telemetry/traces/spans/graphql.rs
enisdenjo: https://github.com/graphql-hive/router/pull/1530

PR #1524, opened 2026-09-09
fix(router): keep demand control cost formula next to the query plan

kamilkisiela:
The cost formula was cached only by the normalized operation hash, in a separate 1,000-entry demand control cache.

The query plan uses a more specific cache key: the operation hash, `StableOverrideContext`, and any plugin filtering. This means requests with different override labels could get different query plans but still reuse the same cost formula. A request could then be charged based on a plan it was not actually running, including the wrong per-subgraph costs.

Compile the cost formula together with the query plan and store both in the same cache entry. This makes sure they always use the same cache key. The separate demand control cache can then be removed.

`evaluate` no longer needs to look anything up. It receives the formula directly from the plan cache, so it is no longer async.

`cost.formula_cache_hit` is removed as well. It was declared in three places but never recorded, and there is no separate formula cache anymore.

### DIFF
--- a/bin/router/src/pipeline/demand_control/runtime.rs
+++ b/bin/router/src/pipeline/demand_control/runtime.rs
@@ -22,7 +22,6 @@ use crate::telemetry::metrics::Metrics;
 use crate::telemetry::traces::spans::graphql::GraphQLSpanOperationIdentity;
 use ahash::{HashMap as AHashMap, HashMapExt};
 use http::{HeaderName, HeaderValue};
-use moka::future::Cache;
 use tracing::{debug, info, warn};
 
 use crate::pipeline::error::{ClientPipelineError, PipelineError};
@@ -36,7 +35,6 @@ pub struct DemandControlRuntime {
     config: DemandControlConfig,
     expose_headers_flags: Arc<DemandControlExposeHeadersConfig>,
     metrics: Arc<Metrics>,
-    formula_cache: Cache<u64, Arc<DemandControlFormulaPlan>>,
 }
 
 impl DemandControlRuntime {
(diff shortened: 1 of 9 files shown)


bin/router/src/pipeline/demand_control/runtime.rs
```diff
@@ -251,7 +227,9 @@ impl DemandControlRuntime {
         over_limit
     }
 
-    fn compile_demand_control_plan(
+    /// Compiled once per plan-cache entry, while the plan is being built. The result is stored
+    /// with the plan, so it is keyed exactly like the plan it describes.
```
kamilkisiela:
```suggestion
```

bin/router/src/pipeline/query_plan.rs
```diff
@@ -21,9 +22,18 @@ use tracing::Instrument;
...
 
 pub enum QueryPlanResult {
-    QueryPlan(Arc<QueryPlan>),
+    QueryPlan(PlannedQuery),
     EarlyResponse(PlanExecutionOutput),
 }
+
+/// A plan and the demand-control cost formula compiled from it. The formula describes this exact
+/// plan, so it shares the plan's cache entry and key - which includes the override context and any
+/// plugin operation filtering, not just the operation hash.
```
kamilkisiela:
```suggestion
```

dotansimha (approved):

bin/router/src/pipeline/demand_control/runtime.rs
kamilkisiela:
```suggestion
```

bin/router/src/pipeline/query_plan.rs
kamilkisiela:
```suggestion
```

PR #1552, opened 2026-09-14
fix(router): honor `parent_based_sampler` for Datadog tracing

enisdenjo:
Make `telemetry.tracing.collect.parent_based_sampler` control Datadog sampling as well as OpenTelemetry sampling.

Previously, Datadog always inherited the sampling decision from an incoming remote parent. This meant `parent_based_sampler: false` had no effect when the Datadog exporter was enabled... A request with a sampled `traceparent` could still be retained even when the router's configured sampling rate rejected it.

This change marks an incoming Datadog parent decision as deferred when parent-based sampling is disabled. Datadog then runs its own configured sampling policy instead of inheriting the sampled flag.

### Why

The Datadog tracer installs and owns its sampler. Applying an OpenTelemetry sampler to the Datadog builder would be overwritten during Datadog initialization and would bypass Datadog-specific sampling behavior such as priorities, rules, rate limits, trace registration, and stats.

Datadog already supports a deferred trace flag (`0x02`). When it sees this flag, it ignores the parent's sampling decision and evaluates the trace with its own sampler. Reusing that behavior keeps the change small and preserves Datadog's existing sampling pipeline.

### Basically

```yaml
telemetry:
  tracing:
    collect:
      sampling: 0.0
      parent_based_sampler: false
    exporters:
      - kind: datadog
```

Given this incoming header:

```text
traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01
```

The final `01` says that the remote parent was sampled.

Before this change, Datadog inherited that decision and retained the trace despite `sampling: 0.0` and `parent_based_sampler: false`.

After this change, the router preserves the trace and parent IDs but changes the extracted sampling state to deferred before Datadog sees it. Datadog evaluates the trace using the configured `0.0` sampling rate, rejects the detailed trace, and still reports request statistics to the agent.

Setting `parent_based_sampler: true` keeps the existing behavior: the sampled remote parent is honored and the trace is retained.


bin/router/src/telemetry/mod.rs
```diff
@@ -568,7 +581,23 @@ impl TelemetryContext {
...
+            let parent = span.span_context();
+
+            // datadog normally inherits this decision before applying collect.sampling and its
+            // own rules. 0x02 defers that decision to the datadog sampler instead.
+            if self.defer_parent_sampling_decision && parent.is_remote() && parent.is_valid() {
+                // replace only the flags; trace identity, remote status, and trace state stay intact
+                return context.with_remote_span_context(SpanContext::new(
+                    parent.trace_id(),
+                    parent.span_id(),
+                    TraceFlags::new(0x02),
```
kamilkisiela, line 597: Could we link to DD repo?
https://github.com/DataDog/dd-trace-rs/blob/0a455eff98accfc42110a2c367c899d333281367/datadog-opentelemetry/src/text_map_propagator.rs#L26

just so we can find it later or see it drifted, as it's not really officially documented

enisdenjo: https://github.com/graphql-hive/router/pull/1552/commits/cda1d605de39fcefeab6fd0acb403f5c1aa827ac

kamilkisiela (approved):

PR #1557, opened 2026-09-15
fix(router): mark HTTP 200 GraphQL errors as failed telemetry spans

enisdenjo:
Mark GraphQL errors as errors on both the `graphql.operation` span and the root `http.server` span, including when the HTTP response status is `200 OK`.

This makes GraphQL failures visible to APM error-rate calculations instead of reporting a 0% service error rate while GraphQL errors are present in successful HTTP responses.

## Problem

GraphQL errors are commonly returned in an HTTP 200 response body. Previously, the router treated every non-5xx HTTP response as a successful span and explicitly recorded:

```text
otel.status_code = "Ok"
```

This caused two problems:

1. `graphql.operation` spans containing GraphQL errors did not receive an OpenTelemetry error status.
2. The root `http.server` span was marked `Ok` based only on its HTTP status, so GraphQL errors were not reflected in service-level APM error rates.

Plugins could not reliably work around this. The plugin middleware invokes `on_http_request` before the handler creates the root HTTP span, and callbacks registered there can run after that span has already closed and been exported. Additionally, recording the final HTTP response could overwrite an error status with `Ok` for any non-5xx response.

## Ok so

- set `otel.status_code` to `Error` on `graphql.operation` when its recorded GraphQL error count is greater than zero
- mark the root `http.server` span as `Error` when the final GraphQL response metric status indicates an error, even if the HTTP response status is 200
- stop explicitly marking non-5xx root HTTP spans as `Ok`. Their status remains unset unless another error is recorded
- continue marking 5xx root HTTP responses as `Error` and recording their HTTP status as `error.type`

## Successful HTTP spans now use an unset status

An unset span status is the OpenTelemetry success state for normal 1xx, 2xx, and 3xx HTTP responses. Explicitly setting `Ok` is unnecessary and can clobber an error discovered elsewhere in the request lifecycle.

The HTTP semantic conventions require status to remain unset for 1xx, 2xx, and 3xx responses unless another error occurred:

> [Span Status](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.59.0/specification/trace/api.md#set-status) MUST be left unset if HTTP status code was in the 1xx, 2xx or 3xx ranges, unless there was another error (e.g., network error receiving the response body; or 3xx codes with max redirects exceeded), in which case status MUST be set to Error.

_https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status_

See snapshot updates

```diff
- Status: message='' code='1'
+ Status: message='' code='0'
```

in

```
e2e/src/telemetry/tracing/otlp_attributes.rs
```

where

- OTLP status code `1` = `Ok`
- OTLP status code `0` = `Unset`
- OTLP status code `2` = `Error`

## GraphQL errors do not have one honest `error.type`

A single GraphQL response can contain multiple errors with different types or codes. This change sets the aggregate span status to `Error` but does not invent one `error.type` for the root or operation span. Individual error details remain on span events.

## TODO

- [x] is hive tracing impacted? no


dotansimha (approved):
lgtm. let's merge and proceed only after confirming this with Console

enisdenjo:
confirmed no changes in hive tracing

Issue #1565, opened 2026-09-16
Add stable supergraph variant labels for observability


kamilkisiela:
In a multi-supergraph setup, we currently can't tell which supergraph variant served a request. This makes it hard to break down traffic, errors, cache usage per variant.

Using a schema version ID or supergraph hash does not solve this well. Those identify a specific schema revision and change frequently, so high  cardinality.

---

Give every selected supergraph a stable variant label or name

```
supergraph.variant = "public"
```

The default configured supergraph could use `default`, while plugin-selected supergraphs could provide their own label.

Schema version/hash can still be exposed separately when we need to identify the exact schema revision.

The distinction is:

```
variant -> stable observability dimension
schema version -> exact revision for debugging
```

PR #1567, opened 2026-09-16
feat(router): added supergraph `name` to plugin api and observability

dotansimha:
Related https://github.com/graphql-hive/router/issues/1565

Every `Supergraph` now carries a human-readable `name`. The supergraph loaded from the router configuration (`supergraph.source`) is named `default`. Plugins that construct their own supergraphs - for example to serve a different schema variant per tenant or feature flag - give each one a name, and that name is what the router reports when describing which supergraph handled a request.

## Plugin API

`Supergraph::from_sdl` and `Supergraph::from_document` take the name as a new, required first argument:

```rust
// before
let variant = Supergraph::from_document(document, options)?;

// after
let variant = Supergraph::from_document("my-supergraph", document, options)?;
```

## Logging

The request summary line (`router::request` target) reports the name of the supergraph that processed the request:

- **Added** `supergraph_name` - the selected supergraph's name, e.g. `"default"` or the name a plugin chose. It is absent when the request was rejected before a supergraph was selected.
- **Removed** `supergraph_identifier` - the internal numeric id previously printed in its place. It was not stable across restarts or reloads and was never meaningful to operators.

Example (JSON format):

```json
{"level":"INFO","target":"router::request","operation_type":"query","status_code":200,"supergraph_name":"default","duration_ms":12,...}
```

## Traces

The `graphql.operation` span gains the attribute:

- `hive.supergraph.name` - the selected supergraph's name.

It is recorded as soon as the supergraph is selected, for both HTTP and WebSocket (`graphql-ws`) operations.

Only the operation span carries it; child spans (parse, validate, plan, subgraph calls) are unchanged and can be attributed to a supergraph through their parent.

## Metrics

A `supergraph.name` label is added to:

- `http.server.request.duration`
- `hive.router.graphql.errors_total`

Both now break down by supergraph, so a deployment serving several schema variants from one router can chart latency and error rates per variant.

For `http.server.request.duration` the label is omitted on requests that failed before a supergraph was selected (e.g. malformed HTTP requests).

For `hive.router.graphql.errors_total` it is omitted for errors raised before selection.


enisdenjo (approved):

kamilkisiela:
I think we should prevent having two variants with the same name, otherwise it may affect prom metrics. One cache of supergraph variant 1 (name "A") can have 69 entries in the cache, and supergraph variant 2 (name "A") can have 420. Looping over the caches and recording their sizes will report wrong numbers.

Issue #1579, opened 2026-09-17
Router overhead metric


dotansimha:
Per-request accounting of the time the router spends waiting on external services (subgraphs, coprocessors) - the "idle" part of a request.

Waits can overlap (parallel plan nodes fan out to several subgraphs at once), so the tracker needs to measure the *union* of the wait intervals rather than their sum.

PR #1580, opened 2026-09-17
feat(router): new metrics to measure router overhead

dotansimha:
Related https://github.com/graphql-hive/router/issues/1579 

Two new histograms report how much of a request's latency is the router's own work versus time spent waiting on external services:

- `hive.router.request.overhead.duration` - the request duration (`http.server.request.duration`) minus the time the request spent waiting on external services. This is the latency the router adds on top of your upstreams: parsing, validation, planning, response merging, plugins.
- `hive.router.request.external_wait.duration` - the time the request spent waiting on at least one external service.

The following are considered "external" and count as waiting:

- Persisted document fetches from the Hive CDN on a cache miss also count as waiting

- Subgraph calls 

- Co-processor calls

They are not recorded for streamed responses (subscriptions, incremental delivery), whose subgraph waits continue after the response headers are sent.

With explicit histogram buckets (the default), `hive.router.request.overhead.duration` uses fine-grained boundaries from 100µs to 1s instead of the shared `seconds` buckets, since router overhead is typically sub-millisecond.


bin/router/src/executor/executors/http.rs
```diff
@@ -455,10 +456,13 @@ impl SubgraphExecutor for HTTPSubgraphExecutor {
                 };
 
                 if deduplicate_request {
+                    let external_wait_guard = external_wait::enter();
```
kamilkisiela, line 459: Shouldn't it be created after the semaphore's acquire?

kamilkisiela: If the goal is to measure "external" work then it should

kamilkisiela: because the semaphore acts as a concurrency limiter and it may cause the increase in "external" latency

bin/router/src/telemetry/external_wait.rs
```diff
@@ -0,0 +1,229 @@
...
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner())
+    }
+
+    pub fn enter(self: &Arc<Self>) -> ExternalWaitGuard {
+        self.enter_at(Instant::now())
+    }
+
+    fn enter_at(self: &Arc<Self>, now: Instant) -> ExternalWaitGuard {
+        let mut inner = self.lock();
```
kamilkisiela, line 49: Did we measure the overhead?

bin/router/src/executor/executors/http.rs
dotansimha: yes, fixed

bin/router/src/telemetry/external_wait.rs
dotansimha: the overhead of the router, or the overhead added by the overhead measurement? 😆

PR #1594, opened 2026-09-21
fix(telemetry): correct span attribute value types

praguevara:
These attributes had the string type, but the semantic conventions (semconv) require the integer type:
* [`http.response.status_code`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/) (used `StatusCode::as_str()`)
* [`http.request.body.size`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/)
* [`http.response.body.size`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/)
* [`server.port`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/server/)
* [`client.port`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/client/)
* [`network.peer.port`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/)

If they don't have the type i64, they are coerced to strings, so we need to explicitly set it.


bin/router/src/telemetry/traces/spans/http_request.rs
```diff
@@ -211,7 +211,7 @@ impl HttpServerRequestSpan {
             self.span,
             "http.response.status_code" = 500,
             "otel.status_code" = "Error",
-            "error.type" = 500,
+            "error.type" = "500",
```
yanns, line 214: `error.type` is now a String?

praguevara: https://opentelemetry.io/docs/specs/semconv/registry/attributes/error/#error-attributes it should be

yanns (approved):

dotansimha (approved):

dotansimha (approved):

